PR #2, opened 2026-06-15
Infer the launch cwd as the default root

tobert:
## Why

kaibo's containment layer already adopts the launch cwd as an allowed tree when no `--root`/`--allow-path` is set (MCP clients start stdio servers with cwd = workspace). But the **default root** — what a call resolves to when it omits `path` — stayed unset, so every `consult`/`explore`/`synthesize`/`run_kaish` call had to pass `path` explicitly even though kaibo was sitting in exactly the directory the caller meant. The convenience layer lagged the security layer. This surfaced as friction on a `/configure` run.

## What changed

- When no `--root` is configured, adopt the canonicalized cwd as the **inferred default root** — but only **when it falls inside the allowed set**. In the zero-config case the cwd *is* the whole allowed set, so it always qualifies.
- The guard is the safety: an `--allow-path` that excludes the cwd leaves the default root unset rather than pointing it at a path the containment check would reject. We never default to something we'd refuse.
- `resolve_root` re-validates the inferred root through the same canonicalize + containment path as an explicit `path` — no special-case trust.
- The boundary stays legible: the `## Scope` handshake and `kaibo://config` (`default_root_inferred`) tag the root as inferred so a reader can tell it wasn't configured by hand.

## Behavior change

This deliberately replaces the old "containment ≠ defaulting" rule (an omitted `path` used to always error without `--root`). Containment tests and `docs/config.md` are updated to the new contract.

## Tests

Failing-first tests in `tests/containment.rs`:
- `omitted_path_zero_config_infers_cwd_as_default_root` — omit `path`, land on the workspace; handler reports the inferred root.
- `omitted_path_with_allow_path_excluding_cwd_still_errors` — cwd outside the allow-path → no default root, omitted `path` still a parameter error.

Plus scope-section unit tests for the inferred-vs-explicit labelling. Full suite green; no new clippy warnings.

> Note per AGENTS.md: still wants a cross-family review pass over the diff before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,11 @@ the git log. Each later release appends a new section at the top.
   and still land on the project. The scope handshake and `kaibo://config` tag the
   root as inferred. An `--allow-path` that excludes the cwd leaves no default root —
   kaibo never defaults to a path its own containment check would reject.
+- **`~` expands in `[server] root` and `allow_paths`** (config-file and `KAIBO_*`
+  env layers), matching the tilde handling key files and `[context]` paths already
+  get. Set `allow_paths = ["~/src"]` once and every project under it is in-bounds —
+  with cwd inferred as the default root, you stop thinking about `path` entirely.
+  (Previously a literal `~` was taken verbatim and failed canonicalization at startup.)
 - **Per-tool gating.** Each tool has a `--no-<tool>` flag (all on by default); an
   all-off server is refused at startup.
 - **Operator ignore files** via a `[kaish.ignore]` config stanza.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,12 @@ the git log. Each later release appends a new section at the top.
   writing `config.toml`, alongside `kaibo://config` (resolved runtime state) and
   `kaibo://config/example` (annotated template) resources. Secrets are referenced by
   env-var name or key-file path, never inlined.
+- **Zero-config workspace root.** When no `--root` is set, kaibo adopts its launch
+  cwd as the inferred default root (it already scoped containment to that cwd, and
+  MCP clients start stdio servers with cwd = workspace), so a call may omit `path`
+  and still land on the project. The scope handshake and `kaibo://config` tag the
+  root as inferred. An `--allow-path` that excludes the cwd leaves no default root —
+  kaibo never defaults to a path its own containment check would reject.
 - **Per-tool gating.** Each tool has a `--no-<tool>` flag (all on by default); an
   all-off server is refused at startup.
 - **Operator ignore files** via a `[kaish.ignore]` config stanza.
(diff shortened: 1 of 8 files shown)


tobert:
## Cross-family review summary (kaibo `synthesize`, deepseek cast)

Per AGENTS.md, the diff got a cross-family review pass — a different model lineage (deepseek) than the Opus that wrote it, run through kaibo's own `synthesize` grounded in the diff and the live source. Two passes, one per behavioral change. Both passed as **safe to land**; all findings were Low severity and the ones with teeth are fixed.

### Pass 1 — default-root inference (core change)

**Verdict:** safe — containment invariant holds in every configuration. Verified: no boundary widening (the `cwd.starts_with(tree)` guard mirrors `resolve_root`'s step-3 check, so we never default to a path containment would reject); the allowed set is byte-identical to before in all four branches (zero-config / `--root` / `--allow-path` / combined); the double-canonicalization in `resolve_root` is TOCTOU *hardening*, not a hole.

Findings → fixes (commit `622de2d`):
- **Ancestor-vs-descendant semantic** now has its own test — when the only `--allow-path` is a *descendant* of cwd, cwd is above the boundary and is not adopted (the prior test used a sibling, same branch but a distinct semantic).
- **Explicit `--root` not tagged inferred** — added an assert so an accidental `(Some(root), false)`→`true` flip is caught (would mislabel `kaibo://config`/scope).
- **Clarifying comments** at the inference site (push-before-guard ordering + resolve_root-mirroring intent).

### Pass 2 — `~` expansion in `[server] root` / `allow_paths`

**Verdict:** safe — expansion is narrow (leading `~`/`~/…` only), every other input (`~user`, mid-path `~`, empty, `$HOME` unset) passes through and fails loudly at canonicalization. Confirmed the single expansion point covers both the config-file and `KAIBO_*` env layers (both write the same `RawServer` fields before `merge`).

Findings → fixes (commit `33d2a81`):
- **`expand_tilde` doc-comment** now forbids widening to `~user` (no `getpwnam` from a path that feeds containment) and notes the `$HOME`-unset passthrough.
- **Standalone `expand_tilde` unit test** added (it had none): leading-`~`, bare `~`, and `~user`/mid-path/absolute passthrough.
- **Env-layer test** added — `KAIBO_ROOT` + colon-split `KAIBO_ALLOW_PATHS` expand `~` (the file-layer test didn't reach this).
- **Latent test flake fixed** — trim trailing `/` off `$HOME` so the expectation matches `PathBuf::join` normalization.
- **CLI asymmetry documented** in `--root`/`--allow-path` help (shell expands `~` on the CLI; kaibo expands it in config/env). Left CLI expansion to the shell rather than round-tripping a `PathBuf` through `to_string_lossy`, which would corrupt a non-UTF8 path.

### State
Full suite green, no new clippy warnings. Four commits: core inference, its review hardening, tilde expansion, its review hardening.

PR #8, opened 2026-06-22
Batch tools: offline, max-effort fan-out — Anthropic + Gemini

tobert:
## What

A new **batch** tool class — the offline, async sibling of `oneshot`. Submit a list of self-contained, tool-less prompts; get back a durable handle; poll it; read every answer when the provider's (cheaper) batch lane finishes. No upstream call is held open per answer.

Four verbs behind one `--no-batch` gate (they're one capability):

- **`batch_submit`** — fan a list of prompts onto the provider's batch lane; returns a `backend/provider-id` handle.
- **`batch_get`** — poll by handle: a progress line while pending, every item's answer (per-item failures surfaced, not dropped) when done.
- **`batch_cancel`** — stop a running batch by handle.
- **`batch_list`** — re-discover the batches a backend still holds (newest-first handles with status/progress) — the way back to a batch whose handle was lost.

Anthropic Message Batches only for now, behind a per-provider `BatchProvider` seam (the `ImageGen` discipline); Gemini and OpenAI are tracked follow-ons (`docs/issues.md`). A non-Anthropic cast is refused loudly, never silently no-op'd.

## Why these shapes

- **A class, not a flag.** Batch is toolless *by construction* — provider batch APIs are offline and can't drive a tool loop — so it's built on the `oneshot` shape, never `run_phase`.
- **Max the knobs.** Batch floors `max_tokens` and forces high thinking effort regardless of how a cast was tuned for interactive use: the latency that makes max-thinking painful synchronously is free once you've accepted "come back later" (Amy's motivating case: Pro is near-unusable interactively, but batch is exactly where you reach for it). `BATCH_EFFORT` is threaded explicitly through `ModelShape::to_params` so it stays decoupled from `consult`'s interactive default.
- **No state.** The handle is the whole address; poll/cancel/list rebuild a fresh client and re-address the provider's own id, surviving a restart. The split now *enforces* the no-`/`-in-backend-name invariant at config load rather than trusting convention.
- **Its own preamble.** Batch shares oneshot's toolless shape but not its words — overridable via `[prompts].batch`. See the review note below.

## Cross-family review (dogfooded)

Per the project's PR discipline, the design got a cross-family review — **Opus run *through* `batch_submit` itself**, critiquing its own tool design and system prompt. It drove three changes landed in this PR:

1. **Tailored batch preamble** — oneshot's "name what you'd need rather than guessing" is right synchronously but wrong offline (no next turn to fill a gap). Batch now tells the model it gets one complete response, to spend on depth, and to *state an assumption and answer under it* rather than stall (also fixing a negative-framing lapse).
2. **`batch_list`** — the review's sharpest finding was the orphaned-batch billing risk (a lost handle = a batch that keeps billing with no way back, since kaibo holds no state). `batch_list` closes it; live-verified by recovering 14 real historical batches.
3. **Backend-name `/` guard** — hardened the handle-split invariant at config load.

Three findings remain tracked as deferred follow-ons in `docs/issues.md` (index-as-custom_id context-loss, prose-vs-structured status, effort-as-floor-vs-force) — none blocking.

## Testing

- Offline-tested throughout: pure request-shaping / response-parsing / render fns, plus a `ScriptedBatch` driving submit→pending→done and a seeded list with no network (mirrors `ScriptedImageGen`).
- **191 lib tests + all integration suites green**; no new clippy warnings.
- Live-verified end to end against the real Anthropic API: submit (Opus override), poll, the per-item provenance footer, and all three `batch_list` paths (default, scoped, non-Anthropic loud refusal).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,25 @@ the git log. Each later release appends a new section at the top.
   caller, not reasoning run into kaibo's own models): prompt → image, returned inline
   as MCP image content. OpenAI-compatible image backends only (hosted
   `gpt-image` / DALL·E, or a local Stable-Diffusion server).
+- **Batch (`batch_submit` / `batch_get` / `batch_cancel` / `batch_list`)** — the
+  *offline, async sibling* of `oneshot`: submit a list of tool-less prompts, get a
+  handle, poll it, read every answer when the provider's batch lane finishes — no call
+  held open per answer. Built for fanning many prompts (or one hard question you'll wait on) at a
+  top-tier model: it maxes the knobs (forces high thinking effort + a generous token
+  budget) regardless of how the cast was tuned for interactive use, and a per-call
+  `model`/`backend` override lets you batch a Pro/Opus tier a cast otherwise synths
+  cheaper. Each prompt is self-contained — no codebase access, no tools. kaibo keeps
+  no state: the handle is the whole address, so poll/cancel survive a restart, and a
+  failed item is surfaced per-item rather than dropped. Anthropic backends only for
+  now (Gemini and OpenAI batch are tracked follow-ons); a non-Anthropic cast is
+  refused with a clear message. Gated by `--no-batch` (one flag over every verb).
+  Batch carries its own system preamble fit to the offline lane — one complete,
+  self-contained response with no follow-up, told to spend on depth — overridable via
+  `[prompts].batch` like the other phases. While a batch runs, `batch_get` reminds you
+  to go do other work and check back rather than wait on it. Lost a handle?
+  `batch_list` re-discovers the batches a backend still holds (newest first, each with
+  its handle, status, and progress), so a batch is never orphaned — defaulting across
+  every Anthropic backend, or scoped to one with `backend`.
 - **`view_image`** — vision-capable consultation phases can read an image *file* from
   the workspace into model context (screenshots, diagrams, assets already in the tree).
 - **Multi-provider model teams.** Anthropic, DeepSeek, and Gemini natively, plus a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,16 +42,18 @@ the git log. Each later release appends a new section at the top.
   `model`/`backend` override lets you batch a Pro/Opus tier a cast otherwise synths
   cheaper. Each prompt is self-contained — no codebase access, no tools. kaibo keeps
   no state: the handle is the whole address, so poll/cancel survive a restart, and a
-  failed item is surfaced per-item rather than dropped. Anthropic backends only for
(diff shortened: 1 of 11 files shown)


tobert:
## Update — Gemini batch provider added (`845479c`)

The second `BatchProvider` behind the seam ships in this PR now: **Gemini inline batch**, plus a ready-made `gemini-batch` cast (synth → `gemini-pro-latest`). The trait, the stateless `backend/provider-id` handle, the `ScriptedBatch` harness, and the four verbs were reused unchanged; kind dispatch moved into `batch.rs` (`submitter`/`poller`/`batch_supported`) so `server.rs` no longer hardcodes a provider. User-facing batch text is now provider-agnostic, and refusals derive their supported set dynamically from `batch_supported` so they can't drift as more backends land.

### Wire shape — probed, not guessed
Gemini's batch differs from Anthropic's in ways reading wouldn't have settled:
- **Body can't be shared**: Gemini nests *both* `maxOutputTokens` and `thinkingConfig` under one `generationConfig` (Anthropic carries `max_tokens` top-level). `batch_shaping` itself is reused verbatim.
- **Results inline** in the long-running-operation object — no separate results URL.
- **`batchStats` counts are JSON strings** (`"7"`) — `as_u64` alone would read 0; handled.
- **Cancel is instant-terminal** (not Anthropic's interim "canceling") → new `BatchPoll::Failed { state, message }` so a terminal-with-no-results batch names the state honestly instead of a misleading empty `Done`; partial results are still returned as `Done`.
- **List under `operations`** + `nextPageToken` (not `data`/`has_more`).
- The slashed provider id (`gemini/batches/<id>`) round-trips for free via split-on-first-slash + the no-`/`-in-backend-name invariant.

### Live dogfood caught a real bug
A real `gemini-batch` run surfaced that the cast's first synth id (`gemini-3-pro-preview`) was **retired** — and that Gemini batch does *no* model validation at submit time (it accepted the dead model as `PENDING`; the failure only landed as a per-item error at run time). The per-item failure surfacing did its job, and the fix is the drift-resistant default: the cast synths `gemini-pro-latest` (the alias), not a pinned preview.

### Cross-family review
DeepSeek (`deepseek-v4-pro`, via kaibo's own `consult`) reviewed the diff: **no correctness bugs**, and it independently verified the slashed-handle round-trip, dispatch coverage, `BatchPoll::Failed` exhaustiveness, and the edge cases. It flagged **5 stale doc comments** — three on input-struct fields that ship in the MCP schema — now fixed.

### Testing
- All offline suites green; clippy-clean; formatted (edition 2021).
- Live end-to-end through the actual MCP server: `submit → list → poll → answer` on the `gemini-batch` cast default, a cancel that lands in `Failed`, and a non-batch cast (`deepseek`) refused with the supported set named.

PR #11, opened 2026-06-22
oneshot attachments + sharpen the attach follow-ups

tobert:
Builds on the batch `attach` feature (PR #10): brings the same workspace-file inlining to the interactive **`oneshot`** tool, and sharpens the tracked follow-ups with what we learned from `gpal` and Google's docs.

## What's here

**`oneshot` grows `attach`** — the interactive twin of batch attach. Name workspace files; kaibo reads + inlines them (text folded into the prompt via the shared `<file>` wrapper, images as native rig image parts) so their bytes never pass through the calling agent's context. "Call Opus once with these files, no tools, no waiting" is now one call.

- **Chose "grow oneshot against rig"** over a batch-shaped interactive seam — oneshot already runs through rig, so map each `Attachment` onto rig's own message parts rather than invent a second submit abstraction. Reuses the whole batch seam: `resolve_attachments`, `classify`, the shared `contained` containment boundary, and `Attachment::wrapped_text` (one source of truth for the `<file>` wrapper).
- **Vision gate DRY'd** into `KaiboHandler::gate_image_attachments`, called by both batch and oneshot — identical refusal, and unit-testable directly (the oneshot handler needs a `Peer`).
- **The loop stays pure.** Per the Gemini Pro holistic review, `run_phase`/`Arm::run`/`PhaseRunner` take a pre-built `initial_prompt: Message` (not `prompt + extra_parts`); oneshot assembles its own multi-part turn, consult/explore pass `Message::user(prompt)` byte-for-byte. Multimodal awareness lives only in oneshot.

**issues.md sharpened** with gpal + Gemini-docs findings (no code impact):
- `FileRef`/Gemini File API for batch is bigger than "a variant beside Image" and maybe the wrong shape — gpal's batch is inline-only; Google's docs only document file refs for *file-based JSONL* batch, not inlined.
- Per-item duplication is bounded by the ~20 MB inlined-batch cap; dropped the dedupe framing (a shared `Arc` still serializes N times on the wire). Real escapes: context caching (`cached_content`, shared-attachment fit) and the file-based JSONL tier spilled to an XDG cache.
- TOCTOU likely fix reframed from `/proc`+`O_NOFOLLOW` to reading through the kaish VFS like `view_image` (portable, structural).
- Noted the `<file>` wrapper's lack of body-escaping (shared by batch+oneshot, self-inflicted, defense-in-depth).

**Tool-description tuning** — de-emphasized `git diff > x.diff` (a tool-less model reviews best from whole files, not changed hunks); noted that top-tier models' 1M+ context windows mean you can attach a lot.

## Review

Dogfooded both ways on the live server: **DeepSeek realtime** via `oneshot`+`attach` (reviewing this very change through the feature it adds), and **Gemini Pro async** via `batch_submit`+`attach`. Both judged it sound; DeepSeek's nits (empty-text-part wart, expect wording) and Gemini's one structural call (the `initial_prompt: Message` refactor above) are folded in. Full suite green throughout.

## Not in scope (tracked in issues.md)

`consult`-attach (a different semantic — pin whole files as trusted starting evidence, overlapping the existing `context` param), the File-API/file-based-batch path, context caching, the payload-cap guard, and the TOCTOU structural fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,11 @@ the git log. Each later release appends a new section at the top.
   prompt in, answer out, no codebase access and no tools, exactly one upstream
   request. The counterpart to `consult` for when you already own the context (you've
   pasted what's needed, or the question is general). Pick the model with `cast`; the
-  answer carries the same provenance footer.
+  answer carries the same provenance footer. Takes the same **`attach`** as
+  `batch_submit` — name workspace files ("review README.md", or `git diff > x.diff`
+  then `attach: ["x.diff"]`) and kaibo inlines them (text as text, images as native
+  image parts on a vision-capable model) so their bytes never pass through your context.
+  So "call Opus once with these files, no tools, no waiting" is a single call.
 - **`run_kaish`** — drive the read-only kaish shell yourself, no model in the loop:
   exit code + stdout + stderr.
 - **`generate_image`** — kaibo's first *capability* (an artifact handed back to the
--- a/docs/devlog.md
+++ b/docs/devlog.md
@@ -14,6 +14,42 @@ per ship date; multiple ships on a date get sub-bullets.
 
 ---
 
+## 2026-06-22 — `oneshot` grows `attach` (the interactive twin of batch attach)
+
+Amy's framing: `oneshot` should be as close to `batch` as the API split allows — "call
+Claude Opus in a single shot with some files, no tools, no wait." So `oneshot` now takes
+the same `attach: [paths]`, reusing the whole containment + classify seam batch built
+(`resolve_attachments`, `classify`, the shared `contained`/`containment_error`). The only
+thing that differs is the wire, and that's the decision this entry records.
+
+**Chose "grow oneshot against rig" over "a batch-shaped seam with an interactive impl."**
+oneshot already runs through rig's completion path, so the cheap, honest factoring is to
+map each `Attachment` onto rig's own message parts rather than invent a second submit
+abstraction. Text folds into the prompt string via `attach::with_text_context` (the same
+`<file>` wrapper batch emits — `Attachment::wrapped_text` is the one source of truth, so
+batch and oneshot wrap identically). Images become `UserContent::image_base64` parts on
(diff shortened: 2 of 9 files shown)


tobert:
## Live verification on a fresh binary

Built the branch, reconnected, and smoke-tested every `attach` path against expectations — all six behave as designed:

| Path | Result |
|---|---|
| **oneshot text inline** (`src/attach.rs` → deepseek) | ✅ Grounded — DeepSeek quoted the exact `classify` error string and named both encodings from the file |
| **oneshot image → vision model** (`docs/brand/banner-teal.png` → anthropic) | ✅ Claude accurately described the real image (teal dinosaur w/ necktie + to-go cup, "kaibo 解剖") — the image rode as a real structured part and was seen |
| **batch attach** (`CHANGELOG.md` → gemini-batch) | ✅ Both batch answers reasoned from the inlined changelog (named the `attach` param, the png/jpeg/gif/webp formats, the vision requirement) |
| **Containment** (`/etc/hostname`) | ✅ Refused: "outside the allowed set […]" + the three widening knobs |
| **Vision gate** (image → non-vision deepseek) | ✅ Refused: "the model `deepseek-v4-pro`… doesn't accept image input" |
| **Directory** (`docs`) | ✅ Refused: "is not a regular file" |

Two things the offline tests can't fully prove, confirmed live end-to-end:
- **Text reaches the model and grounds the answer** (DeepSeek quoting `attach.rs` verbatim).
- **An image rides as a real part to a vision model and is seen** (Claude describing the banner) — and is **refused** to a non-vision model *before* any network call.

The three refusals fire before the network (containment/gate/directory are pure), so they're instant and free. The whole point held: the verified files were passed **by path only** — e.g. the ~2,360 tokens of `CHANGELOG.md` and the banner PNG bytes never crossed the calling agent's context.

Both cross-family reviews (DeepSeek realtime, Gemini Pro async — each dogfooded *through* the attach feature itself) judged the slice sound; their findings are folded in (empty-text-part wart, the `initial_prompt: Message` refactor that keeps the shared loop free of multimodal concerns).

PR #12, opened 2026-06-22
Expand $VAR / ${VAR} in root and allow_paths for portable scratch reads

tobert:
## Why

"Can a user easily allow kaibo to read `/tmp`?" The widening knobs already existed (`--allow-path` / `KAIBO_ALLOW_PATHS` / `[server] allow_paths`), but the *path* you had to write was host-specific — a literal `/tmp` is wrong on macOS (per-user `/var/folders/...`) and on sandboxed Linux (`$XDG_RUNTIME_DIR`). kaibo already followed XDG for its config file but carried no XDG/POSIX awareness into the read boundary.

## What

`root` / `allow_paths` now expand `$VAR` / `${VAR}` on top of the leading `~` they already did, so a scratch space resolves per-host:

```toml
[server]
allow_paths = ["~/src", "$TMPDIR", "$XDG_RUNTIME_DIR/kaibo"]
```

The full surface ships together so a user meets the idiom wherever they hit the boundary:
- **Core** — `expand_path` / `expand_env_vars` in `config.rs`, wired into the `root` / `allow_paths` merge. Env-first-then-tilde ordering; an **undefined variable is a loud load error**, never a silent gap that could misplace the boundary.
- **`configure` prompt** — gained an opt-in read-scope step teaching the portable `$TMPDIR` / `$XDG_RUNTIME_DIR` form and the read-only-but-real tradeoff.
- **Containment error** — now names the portable form when a path lands out of bounds.
- **Docs** — `docs/config.md` documents expansion + the scratch-read workflow; `CHANGELOG.md`, `docs/devlog.md` updated.

## Decisions

- **General `$VAR` over a curated temp-var set or `--allow-tmp` knob** — least-surprising to shell users, costs no more, and `$TMPDIR`/`$XDG_RUNTIME_DIR`/`$HOME` all fall out.
- **Scoped to the boundary knobs.** `[context]` `user_files` and key-file paths stay tilde-only (they're infallible today; converting means threading `Result` through their loaders) — logged in `docs/issues.md` as consistency follow-up rather than smuggled in.
- **Opt-in, never default.** Widening to a shared, world-writable `/tmp` is a real (read-only) boundary move, so kaibo never adds it for you.

## Review (cross-family, per house rule)

- **DeepSeek V4 Pro** (`consult` over the diff — kaibo dogfooding its own review path): cleared the security properties (undefined-var-is-loud holds; canonicalize + `is_dir` + `starts_with` stay the defense in depth). Three low-risk findings folded in commit `dd8cc13`: `${...}` name validation, a doc-comment accuracy fix, and added parser tests (adjacent refs, literal `$$`, brace-name validation).
- **Gemini 3.1 Pro** (batch lane): ⏳ in flight — will fold any further findings before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,9 +95,11 @@ the git log. Each later release appends a new section at the top.
   `path` entirely. (Previously a literal `~` was taken verbatim and failed
   canonicalization at startup.) Environment variables make a scratch space portable:
   `allow_paths = ["$TMPDIR"]` or `["$XDG_RUNTIME_DIR/kaibo"]` lets kaibo read artifacts
-  a workflow drops in a temp dir without hardcoding a host-specific `/tmp`. An undefined
-  variable is a loud load error, never a silent gap that would misplace the read
-  boundary; the `configure` prompt now walks you through this opt-in.
+  a workflow drops in a temp dir without hardcoding a host-specific `/tmp`. A variable
+  that is unset, **set but empty**, or non-UTF-8 is a loud load error, never a silent gap
+  that would misplace the read boundary (an empty `$EMPTY/` would otherwise collapse to
+  `/`); write `$$` for a literal `$`. The `configure` prompt now walks you through this
+  opt-in.
 - **Follow git worktrees automatically.** A `path` in a linked git worktree of an
   already-allowed repo is now reachable without an `--allow-path` — so a sibling
   branch you check out next to the project (even one you spin up mid-session) just
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,11 +88,16 @@ the git log. Each later release appends a new section at the top.
   and still land on the project. The scope handshake and `kaibo://config` tag the
   root as inferred. An `--allow-path` that excludes the cwd leaves no default root —
   kaibo never defaults to a path its own containment check would reject.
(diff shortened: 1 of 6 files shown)


tobert:
**Gemini 3.1 Pro batch review landed and is folded** (commit `cac35e0`).

It cleared single-pass evaluation and the operator-controlled `..` concern, and surfaced one real bug the DeepSeek pass and I had both under-weighted:

- 🔴 **Empty-but-set variable widened the boundary silently.** `undefined = loud error` doesn't catch `TMPDIR=""` — `std::env::var` returns `Ok("")`, so `$TMPDIR/scratch` collapsed to `/scratch` and `$TMPDIR/` to `/` (the whole filesystem). Now refused in a pure `resolve_var` seam (unit-tested without mutating process env).
- 🟡 **Non-UTF-8 error conflation** (both reviewers): a set-but-non-UTF-8 `$HOME` reported as "not set". Now `VarError::NotUnicode` gets an accurate message pointing at `~`.
- 🟢 **Stray `$` / no escape** (reviewers split; took Gemini's stricter line): `$$` escapes a literal `$`; any other stray `$` is a loud error, not a silent literal.

Both cross-family reviews (DeepSeek V4 Pro, Gemini 3.1 Pro) are now folded. Ready for merge.

PR #16, opened 2026-06-23
Read attachments through the kaish VFS — close the check-then-open TOCTOU

tobert:
## The window

`resolve_attachments` canonicalized + containment-checked a path, then read it with a *separate* `std::fs::read`. A concurrent writer could swap the path for an out-of-tree symlink between the check and the read — a classic check-then-open TOCTOU. Flagged by the batch `attach` cross-family review (DeepSeek, 2026-06-22). The attacker model is narrow (a self-attack on the caller's own workspace, sub-millisecond window), but the read-only boundary is meant to be **structural**, not honor-system.

## The fix

Keep the `std` canonicalize + containment check as the *friendly early error*, but move the **read** through the read-only kaish VFS — `worker.read_file(canon)` on a `KaishWorker` rooted at the attachment's containing tree, exactly how `view_image` already reads. The VFS re-resolves at read time and refuses to follow a symlink out of its mount (the `mount_layer_symlink_*` battery proves this), so a swapped escaping symlink is rejected regardless of timing. The window closes structurally — no `/proc`, no `O_NOFOLLOW` per-platform dance.

- New `containing_tree(canon)` returns which allowed tree (or followed worktree) holds a path; `contained()` is now `is_some()` over it, and that tree is the worker's mount root.
- One worker per **distinct** containing tree, reused across attachments under it (the common case — files under one project root — builds a single worker).
- `resolve_attachments` became `async` (both call sites already are).

## Teeth

`tests/attach.rs` proves a file in a *second* allowed tree reads only when the worker mounts the **correct** containing tree — verified by breaking tree selection and watching the read fail (a file outside the worker's mount is unreadable, which is the escape-refusal property). Per the project's no-flaky-tests stance, there's no racy TOCTOU test; the deterministic proof is structural (the VFS refuses escapes + the read goes through it). `resolve_root`'s lesser check-then-open (it hands the canonical path to the kernel as a mount root, working through an fd) stays open and is re-scoped in `docs/issues.md` as the lower-priority remaining half.

## Review

Per Amy's direction, the full source (`server.rs`, `sandbox.rs`, `tests/attach.rs`) is out for a careful unwind-and-analyze pass on a **Gemini Pro batch** (cross-family, different lineage). Findings will land as a follow-up commit here, as with the other hardening PRs. Full suite green (21 binaries), clippy clean.

Closes the attachment half of the path-containment TOCTOU item in `docs/issues.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,5 +131,11 @@ the git log. Each later release appends a new section at the top.
   output, 64 MB scratch — over-cap fails loudly, never a silent drop), and the model
   loops stop at turn limits, so a runaway consultation can't melt the machine or the
   budget. All configurable.
+- **Attachments are read through the read-only VFS.** A named attachment's bytes are
+  read *through* the same read-only kaish mount the shell uses (rooted at the file's
+  containing allowed tree), not a separate `std::fs` read after the containment check.
+  The VFS refuses to follow a symlink out of the allowed tree at read time, so a path
+  swapped for an out-of-tree symlink *after* the check is rejected structurally rather
+  than by racing a re-check — the boundary holds regardless of timing.
 
 [0.2.0]: https://github.com/tobert/kaibo/releases/tag/v0.2.0
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -172,6 +172,29 @@ otherwise sound. The attacker model stays narrow (a self-attack: a concurrent wr
 caller's own workspace, sub-millisecond window). Worth closing for symmetry when the model
 justifies it, but lower priority than the attachment read that one-shot-followed a symlink.
 
+### Attachment reads have no streaming/cumulative resource bounds (DoS, self-attack)
+Surfaced by the Gemini Pro batch review of the VFS-read change (2026-06-23). These are
+**pre-existing** (the old `std::fs::read` + `.collect()` path had them too) and all in the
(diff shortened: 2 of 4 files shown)


tobert:
## Gemini Pro batch review (cross-family, unwound the full source)

**Verdict on the core change: sound — the read-escape TOCTOU is genuinely closed.** A swapped out-of-tree symlink is trapped by the VFS mount and falls through to the empty `MemoryFs` `/` → `NotFound`. The review explicitly validated as **sound**: the HashMap-borrow-across-`.await` (temporary index borrow drops at statement end), the per-tree `KaishWorker` lifecycle (clean thread teardown on drop), and `containing_tree`'s precedence (static `allowed_set` before followed worktrees, matching the old `contained`).

**It also flagged three resource-exhaustion DoS vectors** — but all three are **pre-existing** (the old `std::fs::read` + `.collect()` path had them) and **self-attacks** (the caller owns the request and the workspace), so DoS not confidentiality, and outside this change's scope:
1. per-file size cap is check-then-read (`read_file` slurps unbounded → OOM via size-swap),
2. a FIFO swapped in can hang the worker read, and
3. no cumulative cap across N attachments.

Recorded as a follow-on "attachment resource bounds" pass in `docs/issues.md` (commit in this PR) rather than scope-creeping the TOCTOU fix. The streaming per-file cap is the one with real teeth and needs a kaish-vfs read API that bounds the read.

PR #23, opened 2026-06-25
Async consult: image attachments + fix wait returning too early

tobert:
This branch carries two changes to the async consult/`wait` surface: image attachments to `consult`, and a fix for `wait` returning too early.

---

### Why

`consult`'s `attach` was text-only by construction — it *names* files for the model to `cat -n`, and an image hits kaish's binary refusal. But a vision-capable consult already carries a `view_image` tool (the image-analog of `cat`), so an attached image just needed to be routed there instead of the shell. Two concrete bugs before this: the prompt told the model to `cat` *every* attached file (misdirects an image), and there was **no vision gate** — an image attached to a vision-blind consult was silently uncatchable, with no honest refusal.

### What

Route an attached image through the `view_image` tool the consult already has — keeping consult's "name it, the model opens it when ready" philosophy, and treating an attached image identically to one the model discovers in the repo. (Chosen over eagerly inlining the image as a vision part on the first turn: that always spends the image tokens and adds a second delivery path; view_image is lazy and already solves per-provider transport.)

Three parts:
1. **Classify by content.** The server sniffs each attachment's magic bytes — a PNG named `.txt` is still an image, matching how `view_image` re-sniffs authoritatively at read time.
2. **Gate on vision.** An image to a vision-blind synth is refused up front, naming the cast — the same honest refusal `oneshot`/`batch` give. consult could never show it, so naming the file would be a lie.
3. **Route in the prompt.** `consult_user_prompt` now splits its list: text files under a `cat -n` instruction, images under a `view_image` one — never `cat`-ing a binary.

`ConsultConfig.attachments` carries the classification (`Vec<ConsultAttachment>`), threaded through both `consult` and `consult_submit`.

### Tests

- content-based classification (PNG-as-`.txt` → image, UTF-8-as-`.png` → text)
- the vision gate (blind synth + image → refused; vision synth + image → ok; text-only always ok)
- prompt routing (image under view_image, text under cat -n)
- the `kaibo://tools` doc test gains `view_image` as a needle

Note: the default `deepseek` cast is vision-blind, so image attach there now refuses with guidance — use a vision cast (`anthropic`/`gemini`).

---

### Why

`wait(timeout_secs=600)` was returning near-instantly instead of blocking. The cause was a stale wake signal left in the notification ring:

- A finished async job pushes a **Warn** "job finished/failed" ping into the shared notification ring (`jobs.rs`) — the wake signal a parked `wait` returns on.
- `get` collected a job from the `JobStore` **without ever draining that ping**, so it lingered in the 512-cap ring.
- The next `wait` then drained the stale ping and returned immediately. Net effect: the *first* `wait` on a job blocked correctly, but every later one returned too fast on the previous job's ping.

The `wait` timeout/`select!` arithmetic was never at fault — the empty-timeout path is sound and tested.

### What

`get` now retires a terminal job's ping when it collects it (the ping is stale news once the job is in hand). New `NotificationBuffer::discard_job_pings(id)`, called from `get` only for `Done`/`Failed` jobs (a `Running` job has no ping yet; a `Canceled` one never emitted one). Scoped to the one job id, so an **uncollected** job's ping still wakes a later `wait`, and idempotent — a ping a live `wait` already drained leaves nothing to remove.

### Tests

- the buffer method drops only the named job's ping, leaving others (incl. job-less Warns) intact
- `get` on a terminal job retires its ping while an uncollected job's survives — the integration seam where a regression would hide (the assertion compares against the surviving ping, so it fails if the eviction call is dropped)

---

Full suite green (272 lib tests, 0 failures).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,12 @@ the git log. Each later release appends a new section at the top.
   and per-call `explorer_model` / `synth_model` (+ `_backend`) overrides. **`attach`**
   names workspace files (under the project root) to put in front of the investigation —
   unlike the tool-less tools' attach, kaibo does *not* inline them; it names them in the
-  prompt and the consult model reads them itself with the shell when it's ready, in full,
-  building its own narrative. "Review this diff" is `attach: ["changes.diff"]` with no
-  paste; the files just have to live under the root the consult reads (a worktree counts).
+  prompt and the consult model opens each itself when it's ready, in full, building its own
+  narrative: a text file with the shell (`cat -n`), an **image** with its `view_image` tool.
+  An attached image therefore needs a vision-capable cast — kaibo refuses one to a
+  vision-blind synth up front (the same honest refusal `oneshot`/`batch` give) rather than
+  name a file the model could never open. The files just have to live under the root the
+  consult reads (a worktree counts).
 - **`consult_submit`** — the *async sibling* of `consult` (as batch is to `oneshot`):
   start a consultation in the background and get back a handle (`job-N`) instead of
   holding your turn open while a deep investigation runs. Same investigation, same args
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -51,13 +51,59 @@ is the workflow layer. Rationale recorded here so it survives the conversation.
 - **Media moves by VFS path; base64 only at the edges** (provider wire, MCP
   content). Scratch `MemoryFs` is the working bus — bounded now that kaish landed
   `ByteBudget` (kaish-vfs, 2026-06-10; kaibo pickup tracked in the P2 entry below).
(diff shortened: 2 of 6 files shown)


tobert:
**Scope note (added before merge):** this branch now also carries two commits beyond the consult-image work, folded in by request rather than split to their own PR:

- \`21b2234\` — *Retire a collected job's wait ping so wait stops returning early* (code: \`src/mcp_log.rs\`, \`src/server.rs\`). A behavior fix to the still-unreleased \`wait\` tool — once a job is collected, \`wait\` no longer returns early on its stale ping.
- \`0daa635\` — *Record the RW-mount design, superseding inline-only --out-dir* (\`docs/issues.md\` only).

Full suite green (21 binaries). The wait-ping change did not get the separate cross-family review the consult-image commits did — folded in at the author's direction.

PR #24, opened 2026-06-25
Gate batch casts to the batch lane; rename openai built-in to openai-local

tobert:
Two related cast-semantics changes, folded into one PR at Amy's request.

## Why

A `gemini-batch` cast run through `consult_submit` did the wrong thing: it ran the offline-tuned Pro synth through the **interactive** tool loop — slow and expensive — while getting **none** of the provider batch API. The root cause: "batch" was never a property of a cast, only a naming convention, and `gemini-batch`'s own comment invited consult use ("usable for consult too").

### Batch casts are a declared, enforced lane

- A cast declares its lane positively with **`batch = true`**.
- **Mixing is verboten both ways** (Amy's call — an accidental Opus/Pro batch is just as costly in the other direction, especially once Fable returns):
  - `consult` / `consult_submit` / `oneshot` / `generate_image` refuse a batch cast.
  - `batch_submit` refuses a non-batch cast.
  - Both gates fire right after the cast resolves, **before any network**, naming the right tool.
- The per-tool `cast` **enums split to match**: interactive tools advertise non-batch casts; `batch_submit` advertises only batch casts — advertised menu and runtime gate agree.
- Batch built-ins carry **synth only** (batch is a toolless oneshot): `gemini-batch` loses its now-dead explorer, and a new **`anthropic-batch`** ships (synth Opus, the Anthropic analogue of Gemini Pro).
- A `batch = true` cast whose synth backend has **no batch API** is a loud config-load error, not a 400 at submit time.

### `openai` built-in renamed to `openai-local`

The built-in `openai` connection defaults to a *local* keyless server, so the bare name misrepresented it.

- Now **`openai-local`**, via a new `ProviderKind::builtin_name()` that diverges only for Openai.
- The wire-protocol id stays `openai` — `canonical_name()`, the `kind = "openai"` field, and `FromStr` are unchanged.
- `openai` is **deliberately not an alias**, so a user can name their own `[backends.openai]` (the hosted API) without a collision. Aliases stay `local` / `lemonade` / `gemma` / `gemma4`.

### `consult_submit` description fix + a text-for-models doctrine

Tagging along because it's the same surface (the async tool descriptions) and the same review.

- A Claude Code client called `consult_submit` three times, then wrapped the calls in the host's own `run_in_background` — because the description led with "Start a `consult` in the background … to collect later". The missing signal was that **the call returns at once and kaibo already does the backgrounding**. New opener: "Submit a `consult` and get a `job-N` handle back immediately — kaibo runs it in the background." "immediately" reclaims the dead "to collect later", so the block stays ~flat; also surfaces `wait` ahead of `get`. `batch_submit` is left untouched — its "Fire-and-forget" + durable cross-restart handle already carry it.
- AGENTS.md gains a **Writing for models** section: re-read whole blocks holistically, ratchet toward compression (改善). Two audiences optimized differently — client-facing tool descriptions are *resident* per-session cost (they load whether or not the tool is ever called, so density is existential), agent-facing preambles run on large-context models where verbosity is licensed where it shapes behavior.

## Testing

- Offline suite green: **279 lib + every integration suite**.
- The gate's teeth are proven by a failing-first test (neutering it lets a real batch submit through).
- Both gate directions were **dogfood-verified live** on the rebuilt binary: `oneshot` + `anthropic-batch` and `batch_submit` + `deepseek` both refuse with the right pointer; the lane-split enums show on the live tool schemas.

## Review notes

Per repo discipline this wants a **cross-family** review of the diff (a different lineage than wrote it). The sensitive parts are the gate wiring in `server.rs` and the name-vs-kind separation in the `openai-local` rename (`kind = "openai"`, the rig source path `providers/openai/…`, and `api.openai.com` all correctly stay `openai`).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -139,6 +139,28 @@ Local musl repro (no system zig needed): `pip install ziglang` in a venv,
 x86_64-unknown-linux-musl`. Verify the boundary with `cargo tree -i aws-lc-rs`
 (empty) and `ldd target/.../kaibo` (`not a dynamic executable`).
 
+## Writing for models
+
+Every prompt, preamble, tool description, and cheatsheet is text some model reads.
+When we touch one, we **re-read the whole block and judge it holistically** — not the
+diff in isolation. The ratchet is *compression* (改善): over time a block should carry
+more impact per token and avoid accreting clauses.
+
+Two audiences are optimized differently:
+
+- **Client-facing text** — the MCP server instructions and each tool's `description`
+  in `server.rs`, read by the *calling* agent. Density is existential here because it's
+  *resident* cost: the descriptions load into the caller's context every session
+  whether or not the tool is ever called — an unused tool still bills the user. We ask
+  people to spend a few tokens on kaibo in every session they connect us; each one has
+  to earn its place and be additive to their experience. Terse, concrete, no clause
+  that doesn't pay rent.
+- **Agent-facing text** — preambles, the kaish cheatsheet, the casts' answering
+  instructions, read by the commercial models *we* drive, with windows in the hundreds
+  of thousands to millions of tokens. Here verbosity is *licensed where it shapes
+  behavior*: say it a few ways, frame positively, be explicit (see **Driving the
+  models**). Verbose to install behavior, never verbose by default.
+
 ## Driving the models
 
 How kaibo talks to LLMs — Amy's defaults, made local so any agent here inherits them.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,9 +69,15 @@ the git log. Each later release appends a new section at the top.
   no state: the handle is the whole address, so poll/cancel survive a restart, and a
   failed item is surfaced per-item rather than dropped. Runs on **Anthropic and Gemini**
   backends (OpenAI batch is a tracked follow-on); a cast whose synth has no batch lane is
-  refused with a clear message naming the ones that do. For Gemini there's a ready-made
-  `gemini-batch` cast that synths Gemini **Pro** — the tier you reach for offline, where
-  its latency is free. Gated by `--no-batch` (one flag over every verb). Batch carries its
+  refused with a clear message naming the ones that do. Two ready-made batch casts ship:
+  `gemini-batch` (synth Gemini **Pro**) and `anthropic-batch` (synth Claude **Opus**) —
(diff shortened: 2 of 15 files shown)


tobert:
## Cross-family review (dogfooded via kaibo `consult`, cast `deepseek`)

Ran the repo's cross-family review by pointing kaibo's own `consult` at the diff with the **deepseek** cast (a different lineage than wrote this — DeepSeek explorer + synth). Verdict: the core logic is **sound**, with leftover cosmetic/name spots since fixed in `f2939ba`.

**Confirmed sound:**
- The lane gate fires *before* any network/key work in all five tools (cast resolve → gate → model-override → arm/submitter); `apply_model_override` and `image_backend` can't flip `batch`, so no bypass.
- Load-time validation (`batch = true` ⇒ synth on a batch-capable backend) keys off the single `batch_supported` source of truth.
- The enum split and the sticky-over-builtin `batch` merge are correct.
- The `openai` → `openai-local` name/kind separation is clean — `canonical_name()` (protocol id), `kind = "openai"`, `FromStr`, base-URL/key/alias infra all correctly stay on `openai`.

**Fixed from the review (`f2939ba`):**
- `main.rs` `--cast` help still listed `openai` (user-facing — `--cast openai` would error) → `openai-local`.
- New-openai-backend-without-`base_url` load error + comment said "built-in `openai` backend" → `openai-local` (the kind check was already correct).
- Two stale test doc-comments → `openai-local`.
- `generate_image`'s menu (`image_capable_casts`) didn't exclude batch casts, so a batch cast with an image slot would be advertised yet gate-refused → now excluded, menu matches the gate, with a test that has teeth.

tobert:
## Review trail (pre-merge)

**Cross-family review** — ran the repo's required cross-family pass by dogfooding kaibo's own `consult` with the **deepseek** cast (a different lineage than wrote this). Detailed findings in the comment above; verdict was *logic sound*, with five cosmetic/name leftovers and one menu inconsistency, all fixed in `f2939ba`.

**Follow-up (`5d59327`)** — after the review, codified the "Writing for models" doctrine in `AGENTS.md` and tightened the `consult_submit` tool description so calling agents don't re-background already-async work (a resident-cost prompt the caller reads every session).

Offline suite green; both gate directions dogfood-verified live on the rebuilt binary. Merging.

PR #28, opened 2026-06-26
generate_image: write artifact to out-dir, return the path

tobert:
`generate_image` no longer streams a base64 blob inline — it writes the image to a kaibo-owned **out-dir** and hands back the absolute path. A multi-MiB picture costs the caller's context nothing until it opens the file, and the artifact gets a durable home. This is the narrow *specific-tool* write the [RW-mount deferral](../blob/main/docs/issues.md) pointed at — not the broad `rw_paths` surface, and not `consult`.

## Decisions (from the conversation with Amy)

- **Path, not `ResourceLink`.** A `ResourceLink` would make a spec-compliant client read the bytes back *from the server*, growing a `file://` `resources/read` channel that bypasses project containment — a new read surface — for no win over a plain path the agent opens itself. Dropped.
- **Handler-side write, never kaish.** `write_artifact` uses `std::fs` in the tool handler; all four read-only sandbox levers in `sandbox.rs` hold untouched — kaish still cannot write anywhere. The invariant gained a sentence, it did not move.
- **Read-back mount.** The out-dir is mounted **read-only** into every kaish kernel (the existing `VfsRouter` multi-mount, *not* the deferred general-RW machinery) so a later `consult`/`run_kaish` can read a generated artifact back. Forward substrate for image2image; nothing needs it yet.
- **Default `$XDG_CACHE_HOME/kaibo`, kaibo-owned on purpose.** The read-back mount widens read-*scope* to whatever the out-dir is, and a consult can ship what kaish reads to a remote model. Mounting bare `/tmp` would expose other processes' files; a kaibo-owned subdir keeps the scope to our own artifacts. Durable by choice (cleanup revisited only if users ask).
- **Inline cap removed.** `GENERATE_IMAGE_MAX_BYTES` only bounded base64 in context; path delivery has nothing to bound.

## Review focus

The one part that warrants a hard look is the **read-scope boundary moving** — even narrowly, to a kaibo-owned cache. The new read-only out-dir mount (`sandbox.rs::build_readonly_kernel_and_vfs`) and its containment behavior are where to look. Pinned by the `out_dir_*` battery in `tests/sandbox.rs` (artifact readable, out-dir read-only to kaish, **siblings not exposed**), `out_dir` precedence in `tests/config.rs`, and Battery E in `docs/sandbox-probes.md`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,10 +61,20 @@ project and cannot run external commands.
   disk), and (3) external commands disabled. The `Blocked` wrapper survives only for
   the config-driven `[sandbox].disable_builtins`, which can make the box *stricter* —
   see the module doc-comment. Any change here keeps `tests/sandbox.rs` green and adds
-  a test that can fail. Read-*scope* is also
-  bounded: every call's path must canonicalize (symlinks, `..` resolved) into the
-  allowed set (`--root` / `--allow-path`, launch cwd when unset), enforced in
-  `server.rs::resolve_root` with tests in `tests/containment.rs`.
+  a test that can fail. The one *write* path is a **capability tool writing its own
+  artifact**, handler-side via `std::fs`, only to the kaibo-owned **out-dir**
+  (`config.out_dir`, default `$XDG_CACHE_HOME/kaibo`) — never through kaish, so all four
+  levers above are untouched: kaish still cannot write anywhere. (`generate_image` is
+  the only such tool today.) Read-*scope* is also bounded: every call's path must
+  canonicalize (symlinks, `..` resolved) into the allowed set (`--root` /
+  `--allow-path`, launch cwd when unset) **plus the out-dir**, which is mounted
+  *read-only* into kaish so a consult can read a generated artifact back — a deliberate,
+  narrow widening to kaibo's own cache (a consult *can* ship those own-generated
+  artifacts to a model; that's expected, not a leak). Enforced in
+  `server.rs::resolve_root` and the out-dir mount in
+  `sandbox.rs::build_readonly_kernel_and_vfs`, with tests in `tests/containment.rs` and
+  the `out_dir_*` battery in `tests/sandbox.rs` (artifact readable, out-dir read-only to
+  kaish, siblings of the out-dir *not* exposed).
 - **stdio only.** kaibo can read a filesystem, so it must never bind a socket.
 - **kaish is `!Send`.** The kernel runs on a dedicated thread behind `KaishWorker`;
   rig tools require `Send` futures. Don't hold the kernel across an `.await`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -67,10 +67,12 @@ project and cannot run external commands.
   levers above are untouched: kaish still cannot write anywhere. (`generate_image` is
   the only such tool today.) Read-*scope* is also bounded: every call's path must
   canonicalize (symlinks, `..` resolved) into the allowed set (`--root` /
-  `--allow-path`, launch cwd when unset) **plus the out-dir**, which is mounted
-  *read-only* into kaish so a consult can read a generated artifact back — a deliberate,
-  narrow widening to kaibo's own cache (a consult *can* ship those own-generated
-  artifacts to a model; that's expected, not a leak). Enforced in
+  `--allow-path`, launch cwd when unset) **plus the out-dir** when `out_dir_readable`
+  (default on): the out-dir is mounted *read-only* into kaish **and** joined to the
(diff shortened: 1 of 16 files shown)


tobert:
## Cross-family review trail

Per the AGENTS.md "dogfood the review — cross-family pass before merge" discipline, this diff was reviewed by **three runs against other model lineages, driven through kaibo itself** (a different house than the Opus that wrote it):

| # | Tool | Cast / model | Focus | Verdict |
|---|------|--------------|-------|---------|
| 1 | `consult_submit` (async) | **DeepSeek** (`deepseek-v4-pro` synth) | Investigated the live worktree code: read-only invariant, mount logic, write path, config layering | **No blocker** — read-only/read-scope claim holds |
| 2 | `batch_submit` | **Gemini Pro** (`gemini-pro-latest`) | Static review of the diff: containment, TOCTOU/symlink, config sync, half-write | **No blocker** — claim holds, config in sync, no panics |
| 3 | `batch_submit` (11 files attached) | **Gemini Pro** | Scenario-driven play-through + the safe-default design question | Confirmed a real **symlink-exfil hole** in the temp fallback |

All three independently confirmed the core invariant: **kaish can't write anywhere, read-scope stays structurally bounded** (project + the kaibo-owned out-dir when readable). The read-scope *widening* — the part that moved a guarded boundary — got the explicit hard look the project asks for on containment changes.

tobert:
## Review findings → fixes

Each actionable finding from the reviews was addressed; mapping for the trail:

**DeepSeek (review 1)**
- *Functionality gap:* the read-back mount is decided at kernel-build time and skipped if the out-dir doesn't exist yet — a consult session started **before** the first `generate_image` kept a worker without the mount. → **`fb51bcb`** pre-creates the out-dir at startup so the mount is reliably present.

**Gemini Pro (review 2)**
- *Symlinked out_dir path mismatch:* `write_artifact` returned the un-canonicalized path while the mount is keyed on the canonical one. → **`048664f`** canonicalizes the returned path.
- *Defense-in-depth:* `out_dir = "/"` would read-mount the whole host. → **`048664f`** refuses `/` at load, warns on `$HOME`.
- *Re-attach gap (out-dir not in `allowed_set`):* flagged as a wart. → resolved by Amy's steer in **`be1ef38`** (make it readable, with an off switch).

**Gemini Pro scenario play-through (review 3)**
- *Critical: symlink-exfil in the temp fallback.* No `$XDG_CACHE_HOME`/`$HOME` (a container) → kaibo auto-mounted a world-shared temp; a pre-planted `<tmp>/kaibo` symlink → `/etc` or a home dir would be canonicalized, mounted read-only, added to `allowed_set`, and a consult could exfiltrate it (the `/` guard misses a symlink to a non-`/` target). → **`cf171a6`** classifies the default (`Cache` vs `SharedTemp`) and **defaults read-back OFF for the shared-temp fallback** — still writes + returns the path, but won't auto-mount a shared temp. Plus a startup warning and primed `configure`-prompt prose.
- *S2 misleading error:* an out-dir attach with read-back off suggested "use oneshot/batch," which also can't reach it. → **`700c4b6`** gives a tailored refusal naming the real fix (`out_dir_readable`), no false workaround.

Net: the one **critical** finding (symlink-exfil) and every actionable nit are fixed; all changes carry failing-first tests.

tobert:
## Tracked follow-ups (not blockers)

Deliberately deferred, recorded in `docs/issues.md` so they're not lost:

- **Write-side symlink on the out-dir** (P2). `write_artifact` canonicalizes the out-dir before writing, so a symlinked out-dir lands the artifact at the symlink's target. In a world-shared temp this is an arbitrary file *drop* — but it's a **write** nuisance, not the read-exfil (that's closed: read-back defaults off for the shared-temp fallback), and `unique_artifact_name` prevents overwriting an existing file. Hardening if it ever matters: refuse a symlinked out-dir component, or `O_NOFOLLOW` the final create. Low priority — the read side was the real hole.

- **Flaky `omitted_path_zero_config_infers_cwd_as_default_root`** (pre-existing P2, untouched here). A parallel-execution race on the shared process cwd — passes in isolation and on re-run. The one intermittent failure you may see in a full `cargo test`; not introduced by this work.

### Possible next step (separate PR)
- **Cut a release?** This lands `generate_image` artifact delivery end-to-end. Per AGENTS.md the release checklist (re-run `docs/sandbox-probes.md` incl. the new Battery E, bump version, `cargo tree -i aws-lc-rs` empty, musl `not a dynamic executable`) would be its own pass.

Notes: the repo isn't rustfmt-clean, so only edited lines were formatted (no repo-wide fmt). clippy is clean apart from one pre-existing `telemetry.rs` "complex type".

PR #32, opened 2026-06-29
Advertise the live cast roster in the resident tool descriptions

tobert:
## Why

When a user says "have deepseek review this," their agent needs the name `deepseek` to sit somewhere its tool-picker actually reads — the resident tool **description** — to index from that phrasing to a kaibo tool. We already stamp the usable cast names onto the `cast` *parameter* as a JSON-Schema enum (`inject_cast_enum`), but a host's tool-picker leans on description prose and may skim past param schemas (the same reason the enum injection exists: the handshake list a host may truncate). So the names weren't reliably resident where it counts.

## What

`append_cast_roster` (sibling to `inject_cast_enum`, built at startup from the same usable set, same batch/interactive lane split, same empty-roster guard) appends `Casts ready now: deepseek (default), gemini, …` to the resident descriptions of `consult`/`consult_submit`/`oneshot` (interactive roster) and `batch_submit` (batch roster), default flagged.

Because the name now reaches the client, naming a cast for *intent* (`local-only`, `deep-dive`) is what lets a host route by name without reading the config — documented at each cast-definition site.

## Decisions

- **Density split for the docs.** The enum / "Casts ready now" *mechanics* are kept in the design doc (`docs/casts.md`) and `CHANGELOG`; the resident user-facing texts (`README`, `config.example.toml`) are purely descriptive — "the calling agent sees your cast names." Repeating implementation mechanics in resident text doesn't pay rent (per AGENTS.md "Writing for models").
- **Folded in an example-config fix.** A review pass cross-checking `config.example.toml` against `config.rs` caught a real bug — `[server.tools]` omitted `batch` and said "three" when the all-off refusal needs all *four* false — plus several features the example never demonstrated. Fixed the bug and added worked examples: a `batch = true` cast (`pro-batch`), `[prompts].batch`, backend `aliases`, and the slot tunables `temperature`/`thinking_budget`/`thinking_style`. All validated by the real loader via `the_shipped_example_config_parses`.
- **Two clarity leftovers recorded, not fixed** (`docs/issues.md`, P3): the illustrative backend stanzas read as required, and `gpt` names both a backend and a cast without the file noting the namespaces are separate.

## Review

Dogfooded a cross-family pass *through kaibo itself* — `consult` (deepseek) over the diff, a second `consult` (anthropic) cross-checking the example against the parser, a `batch` (gemini-pro) over source + docs, and a deepseek/gemini/anthropic parse-back of the example config. Findings and the changes they drove are posted as PR comments. The two test gaps the code reviewers flagged (batch lane, empty guard) are closed in `b83935a`.

## Tests

- `roster_splits_by_lane_in_descriptions` — mirrors `cast_enums_split_by_lane` for descriptions (interactive tools name the interactive cast and never a batch cast; `batch_submit` the reverse).
- `empty_roster_leaves_descriptions_unchanged` — the empty-guard, byte-for-byte.
- Full suite green (the lone intermittent failure is the known tracing-callsite flake, unrelated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,11 @@ the git log. Each later release appends a new section at the top.
   with a hosted synth. Built-in casts ship so
   kaibo runs with zero config; `config.toml` merges over them. Precedence:
   per-call > CLI > env > file > built-in, and a missing config file is not an error.
+  Your usable casts' names are advertised to the *calling* agent — both as the `cast`
+  param's enum and in the consult/oneshot/batch tool descriptions ("Casts ready now:
+  …", with the default flagged) — so a host told "have deepseek review this" routes off
+  the roster, and a meaningful name (`local-only`, `deep-dive`) reads as intent without
+  the caller opening your config.
 - **Guided setup.** A built-in `configure` MCP prompt walks your host agent through
   writing `config.toml`, alongside `kaibo://config` (resolved runtime state) and
   `kaibo://config/example` (annotated template) resources. Secrets are referenced by
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,10 +125,10 @@ the git log. Each later release appends a new section at the top.
   kaibo runs with zero config; `config.toml` merges over them. Precedence:
   per-call > CLI > env > file > built-in, and a missing config file is not an error.
   Your usable casts' names are advertised to the *calling* agent — both as the `cast`
-  param's enum and in the consult/oneshot/batch tool descriptions ("Casts ready now:
-  …", with the default flagged) — so a host told "have deepseek review this" routes off
-  the roster, and a meaningful name (`local-only`, `deep-dive`) reads as intent without
-  the caller opening your config.
+  param's enum and in the tool descriptions ("Casts ready now: …", with the default
+  flagged) — so a host told "have deepseek review this" routes off the roster, and a
+  meaningful name (`local-only`, `deep-dive`) reads as intent without the caller opening
+  your config.
 - **Guided setup.** A built-in `configure` MCP prompt walks your host agent through
   writing `config.toml`, alongside `kaibo://config` (resolved runtime state) and
   `kaibo://config/example` (annotated template) resources. Secrets are referenced by
--- a/README.md
+++ b/README.md
@@ -129,7 +129,9 @@ config has three concepts for configuring models:
 - **role** — a *job* a model does: `explorer` (fast sweeps) and `synth` (the voice that
   answers). A slot that reads images carries a `vision` pin (see [`docs/casts.md`](docs/casts.md)).
 - **cast** — a *composition*: a named team assigning models to roles. The `cast` call
-  argument selects the ensemble.
+  argument selects the ensemble. The name is advertised to the *calling* agent — on the
+  `cast` param and in the tool descriptions ("Casts ready now: …") — so a meaningful name
+  (`local-only`, `deep-dive`) lets it pick one by intent without reading your config.
 
 ```toml
 # ~/.config/kaibo/config.toml
--- a/README.md
+++ b/README.md
@@ -129,9 +129,9 @@ config has three concepts for configuring models:
 - **role** — a *job* a model does: `explorer` (fast sweeps) and `synth` (the voice that
   answers). A slot that reads images carries a `vision` pin (see [`docs/casts.md`](docs/casts.md)).
 - **cast** — a *composition*: a named team assigning models to roles. The `cast` call
-  argument selects the ensemble. The name is advertised to the *calling* agent — on the
-  `cast` param and in the tool descriptions ("Casts ready now: …") — so a meaningful name
-  (`local-only`, `deep-dive`) lets it pick one by intent without reading your config.
+  argument selects the ensemble; the calling agent sees these names in its tool listing,
+  so a descriptive name (`local-only`, `deep-dive`) lets it pick a team by intent without
+  reading your config.
 
 ```toml
 # ~/.config/kaibo/config.toml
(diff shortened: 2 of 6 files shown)


tobert:
# Cross-family code review — DeepSeek + Gemini-pro

Two independent reviews of the `append_cast_roster` change, run *through kaibo itself*. They converged: the implementation is correct and the mirroring of `inject_cast_enum` is tight; the ~40 resident tokens are justified because host tool-pickers rank on description text, not param enums; and both named the **same two test gaps** — the batch lane and the empty-roster guard. Both gaps are now closed in `b83935a` (`roster_splits_by_lane_in_descriptions`, `empty_roster_leaves_descriptions_unchanged`).

---

## DeepSeek (`consult`, deepseek-v4-pro over the diff)

**Correctness:** empty-guard mirrors `inject_cast_enum` exactly; default-cast flagging is per-lane honest (a batch-only default simply won't match on the interactive lane); the `Option<Cow>` description mutation preserves the original prose and correctly produces `Cow::Owned`. Lane split matches `inject_cast_enum` with no drift.

**Resident cost:** ~60–100 tokens total across four tools; earns its place because the param enum is buried where a host's tool-picker won't rank it. Wording is terse enough; leave it.

**Test gaps (now fixed):**
1. *Batch lane untested* — the roster test's `batch_usable` was empty, so `batch_submit`'s description went unchecked. The path that would regress if someone refactors the lane split.
2. *Empty guard untested* — `inject_cast_enum` has `empty_cast_roster_leaves_the_param_unconstrained`; `append_cast_roster` had no equivalent.
3. *(minor)* assert the original description survives the append.

> Verdict: "The function and its integration are correct and well-mirrored. The token cost is justified… two concrete gaps: batch lane coverage and empty-guard coverage. I'd add both before merging."

---

## Gemini-pro (`batch`, gemini-pro-latest over source + docs)

Corroborated the above point-for-point — correctness, the lane-split invariant, and the **same two missing tests** (batch lane on `batch_submit`, and an empty-guard test). Its framing of the cost question:

> "Host agents often use semantic search or standalone evaluation of tool descriptions to decide which tool to pick… Relying on the JSON-Schema `enum` alone is notoriously brittle for LLM tool selection. Appending this to all four tools costs ~40 tokens globally but buys reliable, intent-based routing."

One extra, theoretical: if a tool ever had an empty base description the format string would leave a leading space. Can't happen — all four carry `#[tool(description = …)]` — so left as-is.

tobert:
# Docs & example-config review — Anthropic + Gemini-pro

## Anthropic (`consult`, sonnet-4-6) — cross-checked `config.example.toml` against `config.rs`

Found one real **bug** and several undemonstrated features. The bug and the four gaps are fixed in `4c1c5e6`; two clarity items are recorded in `docs/issues.md`.

| # | Issue | Disposition |
|---|---|---|
| **A** | `[server.tools]` omits `batch`, and the comment says "three" when `all_disabled()` needs all **four** false (`config.rs` `RawTools`/`ToolGating`). An operator copying the block could leave batch running. | **Fixed** — added `batch = true`, corrected the count |
| B | Cast `batch = true` never shown (`RawCast.batch`) | **Fixed** — added `[casts.pro-batch]` |
| C | `[prompts]` missing the `batch` phase (`RawPrompts.batch`) | **Fixed** — added |
| D | Backend `aliases` never demonstrated (`RawBackend.aliases`) — looked cast-only | **Fixed** — added `aliases = ["oai"]` on `[backends.gpt]` |
| E | Slot tunables `temperature`/`thinking_budget`/`thinking_style` enumerated but never shown at slot level | **Fixed** — added to the deep-dive explorer |
| F | Header says "four casts"; the registry builds six (four interactive + two batch) | **Fixed** — corrected the count |
| G | The cast-name note said "consult/oneshot" but the roster also lands on `consult_submit`/`batch_submit` | **Fixed** in `b83935a` |
| H | "same as today" is stale filler in the `[server.tools]` comment | **Fixed** |
| I | Illustrative built-in backend stanzas shown uncommented → read as required | **Recorded** (issues.md, P3) |

## Gemini-pro (`batch`) — docs prose pass

Flagged **oversaturation**: the enum / "Casts ready now" *mechanics* were repeated across all four doc sites. Its rule — and the one applied here — is that in the resident user-facing texts (`README`, `config.example.toml`) the mechanics don't pay rent; the reader only needs to know the calling agent sees the names.

> "Explaining *how* the name reaches the agent belongs in the design record (`docs/casts.md`) and the feature announcement (`CHANGELOG.md`). In the resident user-facing texts… the implementation mechanics are rent-free clauses."

Applied in `4c1c5e6`: `README` and `config.example.toml` are now descriptive; the mechanics stay in `docs/casts.md` and the `CHANGELOG`. It also independently caught accuracy item **G** above.

tobert:
# Config-example parse-back — DeepSeek / Gemini / Claude

A comprehension test: the same `config.example.toml`, read **cold** (no other context) by one model from each family via `oneshot`, asked to resolve every backend and cast, follow aliases, and report anything ambiguous. The point was to test how legibly the example reads to the calling models themselves.

**All three resolved it correctly** — every backend `kind`, every cast's explorer/synth (following backend aliases), the `chimera` families, `deep-dive`'s alias + `effort`/`max_tokens` tunables, and the cast-name-visibility note. No misreads of structure.

The useful signal was part (f), "one thing easy to misread" — three distinct, real ambiguities, all in **pre-existing** parts of the example:

- **Gemini** — backend aliases appear only in comments (`# alias: claude`) while cast aliases use an `aliases = [...]` key; a reader might think backend aliases aren't TOML-definable. → **Fixed** in `4c1c5e6` by adding a worked `aliases = ["oai"]` on `[backends.gpt]`.
- **Claude** — `gpt` names both `[backends.gpt]` and `[casts.gpt]`; the file doesn't say whether sharing a *primary* name across the two namespaces is intentional or a latent collision. → **Recorded** (issues.md, P3).
- **DeepSeek** — the `OPENAI_BASE_URL` ↔ `openai-local` `base_url` fallback exclusivity reads as if it might override any openai-kind backend. → noted; minor.

Net: the example parses well across families; the one actionable clarity gap (backend-alias asymmetry) is now closed, the rest tracked.

PR #35, opened 2026-06-29
feat(handshake): name each cast's synth model in the ## Casts roster

tobert:
## What

The startup handshake's `## Casts` roster now renders each line as `- \`name\` (tags) → backend/synth-id`, naming the cast's **answering (synth) model** and tagging a batch-only cast `batch`:

```
- `anthropic` (default) → anthropic/claude-sonnet-4-6
- `gemini` → gemini/gemini-3.5-flash
- `gemini-batch` (batch) → gemini/gemini-pro-latest
- `zorak` → gemma-26b/gemma4-26b
```

## Why

Follows the Gemini-Pro-batch-only config change (#34). With Pro reachable only via the `gemini-batch` cast, an agent told "ask Gemini Pro" still had no way to *find* it from the handshake — the roster listed bare cast names, so "pro" was invisible until the agent read `kaibo://config`. Now `gemini-batch → gemini/gemini-pro-latest (batch)` makes Pro self-evident **and** correctly routed to the `batch_submit` lane.

The data was already there: every `Cast` carries its `slots` map, `usable_casts` already iterates the full struct, and the synth id is exactly what `kaibo://config` prints. This is pure rendering — `casts_section` now takes `&Config` and resolves `slot(Synth)` + `cast_is_batch` per line.

## Scope — where it does *not* go

Deliberately confined to the **read-once handshake** block. The per-tool `description` roster (`append_cast_roster`) and the `cast` param `enum` stay **name-only**: those are *resident* cost — loaded into every caller's context every session whether or not a tool fires — and per AGENTS.md that's the existential-density surface. Model ids per cast on every description is exactly the accretion we avoid. The enum values must also stay bare cast names (they're the accepted call args).

## Test

Failing-first `casts_section_names_each_synth_model_and_tags_batch` (kaish_syntax.rs): asserts the rendered handshake names `anthropic/claude-sonnet-4-6`, names `gemini/gemini-pro-latest`, and tags the `gemini-batch` line `batch`. Watched it fail on bare-name rendering, then pass. Full suite green (280 lib + integration).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,7 +131,10 @@ the git log. Each later release appends a new section at the top.
   param's enum and in the tool descriptions ("Casts ready now: …", with the default
   flagged) — so a host told "have deepseek review this" routes off the roster, and a
   meaningful name (`local-only`, `deep-dive`) reads as intent without the caller opening
-  your config.
+  your config. The startup handshake's `## Casts` roster goes further: each line names
+  the cast's **answering (synth) model** and tags a batch-only cast `batch`, so a host
+  told "ask Gemini Pro" indexes `gemini-batch → gemini/gemini-pro-latest (batch)` — and
+  knows it's the `batch_submit` lane — without reading `kaibo://config`.
 - **Guided setup.** A built-in `configure` MCP prompt walks your host agent through
   writing `config.toml`, alongside `kaibo://config` (resolved runtime state) and
   `kaibo://config/example` (annotated template) resources. Secrets are referenced by
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -210,6 +210,15 @@ the git log. Each later release appends a new section at the top.
 - **Single self-contained binary** per platform; Linux builds are fully static
   (musl). TLS is rustls + ring — no OpenSSL, no aws-lc, no C toolchain.
 
+### Fixed
+
+- **The advertised cast roster marks the default even when it's set by an alias.**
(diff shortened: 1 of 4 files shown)


tobert:
## Cross-family review done — and it found a bug

Per the dogfood discipline, ran kaibo on this change with two families, neither of which wrote it (Claude did), no diff supplied so they'd judge holistically:

- **Gemini Pro** (via `batch_submit`, cast `gemini-batch` — the very Pro-on-batch routing this PR makes visible), whole files attached.
- **DeepSeek** (`consult_submit`, agentic — read the branch itself).

**Consensus:** correct rendering across cases (default / local-unverified / batch / synthless / canonical names), scoping sound (handshake enriched, resident roster + enum stay name-only), synth-only the right call, new test has teeth. The `→` (U+2192) is a single `cl100k_base` token, FWIW.

**Gemini caught a latent bug DeepSeek missed** — the `default_cast` *alias* trap: setting `server.cast = "claude"` (alias for `anthropic`) silently dropped the `(default)` tag, because the tag compared the raw string against the canonical names `usable_casts` yields. Pre-existing in `append_cast_roster` too; the new synth-model line inherited it.

**Fixed in `fe9bf1d`:** resolve the default before comparing, via a shared `Config::is_default_cast` helper used by both renderers. Failing-first tests at the config level (`is_default_cast_matches_through_an_alias`) and through the rendered handshake (`casts_section_marks_the_default_even_when_set_by_alias`), plus DeepSeek's suggested backfill for the synthless `None` arm. New `Fixed` CHANGELOG entry. Full suite green (283 lib + integration).

PR #43, opened 2026-07-02
feat(prompts): a Phase seam, kaibo://prompts, and the offline-synth override fix

tobert:
## Why

Two things came out of reading how the deliberate PR wired prompts:

1. **A per-slot synth `preamble` silently never reached the offline synth.** `resolved_prompts` folded a synth slot's `preamble` into the `consult`/`oneshot` keys but not `batch`, and both offline read sites (`batch_submit`, `deliberate`) read the *global* `self.config.prompts.batch` directly — bypassing `resolved_prompts` entirely. So a per-cast voice set on a `batch`/`deliberate` cast framed nothing it actually answered.
2. **We had no way to see the prompts models get.** The composition already funnelled through `phase_preamble`, but the *decisions* around it (which built-in, which override key, does the phase read the project) were scattered across four call sites — exactly the kind of thing a hand-written docs surface would drift from.

## What

- **Fix** — `resolved_prompts` now feeds `p.batch` from the synth slot, and both offline sites route through it. A per-cast synth `preamble` reaches `batch_submit`/`deliberate` answers, not just the interactive phases.
- **Refactor** — one `Phase` enum + `resolve_phase_preamble`: the single keyed layering point (built-in vs override key vs project-reading gate). Every live tool routes through it. Behaviour-preserving (295 → 298 tests green).
- **Add** — `kaibo://prompts`: renders each phase's system preamble through that *same* seam (any `[prompts]` override folded in), plus the user-turn framing via the real `consult_user_prompt`/`deliberation_prompt`. The path-/cast-dependent layers (`[orientation]`, `[context]`, per-slot `preamble`) are named, not rendered — a static resource can't resolve them.

## Design notes

- The point of routing the resource through `resolve_phase_preamble` (per Amy's steer) is that it calls the *exact* code a live call does — the verbatim-render test is the anti-drift guard, and if anyone ever restates a preamble in the doc instead of rendering it, that test breaks.
- The `Phase::reads_project()` gate is where the old bare `None, None` at the oneshot/batch call sites now lives — one place decides that oneshot and the offline synth own their context.

## Tests

- `resolve_phase_preamble_routes_each_phase_and_gates_project_layers` — routing + project-layer gate + per-key override isolation.
- `slot_preamble_..._feeds_all_synth_phases` / `phase_prompts_are_the_fallback_...` — extended to cover the `batch` key (the fix).
- `prompts_resource_is_advertised_and_renders_each_phase_verbatim` / `prompts_resource_reflects_a_prompts_override`.

Full suite green; fmt clean on edited files; no new clippy warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -191,10 +191,14 @@ the git log. Each later release appends a new section at the top.
   into the user turn. It's an audit surface (what is a model actually reading?) and the
   companion to tuning a preamble: override a phase's role framing globally with the
   `[prompts]` table or per cast with a slot's `preamble`, and the resource shows the
-  result. Relatedly, a **synth slot's `preamble` now frames the offline synth too** — a
-  per-cast voice set on a `batch`/`deliberate` cast reaches its `batch_submit` /
-  `deliberate` answers, not just the interactive `consult`/`oneshot` phases (previously
-  only the global `[prompts].batch` did).
+  result. **`kaibo://prompts/<cast>`** goes one step further — it resolves *that cast's*
+  framing, its per-slot `preamble`s folded in the way a live call layers them, and
+  attributes each phase to whichever set it (cast slot › global `[prompts]` › built-in) —
+  so you see precisely what one cast's models are told. Relatedly, a **synth slot's
+  `preamble` now frames the offline synth too** — a per-cast voice set on a
+  `batch`/`deliberate` cast reaches its `batch_submit` / `deliberate` answers, not just
+  the interactive `consult`/`oneshot` phases (previously only the global `[prompts].batch`
+  did).
 - **Zero-config workspace root.** When no `--root` is set, kaibo adopts its launch
   cwd as the inferred default root (it already scoped containment to that cwd, and
   MCP clients start stdio servers with cwd = workspace), so a call may omit `path`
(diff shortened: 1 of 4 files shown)


tobert:
## Follow-up: `kaibo://prompts/{cast}` + explicit phase-sharing note

Rebased onto current `main` (post-#42), then two more commits:

- **`3ac4d5b`** — `kaibo://prompts/<cast>`: resolves *that cast's* framing (per-slot `preamble`s folded in the way a live call layers them) and attributes each phase to whichever layer set it — cast slot › global `[prompts]` › built-in. The per-slot layering moved out of the handler into `Cast::resolved_prompts` so the live path and the resource share one function (no second place to drift); `Phase::slot_role` gives the resource the phase→slot mapping. Unknown cast → `resource_not_found` naming the roster.
- **`c677fd4`** — base `kaibo://prompts` now spells out that a phase is a role several tools share (explorer framing drives `explore` + consult's sweep + `deliberate`'s dossier; offline-synth framing serves `batch_submit` + `deliberate`'s synth), pinned by a test.

## Cross-family review (dogfooding kaibo)

Two passes, different lineages, both clean:

- **DeepSeek** (base seam + offline-synth fix): "correct and well-designed"; independently flagged the offline-synth change as a genuine bug fix.
- **Gemini** (per-cast delta): "flawless, no changes required" — confirmed the attribution tag is correct in every precedence combination (incl. a synth-slot `preamble` coexisting with a global `[prompts].batch`), no `strip_prefix` ordering hazard (exact URI matches first; `kaibo://prompts/` empty-cast → not-found), the `Cast::resolved_prompts` move is behavior-preserving, and `cast.expect("slot_set ⇒ cast")` is provably unreachable when `cast` is `None`.

Full suite green (0 failures), no clippy warnings in the new code.

PR #54, opened 2026-07-03
fix: whole-call wall-clock deadline so a wedged backend can't hang a call overnight

tobert:
## Why

A live `consult` (cast `lemonade`, a since-decommissioned local server) parked a Claude Code session **~17 hours overnight** until Amy interrupted it. Forensics on the still-alive process: kaibo idle, no in-flight work, all MCP socket queues empty — the tool call had *reached* kaibo (two `kaish` worker threads spawned the same second the transcript shows `consult` dispatched) and simply never returned.

Two problems:
1. **Trigger (fixed out-of-tree):** `--cast lemonade` came from a stale `~/.claude.json` pointing at `:13305`, a server retired 2026-06-26. Repointed to the default cast.
2. **The real defect (this PR):** the per-request `request_timeout` (down in reqwest, injected via rig) was the *only* brake. The 2026-06-06 fix (`tests/llm_timeout.rs`) proved it catches "no bytes ever", but not every wedge — a stalled response *body* across rig's split send/bytes read, or a pooled keep-alive to a wedged server — and with 900s configured it still ran 17h. kaibo trusted the transport for liveness with no backstop of its own.

## What

A kaibo-owned `tokio::time` ceiling on the whole interactive call: **`ConsultConfig::call_deadline`** (config `call_deadline_secs` / env `KAIBO_CALL_DEADLINE_SECS`, default **1h**), wrapping `consult_with` / `explore_with` / `oneshot` via a new `with_call_deadline`. Transport-agnostic, so it fires whatever the wedge shape.

- Default sits **above the largest per-request timeout** (even a 30-min local model) so it never cuts a legitimately slow completion — it turns "overnight" into "you'd notice within the hour".
- **Exemptions by construction:** `deliberate`'s offline Stage-2 synth and the batch lanes intend to wait hours; the deliberate Stage-1 dossier build the caller *does* wait on is bounded. `consult_submit` inherits the deadline (a wedged background consult should abort, not burn tokens forever).
- A deadline abort classifies as a **transient/retryable** condition (`classify_failure`), not a kaibo bug to report; the error names the deadline and the knob to raise.

## Tests (failing-first)

- `consult_aborts_when_a_backend_wedges` / `explore_aborts_when_the_explorer_wedges` — the offline consult/explore paths abort a wedged `hang_model` scripted backend (new harness capability). Teeth verified by making the deadline ineffective and watching the outer guard fire.
- `call_deadline_bounds_a_call_even_with_a_generous_request_timeout` — integration test proving `call_deadline` (2s) dominates a **generous 300s** wire timeout through a real black-hole socket.
- `wall_clock_deadline_is_transient_not_a_kaibo_bug` — the failure classification/guidance.

Full suite green (all targets); clippy clean; edited-file-only formatting.

## Review

Cross-family reviewed by **DeepSeek via kaibo** (holistic, no diff): confirmed the wrap layer (no nesting/escape hazard), that no synchronous entry point was missed, the deliberate Stage-1/Stage-2 split is exact, and the classification is sound. Verdict: "correct at every seam."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -285,6 +285,21 @@ the git log. Each later release appends a new section at the top.
 
 ### Fixed
 
+- **A stalled backend can no longer hang a call overnight.** An interactive
+  `consult` / `explore` / `oneshot` now runs under a whole-call wall-clock deadline
+  (`call_deadline_secs`, default 1 hour; env `KAIBO_CALL_DEADLINE_SECS`), independent
+  of the per-request `request_timeout`. The per-request timeout catches a backend that
+  never answers, but not every wedge shape — a stalled response body, or a pooled
+  keep-alive to a server that stopped responding, once parked a real consult ~17 hours.
+  Past the deadline the call aborts with a clean tool-result error naming it (classed as
+  a transient/retryable condition, not a kaibo bug), so your session keeps moving instead
+  of waiting forever. Keep the value above your slowest legitimate single completion. It
+  bounds the interactive loop tools — `consult` / `explore` / `oneshot` and async
+  `consult_submit`. `deliberate`'s direct lane (one long local completion) is bounded
+  instead by its synth backend's `request_timeout`, so a slow local model keeps its full
+  patience without forcing this ceiling high; the batch lane holds no in-process wait
+  (the work runs on the provider's queue).
+
 - **The advertised cast roster marks the default even when it's set by an alias.**
   Setting `server.cast` (or `--cast` / `KAIBO_CAST`) to a cast *alias* — say `claude`
   for `anthropic` — used to drop the `(default)` tag from the handshake's `## Casts`
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -110,6 +110,20 @@ thinking_style = "auto"
 # tell wedged from slow. 0 is rejected at load.
 request_timeout_secs = 900
 
+# Whole-call deadline (seconds): the wall-clock ceiling on ONE interactive call
+# (consult / explore / oneshot), across all its completions. This is the backstop
+# for when the per-request timeout above doesn't fire — a wedged local server on a
+# pooled keep-alive, a stalled response body — so a stalled backend aborts loudly
+# instead of hanging your session (a real consult once parked ~17h). Keep it
+# comfortably ABOVE the largest request_timeout a call can reach, so it never cuts a
+# legitimately slow single completion. Bounds the interactive loop tools — consult /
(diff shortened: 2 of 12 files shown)


tobert:
## Follow-up (4407a36): async ≠ unbounded

Amy pushed back on the "deliberate and batch are exempt" framing — correctly. It conflated two different things, now split:

- **deliberate's DIRECT lane** holds one long *in-process* local completion (spawned as a `job-N`, but kaibo holds the wait) — same category as a backgrounded `consult_submit`. It now wears `call_deadline`, so a wedged local synth can't leave a job running forever and `job_wait`/`job_get` resolve within the deadline. Threaded a `deadline` arg through `deliberate_direct`.
- **The BATCH lane** genuinely escapes, and only because kaibo holds no in-process wait there: submit is one quick handle-returning call (bounded by `request_timeout`), and the deliberation runs on the *provider's* queue, collected later by polling. `call_deadline` structurally can't bound a provider-side batch turnaround — there's nothing in-process to wrap.

**The rule is now uniform:** every in-process model call — sync (`consult`/`explore`/`oneshot`) and async (`consult_submit`, `deliberate` direct) — is bounded by `call_deadline`; only the provider batch queue is outside it. New failing-first test `deliberate_direct_aborts_when_the_local_synth_wedges`. Docs updated to say this precisely.

tobert:
## Follow-up (b55c12d): deliberate-direct rides `request_timeout`, not the interactive ceiling

Answering Amy's "would we advise a 2–4h deadline for a slow local deliberate?" — no, and the previous commit had a smell: with one global `call_deadline` bounding deliberate-direct too, a slow deliberate would force `call_deadline` high, re-opening the interactive-hang window this whole PR closed.

The bound now keys off the **shape** of the work, not sync-vs-async:

| Path | Bound by |
|---|---|
| `consult` / `explore` / `oneshot` / `consult_submit` (multi-completion loops) | `call_deadline` (tight, ~1h) |
| `deliberate` **direct** (exactly one completion) | its synth backend's **`request_timeout`** + 60s margin |
| `deliberate` **batch** / `batch_submit` (provider queue) | nothing in-process — kaibo holds no wait |

So a 3h-patience local synth just needs a 3h `request_timeout` (which you set anyway for it to answer at all) and the direct job inherits it — **you tune one knob per model, and never touch `call_deadline` for deliberate**. The interactive ceiling stays tight.

Extracted the sizing into a pure `deliberate_direct_deadline(backend)` with a test pinning that it tracks `request_timeout` and outlasts the interactive `call_deadline`. Full suite green (21 suites), clippy clean.

PR #55, opened 2026-07-03
feat: first-class OpenRouter provider — one key, every family, reasoning on by default

tobert:
## What

A new keyed `ProviderKind::OpenRouter` on rig 0.38.2's native openrouter client: a user with an [OpenRouter account](https://openrouter.ai/docs/quickstart) sets `OPENROUTER_API_KEY` (or `~/.openrouter-key`) and the built-in `openrouter` backend + cast light up — one key reaching every major model family. The `configure` prompt now walks a setting-up agent through OpenRouter, including using the **public model catalog** (`GET /api/v1/models`, filterable by tools support, category, price) to propose cast models with live data.

## Decisions (from the research arc with Amy, 2026-07-03)

- **Native rig provider over the generic openai wire.** rig's openrouter module round-trips structured `reasoning_details` — including Anthropic's signed thinking blocks — across tool-call turns, which the generic path cannot preserve. For kaibo's multi-turn consult loop with thinking on, that's correctness, not polish.
- **Reasoning maximized by default, opt-out per slot.** Doctrine: anything kaibo calls has thinking on unless the user says otherwise. The new `ThinkingStyle::OpenRouterEffort` emits OpenRouter's unified `{reasoning:{effort:<role>}}`, which the gateway translates per upstream provider and silently drops where unsupported — so unconditional emission is safe. Effort passes through verbatim: slot `effort = "xhigh"`/`"max"` reaches OpenRouter's deeper rungs, `effort = "none"` is the documented opt-out.
- **rig 0.38 defect worked around, with a tripwire.** rig's `OpenrouterCompletionRequest` silently drops `max_tokens` (confirmed still broken on rig `main`) — a thinking-on answer would starve on the gateway's default budget. The budget rides the flattened `additional_params` as `max_completion_tokens`; a canary test fails on any rig bump past 0.38 and names the re-audit, so the fix upstream can't silently double-send. Exit strategy if rig gaps accrete is logged in `docs/issues.md` (direct-SDK breakout; `openrouter-rs` surveyed as pilot candidate).
- **Tool-result images stay off this transport.** rig's converter rewrites a tool-result image to placeholder text — silent data loss — so the arm routes seen images through the existing user-turn rewrite path instead.
- **Drift-proof built-in cast.** Explorer `~google/gemini-flash-latest`, synth `~anthropic/claude-sonnet-latest` — OpenRouter's live `~author/family-latest` catalog aliases track the newest family member instead of rotting like hardcoded ids. Both are multimodal-in per the catalog, so vision is pinned on both slots.
- **Loud edges:** keyed kind, fixed endpoint (`base_url` in config is a load error), no batch lane (honest refusal naming the batch-capable kinds).

## Review

Cross-family per house discipline, no diff handed over: **Gemini Pro** (batch lane, twelve whole files attached) and **DeepSeek** (agentic consult over the worktree). Both judged the design sound; their three real findings are closed in the final commit (ModelCaps test coverage for the new kind, an arm-level params-assembly integration test, the rig-bump canary) plus the `effort = "none"` opt-out documentation. One DeepSeek finding (missing CHANGELOG entry) was a false positive.

## Verification

- Offline suite green: 319 lib tests + all integration suites, 0 failures; instructions budget tests pass (the resident-text additions were paid for by compression).
- `cargo tree -i aws-lc-rs` / `aws-lc-sys` / `openssl-sys` all empty — TLS invariant holds.
- The `#[ignore]`d live probe (`tests/consult.rs`, gated on `OPENROUTER_API_KEY`) has **not** been run — no key on this machine. Worth one live round-trip before this ships in a release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -280,9 +280,23 @@ the git log. Each later release appends a new section at the top.
 - **OpenRouter as a first-class provider.** One `OPENROUTER_API_KEY` now reaches
   every major model family through the built-in `openrouter` backend and cast, with
   reasoning on by default (OpenRouter's unified `effort` param, forwarded verbatim so
-  a synth slot can reach past the usual `high` default into `xhigh`/`max`). The
-  built-in cast pins OpenRouter's `~author/family-latest` catalog aliases, so it stays
-  current as new models ship instead of rotting on a dated id.
+  a synth slot can reach past the usual `high` default into `xhigh`/`max` — measured
+  live: effort rides through the gateway and bills real reasoning tokens). Setting a
+  slot's `effort = "none"` sends OpenRouter's structural disable
+  (`{"reasoning": {"enabled": false}}`), so the opt-out doesn't depend on how the
+  gateway's effort ladder happens to read the string. The built-in cast pins
+  OpenRouter's `~author/family-latest` catalog aliases, so it stays current as new
+  models ship instead of rotting on a dated id. Every OpenRouter call carries an
+  explicit prompt-cache breakpoint, so Anthropic-family models behind the gateway
+  bill their (large, resident) system preamble at cache-read rates instead of full
+  input price every turn — providers whose caching is implicit simply ignore it.
+  **Data collection is denied by default**: one OpenRouter slug routes across
+  competing upstream hosts whose data policies differ, and kaibo's prompts carry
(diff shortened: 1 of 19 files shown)


tobert:
Gonna wait for #56  on this. GLM explorer was really expensive in my first try, needs to do more whole-file reads for the kaibo style.

PR #56, opened 2026-07-03
attach means the model sees the bytes — inline for consult, read-WHOLE directives for sweeps

tobert:
## What

Unifies `attach` semantics across every tool: **attaching a file puts its bytes in front of the answering model.** Previously `consult` only *named* attached files in the driver prompt, the delegated `explore′` sweeps never heard about them at all, and the 64 KB kaish output cap made the "read it in full with `cat -n`" instruction physically impossible in one pass for a large file — the caller's strongest signal was the one input kaibo delivered weakest.

## How each tool delivers it

- **`consult` / `consult_submit`** — text attachments inline whole into the driver prompt, numbered `cat -n` style inside the shared `<file>` wrapper. Numbering now applies at every inline site (oneshot/batch too): raw bytes invite guessed line numbers, and exact `file:line` citations are the product.
- **Inline budget** — `[defaults] inline_attach_budget` / `KAIBO_INLINE_ATTACH_BUDGET` (default 256 KiB; `0` legal = inline nothing, the escape hatch for small-context local casts). Inlined bytes ride every driver-loop turn, so the budget bounds resident prompt cost. A file past the remaining budget demotes **loudly** — named with its size under a command-voice "Read each one WHOLE" directive with the sed-span paging idiom, never silently dropped.
- **Explorer sweeps** — every `explore′` preamble carries the same command-voice directive listing the text attachments (a sweep is a fresh agent that saw neither the driver prompt nor the inlined bytes). The top-level `explore` tool grows an `attach` arg with directive semantics; images are refused there (the sweep has no `view_image`).
- **Security posture preserved** — inlined consult bytes are read through the read-only kaish VFS (the same TOCTOU-safe mount read `resolve_attachments` uses); the 16-byte `std::fs` sniff survives only as a routing hint for files never inlined. Non-UTF-8 non-image attach is now refused loudly on consult too.

## Decisions (from the conversation with Amy)

- One semantic everywhere beats per-tool cleverness — the calling agent shouldn't memorize which tool inlines.
- Command voice for the directives ("Read each one WHOLE"), not soft steering — whole reads are the point; they cut cost and make everything faster.
- Byte-threshold demotion over a per-cast switch; a per-call budget override is deferred (`docs/issues.md`).
- Directives-not-bytes for sweeps: the explorer keeps agency over *when* the read happens and the bytes aren't multiplied per sweep.

## Tests

Numbered-wrapper units, budget partition (caller order, greedy, budget-0), scripted-loop tests pinning the directive into the `explore′` sweep preamble and `explore_with`, config-ladder tests for the knob. 21 test binaries green; clippy adds no new warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,8 +76,10 @@ the git log. Each later release appends a new section at the top.
   explorer with an offline synth (the example config's `fable`, `gemini-deliberate`, or
   `local-direct`) — `deliberate`'s `cast` enum lists the usable ones and `kaibo://config`
   shows each cast's lane. Reads the repo itself, so it takes `path` / `cast` /
-  `explorer_model` / `synth_model` (+ `_backend`s); the offline synth is a single turn, so
-  no `attach` / `context` / `session_id`. Gated independently by `--no-deliberate`. This is
+  `explorer_model` / `synth_model` (+ `_backend`s), plus `attach` — text files the
+  dossier-building explorer is directed to read WHOLE, so their content reaches the
+  offline synth through the dossier; the synth itself is a single turn, so no `context` /
+  `session_id`. Gated independently by `--no-deliberate`. This is
   the tool that finally routes the `direct` lane the per-slot lane reshape introduced. For
   an answer this turn, use `consult`.
 - **`oneshot`** — a thin, direct second opinion from a model outside your family:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,14 +22,19 @@ the git log. Each later release appends a new section at the top.
   answer carries a provenance footer naming the cast and the models that produced it.
   Args: `question`, `context`, `path`, `cast`, `session_id`, `attach`, `include_report`,
   and per-call `explorer_model` / `synth_model` (+ `_backend`) overrides. **`attach`**
-  names workspace files (under the project root) to put in front of the investigation —
-  unlike the tool-less tools' attach, kaibo does *not* inline them; it names them in the
-  prompt and the consult model opens each itself when it's ready, in full, building its own
-  narrative: a text file with the shell (`cat -n`), an **image** with its `view_image` tool.
-  An attached image therefore needs a vision-capable cast — kaibo refuses one to a
-  vision-blind synth up front (the same honest refusal `oneshot`/`batch` give) rather than
-  name a file the model could never open. The files just have to live under the root the
-  consult reads (a worktree counts).
+  puts workspace files (under the project root) in front of the investigation — attach
+  means *the model sees the bytes*. Text files are **inlined whole** into the
+  investigation prompt, lines numbered like `cat -n` so the model cites them by exact
+  `file:line`; a file past the cumulative inline budget (`[defaults]
+  inline_attach_budget` / `KAIBO_INLINE_ATTACH_BUDGET`, default 256 KiB; `0` = inline
+  nothing, the escape hatch for small-context local casts) is instead ordered read WHOLE
+  through the model's shell — demoted loudly with its size, never silently dropped. Every
+  delegated explorer sweep also gets a read-them-WHOLE directive for the attached files,
+  so a sub-agent is never blind to what you flagged as central. An **image** opens via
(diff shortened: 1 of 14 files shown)


tobert:
## Cross-family review (pre-merge dogfood)

Ran the usual combo over the whole changed files, no diff: **Gemini Pro** (batch lane) + **DeepSeek** (`consult`, agentic over the worktree). Both reports verified against the code before acting; fixes landed in 16eb5b3.

**Confirmed and fixed**
- *Unescaped paths in directive lists* (both reviewers, DeepSeek 1B / Gemini "minor injection"): a filename can legally contain `\n`, which would forge extra entries in the read-WHOLE / image / sweep lists. All prompt render sites now share `escape_attr_value`, with a forgery test.
- *`deliberate` lacked `attach`* (Gemini, consistency): contradicted "one semantic everywhere." Now takes `attach` as dossier-stage read-WHOLE directives; content reaches the offline synth through the dossier.
- *`explore` description omitted `attach`* (DeepSeek 3A): added within the 2048-char budget.
- *Test gaps* (both): end-to-end tempdir pipeline test (raw paths → resolver → final prompt text), all-three-kinds ordering test, impl-independent bare-tag assertions in the escape battery, newline-less-final-line escape edge.

**Rebutted**
- Gemini's *"critical wrapper breakout via `<\/file>`"* misreads the escape scheme: an attacker's literal `<\/file>` is already the escaped form and cannot read as a bare terminator — the "exactly one bare `</file>`" invariant holds either way. Now pinned by a test that doesn't reuse the impl regex.
- Gemini's VFS TOCTOU concern: the consult resolver uses the identical `KaishWorker::read_file` mount path `resolve_attachments` uses, covered by the existing `mount_layer_symlink_*` battery.

**Deferred (logged in docs/issues.md)**
- `KaishWorker::read_file` is unbounded → stat-then-read growth race can spike memory (real, preexisting, kaish-side; robustness not escape — the model can't drive filesystem timing).
- Per-call `inline_attach_budget` override; observe real inlined-attachment cost on hosted casts before adding surface.

DeepSeek confirmed no containment/TOCTOU regressions, correct `cat -n` numbering semantics (incl. CRLF matching shell behavior), and that all other tool descriptions already tell the new truth. 21 test binaries green.

tobert:
## Second arc on this branch: explorer read behavior (live-watch driven)

After the review fixes, live-watching consults on fresh builds drove four more commits — all in service of the same goal as attach-inlining (fewer, wider, cheaper reads):

- **Whole-first reading everywhere** (`aa34e5b`): every prompt surface now leads with "read files WHOLE — `cat -n FILE` is the default move"; grep is framed as the locator, the `wc -l` pre-probe is gone. The old "a *short* file: read it whole" made models classify before daring a whole read, then nibble — watched live.
- **Staged truncation, cap stays 64 KiB** (same commit): Amy's context arithmetic beat the naive cap-raise — measured on this tree, kaibo Rust runs **2.79 bytes/token**, so a maxed 256 KiB read ≈ 94K tokens riding every turn of a 128–250K-context explorer's sweep. Instead: exit 3's head+tail sample is stage one, then `grep -n SYMBOL FILE` + ~1,200-line targeted spans for what the question needs. Caller-flagged attachments keep their read-it-ALL directive.
- **grep `-E` alternation taught** (`70c6d52`): a live sweep burned ~10–15 turns on `grep 'a\|b'` patterns kaish silently matches nothing on. Cheatsheet now teaches `grep -rnE 'foo|bar'`; the real fix ships in kaish 0.11.0 (issues.md tracks jumping on that pin).
- **Preamble compression** (`050b2c5`, `ce4d49a`, `726bc11`): the read playbook now lives in exactly one place per role (the addendum slimmed to sandbox facts), and the upstream kaish contract's "Overlay mode" paragraph — write-layer teaching inside a read-only preamble — is filtered at the composition boundary, with a teeth test that fails loudly if upstream renames the heading. Caller-facing `kaibo://tools` aligned to the same doctrine (its explore section also stopped denying the new `attach` arg).

Verified live across three binary swaps: consult inlines numbered whole files, explore quotes its read-WHOLE orders back and complies, sweeps read small files whole and stage giants. 21 test binaries green throughout.

PR #57, opened 2026-07-04
fix(attach): bound every raw read so a stat-then-read swap can't OOM

tobert:
## Why

A Gemini Pro batch review of the attach-inline work (2026-07-03) flagged an **OOM vector** on the file-read path: callers `stat` a file, check it against a budget, then read it — and the read (`KaishWorker::read_file`) was **unbounded**. A file swapped to something enormous in the stat→read window would slurp whole into memory before the post-read size check could refuse it. Verified real (the VFS `Job::Read` handler called `project_vfs.read`, no ceiling).

The same review's *critical* finding — a `<file>` wrapper breakout — was a **false positive** (it read the backslash escape *output* `<\/file>` as the wrapper's delimiter, but the wrapper's own tags are the bare `<file>`/`</file>`). Both this branch's cross-family reviewers independently confirmed the escape is sound.

## What

- **Bound the read at the read itself.** `read_file` → `read_file_capped(path, max_bytes)`, carried through `Job::Read` as a `limit` honoured by the VFS `read_range` (`ReadRange::bytes(0, limit)` → `LocalFs`'s `File::take` — a real ceiling, no whole-file slurp). The unbounded read method is **removed** — an uncapped read of caller-swappable bytes is a footgun with no user.
- **Every caller passes `budget + 1`**, so the *existing* post-read size check becomes the overflow detector — no new branching. consult demotes to `TextOversize`, view_image refuses, oneshot/batch let `classify` refuse by length. Honest-file behaviour is byte-for-byte unchanged; only a raced swap now stops at the ceiling.
- **Commentary at every touched site** spells out the TOCTOU and how the cap closes it, and a note in `escape_file_body` pre-empts re-raising the false-positive breakout.
- **Test teeth**, proven by removing the guard and watching the test fail:
  - `read_file_capped_bounds_the_bytes_pulled` — an over-cap read returns exactly `cap` bytes (flip the handler to the unbounded read → 4096 vs 400).
  - `file_tag_lookalikes_in_body_are_all_escaped` rewritten to assert each injected variant survives in its *escaped* form (drop `(?i)` → the whitespace/case-variant assertion fails). This closes a tautology the reviews caught — see comments.

## Review

Cross-family, no diff shown to either — a Gemini Pro batch (whole files) and a live DeepSeek agent on the worktree. Both returned a clean bill on correctness and security. The detailed findings, the one defect fixed, and the harmless-by-design items we're *not* acting on are in the PR comments.

## Changelog

Security bullet added — the size-swap sibling of the existing symlink-swap defense.

### DIFF
--- a/src/attach.rs
+++ b/src/attach.rs
@@ -482,14 +482,31 @@ mod tests {
             .strip_prefix("<file path=\"evil.md\">\n")
             .and_then(|s| s.strip_suffix("\n</file>"))
             .expect("wrapper brackets the body");
-        let tag = Regex::new(r"(?i)<\s*/?\s*file\b[^>]*>").unwrap();
+        // Teeth for the escape itself, one assertion per dimension: each injected
+        // lookalike must appear in the body in its *escaped* form (`<` → `<\`). We assert
+        // the escaped literal is *present* rather than that the impl's own regex no
+        // longer matches `inner` — the latter is a tautology, and two cross-family
+        // reviews (2026-07-03 and 2026-07-04) flagged the earlier `!tag.is_match(inner)`
+        // for exactly that: `replace_all` removes every match by construction, so a regex
+        // that stopped catching `< / FILE >` would leave it unescaped *and* fail to match
+        // it, and the assertion would pass blind. Checking the escaped literal fails loud
+        // if any dimension regresses — the close form, the *opening* form, and the
+        // whitespace+case variant that proves the `(?i)` and `\s*` handling.
         assert!(
-            !tag.is_match(inner),
-            "no bare <file>-tag lookalike survives in the body: {inner}"
+            inner.contains(r"<\/file>"),
+            "the bare close lookalike is escaped: {inner}"
         );
-        // Independent of the escape regex (so this isn't circular with the impl): on
-        // the FULL wrapped output, the only literal open/close tags are the wrapper's
-        // own — exactly one each (Gemini cross-family review, 2026-07-03).
+        assert!(
+            inner.contains(r#"<\file path="x">"#),
+            "the opening lookalike is escaped: {inner}"
+        );
+        assert!(
+            inner.contains(r"<\ / FILE >"),
+            "the whitespace+case variant is escaped (proves (?i) and \\s* tolerance): {inner}"
+        );
+        // And on the FULL wrapped output, the only literal open/close tags left are the
+        // wrapper's own — exactly one each. A bare delimiter that slipped the escape
+        // would push a count to 2.
         assert_eq!(
             wrapped.matches("<file path=").count(),
             1,
--- a/src/attach.rs
+++ b/src/attach.rs
@@ -116,6 +116,16 @@ pub fn check_attachment_bounds(
 /// `if x < y` reads truer than `if x &lt; y`. The backslash form leaves the body legible
 /// while removing the tag (`<\/file>` is no longer matched as a tag), so the only bare
 /// `<file>`/`</file>` left in the wrapper is kaibo's own.
+///
+/// A note to the next reviewer, because a cross-family pass mis-flagged this as a
+/// *critical* breakout: the wrapper's own delimiters (in [`Attachment::wrapped_text`])
+/// are the **bare** `<file …>` / `</file>` — there is no backslash in them. The
+/// backslash form `<\/file>` is *only ever this function's neutralized output*. So a
+/// body already containing a literal `<\/file>` is not a breakout — it is byte-identical
+/// to what we'd emit for a defanged tag, i.e. already safe, and the regex rightly leaves
+/// it alone. Breaking out would require a *bare* `</file>` to survive into the body; that
+/// is exactly what the regex rewrites, and what the wrapper-count assertions in
+/// `file_tag_lookalikes_in_body_are_all_escaped` verify with a literal (non-regex) scan.
 fn escape_file_body(body: &str) -> String {
     static FILE_TAG: LazyLock<Regex> = LazyLock::new(|| {
         Regex::new(r"(?i)<\s*/?\s*file\b[^>]*>").expect("static file-tag regex compiles")
--- a/src/sandbox.rs
+++ b/src/sandbox.rs
@@ -43,7 +43,7 @@ use kaish_kernel::interpreter::ExecResult;
 use kaish_kernel::tools::{Tool, ToolArgs, ToolCtx, ToolRegistry, ToolSchema};
 use kaish_kernel::vfs::{ByteBudget, Filesystem, LocalFs, MemoryFs, VfsRouter};
 use kaish_kernel::{
-    IgnoreConfig, Kernel, KernelBackend, KernelConfig, LocalBackend, OutputLimitConfig,
+    IgnoreConfig, Kernel, KernelBackend, KernelConfig, LocalBackend, OutputLimitConfig, ReadRange,
 };
 
 /// Wraps a real builtin, preserving its identity and schema but refusing to run.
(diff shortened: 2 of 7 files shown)


src/sandbox.rs
```diff
@@ -455,16 +464,25 @@ impl KaishWorker {
...
+                                // read of `[0, limit)`. `LocalFs::read_range` honours it
+                                // with `File::take(limit)`, so a raced swap to a huge
+                                // file stops at `limit` instead of slurping to OOM — the
+                                // caller sizes `limit` one byte past its budget and
+                                // detects the overflow by the returned length. (The
+                                // kernel's own `vfs()` carries only its internal `/v/*`
+                                // scratch under `with_backend`, not these mounts — hence
+                                // the retained handle.)
+                                let out = project_vfs
+                                    .read_range(&path, Some(ReadRange::bytes(0, limit)))
```
tobert, line 483: **The fix, in one line.** The read now rides `read_range` with a byte `limit`, which `LocalFs` honours via `File::take(limit)` — a real ceiling, never a whole-file slurp-then-slice. This closes the stat→read TOCTOU: a file swapped to something enormous after the caller's `metadata()` stops here at `limit`, and the caller detects overflow by length. The uncapped `read_file` is **gone**, so no future caller can reintroduce an unbounded read.

Both reviewers traced this independently and rated it clean — DeepSeek: *"every disk read bounded"*; Gemini Pro: *"the OOM windows are successfully closed."*

src/view_image.rs
```diff
@@ -130,12 +130,20 @@ impl ViewImage {
...
-                canon.display()
-            ))
-        })?;
+        // Read at most one byte past the limit: a file swapped to something enormous
+        // between the caller's view and this read stops at the cap instead of slurping
+        // to OOM, and a returned length past `max_bytes` is exactly the over-limit
+        // signal the size-check below refuses on.
+        let bytes = self
+            .worker
+            .read_file_capped(&canon, self.max_bytes as u64 + 1)
```
tobert, line 139: Call site 1/3. Cap = `max_bytes + 1`; the existing `bytes.len() > max_bytes` check below becomes the overflow detector — a raced-huge image is refused, not slurped. The `budget+1` convention was verified correct here by both reviews (exact-at-budget passes, `+1` caught).

src/server/containment.rs
```diff
@@ -202,12 +202,22 @@ impl super::KaiboHandler {
...
-                    None,
-                )
-            })?;
+            // Cap the read one byte past the largest a single file may legally be — the
+            // greater of the two per-encoding caps `classify` enforces. The stat above
+            // fed the *batch* budget; this bounds the *per-file* read so a stat-then-read
+            // swap can't slurp a raced-huge file into memory. An over-cap file comes back
+            // truncated at cap+1 and `classify` refuses it loudly by length, exactly as it
+            // would refuse an honest over-cap file — same outcome, no OOM window.
+            let read_cap = DEFAULT_MAX_TEXT_BYTES.max(DEFAULT_MAX_IMAGE_BYTES) as u64 + 1;
```
tobert, line 211: Call site 2/3 (oneshot/batch). Ceiling is the larger of the two per-encoding caps `classify` enforces; an over-cap file reads truncated at `cap+1` and `classify` refuses it by length — same outcome as before, now bounded.

**Gemini** noted a residual: the batch-level `total_bytes` budget (summed from the stats above) can be bypassed by a post-stat swap — but since each read is individually capped, peak memory is bounded to 64 × 5 MiB ≈ 320 MiB, a safe degradation, not an OOM.

**DeepSeek** noted this resolver has **no direct integration test**. Low-risk — the read-cap is teeth-tested at the worker level and the unbounded read is gone (a regression won't compile) — tracked as an optional follow-up.

src/server/mod.rs
```diff
@@ -1037,19 +1037,27 @@ impl KaiboHandler {
...
+            // Capping the read at `remaining + 1` bounds that window — a raced-huge file
+            // comes back as `remaining + 1` bytes, which the length check below reads as
+            // "over budget" and demotes to `TextOversize` (the model then `cat`s it
+            // itself), never an OOM. So the demotion path does double duty: it catches
+            // both an honest over-budget file and a raced swap, from the one length test.
             let bytes = worker
                 .as_ref()
                 .expect("worker was just spawned")
-                .read_file(canon.clone())
+                .read_file_capped(canon.clone(), remaining + 1)
```
tobert, line 1051: Call site 3/3 (consult). Cap = `remaining + 1`; the existing `bytes.len() > remaining` check demotes to `TextOversize` — the demotion path now does double duty (honest over-budget **and** raced swap) from one length test.

Two notes from the reviews, both harmless: (a) the only other `std::fs` read in the tree is the 16-byte prefix sniff at ~L1006 — a 16-byte stack buffer, bytes dropped, and the authoritative read still goes through the capped VFS; (b) `remaining + 1` would wrap at `u64::MAX`, but `remaining` comes from config (256 KiB default), never an attacker.

src/attach.rs
```diff
@@ -472,14 +482,31 @@ mod tests {
...
+            "the bare close lookalike is escaped: {inner}"
+        );
+        assert!(
+            inner.contains(r#"<\file path="x">"#),
+            "the opening lookalike is escaped: {inner}"
+        );
         assert!(
-            !tag.is_match(inner),
-            "no bare <file>-tag lookalike survives in the body: {inner}"
+            inner.contains(r"<\ / FILE >"),
```
tobert, line 504: **Remediation of the one real defect the reviews found.** Two cross-family passes flagged the previous `!tag.is_match(inner)` assertion as a tautology: it re-ran the impl's *own* escape regex, which `replace_all` guarantees no longer matches — so a regex that regressed on the `< / FILE >` whitespace/case variant would leave it unescaped **and** fail to match it, passing blind. The neighbouring literal-count assertions only had teeth for the *exact* bare delimiters.

Now each injected variant is asserted present in its **escaped** form, one assertion per dimension (close, open, whitespace+case). Proved teeth: dropping `(?i)` from the escape regex makes *this* assertion fail (`src/attach.rs:504`), where the old one stayed green.

(DeepSeek independently rated the escape secure but missed the tautology — the cross-family split earned its keep here.)

tobert (commented):
## Cross-family review — grounded, no diff shown

Two lineages reviewed the branch holistically (whole files, not a diff): a **Gemini Pro** batch and a live **DeepSeek** agent reading the worktree. Neither was told the earlier "critical breakout" was a false positive — both independently ruled the `<file>` escape **secure**, which refutes it.

| Dimension | DeepSeek | Gemini Pro |
|---|---|---|
| TOCTOU / unbounded reads | Clean | Clean ("OOM windows closed") |
| `budget+1` off-by-one | Clean | Clean ("flawlessly applied") |
| `<file>` escape (clean-room) | Clean, no breakout | Secure |
| Test teeth | missed the tautology | **caught the tautology** |
| Extra | containment.rs lacks a direct test | binary-over-budget burns one model turn |

**Fixed:** the tautological escape test — see the inline comment on `src/attach.rs`.

**Not acting (both rated harmless):** batch `total_bytes` post-stat bypass (each read is individually capped, so peak memory is bounded to ~320 MiB); FIFO/pipe stall (not model-drivable — read-only mount); binary-over-budget demotes without the `from_utf8` check and burns one model turn (a deliberate memory optimization — reading megabytes just to prove it's binary defeats the point).

**Open thread:** a direct integration test for `resolve_attachments` (oneshot/batch) — low-risk (read-cap is teeth-tested at the worker level and the unbounded read is gone), offered as a follow-up.

Inline comments map each finding to its `file:line` and how it was handled.

PR #58, opened 2026-07-04
fix(job_wait): park-and-coalesce so `level:"info"` stops poll-storming

tobert:
## The problem

Running the kaish 0.11.0 pre-release review as one background consult, Amy parked with `job_wait(timeout_secs=300, level="info")` and watched **~130 back-to-back calls return in seconds each** — never near the 300s window. A live A/B on the *same* running job confirmed the split: flipped back to default `level:"warn" timeout_secs=600`, it parked the full 600s ("Nothing new in 600s") and returned only on the clean timeout.

**Root cause (contributing factors):** `level` was doing two jobs at once. `wait_level_floor` became `wait_drain_with`'s `return_floor`, and the drain loop returned the moment it collected *any* record at that floor. At `level:"info"` that's the first `running kaish: …` line. Default `warn` only parked nicely because the sole Warn records are the completion/failure pings — so observability ("watch the narrative") and "park until something real happens" were **structurally mutually exclusive**. The tool *description* already promised the right behavior ("returns as soon as a job finishes or fails, or on a clean timeout") — the code had drifted from it.

## The fix

Two forks settled with Amy before writing code:

1. **Reinterpret `level`** as the observability *sample floor* only — never the return trigger.
2. **Wake on any Warn+** — a job finished/failed, or a real mid-flight warning. Narrative below Warn rides along in the tail but never cuts the park short.

`wait_drain_with` grew a third floor, so the three concerns stop fighting:
- `drain_floor` — what to consume (and stream to the human's progress channel).
- `sample_floor` — what rides back in the returned tail.
- `wake_floor` — what ends the block early (Warn+, always).

The returned sample also went from first-`limit` (**head**, then silence) to a sliding last-`limit` **tail** (newest activity), capped at Warn so the terminal ping is always present even if a caller asks for a higher floor. **The default (`warn`) behavior is unchanged** — only sub-warn levels stop poll-storming.

Why this matters beyond the poll storm: the live progress stream reaches the **human only** (nothing on the MCP channel wakes the calling agent), so the returned tail is the model's *sole* window into the narrative. Coalesce-and-return-the-tail is the actual mechanism by which a watching model sees anything, not a nicety.

## Verification

- **TDD, teeth proven** by reverting the loop to the old behavior and watching two of three new `mcp_log` tests fail:
  - `sub_wake_narrative_parks_to_timeout_then_returns_the_tail` — wake decoupled from sample (the poll-storm fix).
  - `over_limit_returns_the_newest_tail_including_the_wake` — tail, not head.
  - `a_warn_wakes_the_park_and_the_tail_carries_the_narrative` — behavior lock (passes both ways).
- 341 lib + all integration suites green; clippy clean on touched files.
- Model-facing text re-read holistically (per *Writing for models*): the `level` schema doc (was literally "Lowest level to return"), the `job_wait` tool description, and the `kaibo://tools` resource all now say *`level` sizes the sample, never the timing — pass a shorter `timeout_secs` to check in more often*.
- CHANGELOG `job_wait` bullet refined (unreleased 0.2.0 — this corrects an unreleased feature, not a shipped one) + a devlog entry.

## Open design note for review

Wake = **any Warn+**, so a genuine mid-flight `warn!` (rate-limit/retry) also cuts the park short — matching how the default already treats the Warn bar. The original observation phrased it as "terminal event (job finished/failed)"; in practice those *are* the Warn records. If we'd rather a mid-flight warn *not* wake the park (strictly job-tagged pings only), it's a one-line change to the wake predicate — flagging for the reviewers' judgment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,16 +128,20 @@ the git log. Each later release appends a new section at the top.
   file, or a binary that isn't a known image is refused with a clear error.
 - **`job_wait`** — block briefly and productively for your async work instead of
   blind-polling `job_get`. Fire off consults and batches, do your other work, then `job_wait`
-  when you're ready to spend a minute on kaibo: it blocks up to `timeout_secs` (you
-  choose — no clamp; interruptible) and returns as soon as something lands, or on a clean
-  timeout. By default it hands back what kaibo flags as worth your attention (a job
-  finished/failed, a research-limit) plus which consult jobs are still running; pass
-  `level: "info"` to also pull the watchable narrative — each kaish command, sweep, and
-  milestone the agents ran — into your context. Name batch handles in `handles` to fold a
-  one-shot poll of them in too. Nothing wakes you (you choose when to block) and it isn't
-  the source of truth — `job_get`/`job_list` are; a clean empty return just means nothing new yet.
-  This pairs with launching work in parallel: submit several, do everything else, then
-  `job_wait` to merge the outputs.
+  when you're ready to spend a minute on kaibo: it parks up to `timeout_secs` (you
+  choose — no clamp; interruptible) and returns early only when a job finishes or fails (a
+  real event), else on a clean timeout — narrative alone never cuts the park short, so a
+  single `job_wait` watches a long job without turning into a poll storm. On return it hands
+  back a sample of what happened plus which consult jobs are still running. `level` sizes
+  that sample, not the timing: the default (`warn`) is the flagged milestones; `level:
+  "info"` folds in the watchable narrative too — each kaish command, sweep, and milestone
+  the agents ran, coalesced to the most recent `limit` — so a richer level fills the tail
+  without ever making the call return sooner (to check in more often, pass a shorter
+  `timeout_secs`). Name batch handles in `handles` to fold a one-shot poll of them in too.
+  Nothing wakes you (you choose when to block) and it isn't the source of truth —
+  `job_get`/`job_list` are; a clean empty return just means nothing new yet. This pairs with
+  launching work in parallel: submit several, do everything else, then `job_wait` to merge
+  the outputs.
 - **Async consults are watchable again.** A `consult_submit` job now streams its liveness
   (each kaish command, sweep, and milestone) onto kaibo's logging channel — the live
   "watch it work" view a synchronous `consult` always had, restored for the async path.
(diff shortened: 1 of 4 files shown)


tobert:
## Cross-family review 1 — DeepSeek (`consult`, agentic, holistic; read the worktree, no diff handed over)

**Verdict: merge-ready.** Traced the loop, the handler floor math, the default-path equivalence, and the tests; found no correctness bugs. Highlights:

- **Loop ordering is correct.** The drain → `notified()` → deadline → `select!` sequence closes the race: a push between the drain and the `notified()` registration calls `notify_one()`, which stores a permit that the just-registered `notified` future consumes → the `select!` fires immediately and the next iteration drains it. No record missed.
- **Wake decoupling is correct.** `wake_floor` is hardcoded to Warn in the handler, so narrative below Warn never ends the park — the actual fix for the poll storm.
- **Terminal ping always in the tail.** `sample_floor = wait_level_floor(level).min(wake_floor)` means even `level:"error"` caps to Warn, so a completion ping (Warn) always rides back.
- **`drain_floor` drops to Info only with a human progress token**, so narrative isn't consumed out of the ring when nobody's watching; the live Info tee still streams (gated `>= info_floor`, fire-and-forget).
- **Default `level:"warn"` genuinely unchanged** — without a token all three floors are Warn (identical to the old single-floor call); with a token it's strictly additive (human gains a live stream; the model's return is unchanged).
- **Three-floor decomposition judged the right, minimal shape** — the three concerns (consume / return / wake) are genuinely independent; collapsing drain+sample would either drop the live view or starve a later `job_wait`.
- **Tests have teeth:** `sub_wake_…parks_to_timeout` is the wall-clock regression test (≈0ms under the old bug vs ≥100ms now); `over_limit_…newest_tail` pins the sliding eviction; the streaming test pins the consume/return split.

### Nits to fold (follow-up commit)
1. **Stale comment** at `src/server/mod.rs:2529` still says "the model's return is the separate `return_floor` collection" — `return_floor` is the old name, now `sample_floor`. *(Will fix.)*
2. **`limit == 0`** is degenerate: the tail is pushed-then-popped empty, so a *wake* would return an empty result that `render_wait` prints as "Nothing new in Ns." Not reachable from `job_wait` (default 20), but a `limit.max(1)` clamp would be defensive. *(Considering.)*
3. Minor: the "before re-checking" phrasing in the loop comment, and a couple of test gaps (`limit==0`, an explicitly-raced push) — low-risk, `Notify`'s permit is Tokio-tested.

Second-family review (Gemini Pro batch, `mcp_log.rs` attached whole) is running; findings + any fixes will follow here.

PR #59, opened 2026-07-04
chore(deps): upgrade kaish-kernel 0.10.0 → 0.11.0

tobert:
## What

Bumps the embedded `kaish-kernel` dependency `0.10.0 → 0.11.0` (just released), and — capitalizing on it — trims shell coaching kaibo no longer needs to pay tokens for. **Zero kaibo *code* changes for the bump itself** — verified by a clean build and the full offline suite (525 tests green).

## Why the bump is safe despite four BREAKING entries

kaish 0.11's changelog carries four **BREAKING** items; we checked each against kaibo's real code (a cross-family `deepseek` review agreed, citing file:line — see the review comment below):

| kaish 0.11 BREAKING | Reaches kaibo? | Why |
|---|---|---|
| `ToolSchema.raw_argv` field added | No | kaibo builds `ToolSchema` only via `::new()`, never a struct literal; reads only `.name` |
| `ExecResult.latch` typed field; latched `--json` nonce moves out of `.data` | No | kaibo builds `ExecResult` via `::failure()` and reads only `code`/`err`/`out_bytes()`/`text_out()`. And it's **read-only** — no destructive op ever latches a nonce to relocate |
| `kaish-help` `Fragment` gains required `rank: u8` | No | kaibo never names `Fragment`; it composes help via `Recipe` / `SchemaContent` / `get_help` |
| scatter/gather `--format`/`--first` removed; `sed` `\|`/`\(…\)`/`\{n,m\}` → ERE | No | kaibo hand-rolls no docs pinning those flags or relying on `sed \|` erroring (its `sed` examples are line-range only) |

## What agents *gain* — for free

kaibo single-sources its kaish guidance from `kaish_kernel::help`, so the whole new agent-facing surface flows in on the bump with **no new resident cost**:

- native collections (`xs=[a b c]`, `{port: 8080}`), typed subscripts/slices, `keys`/`values`/`typeof` + `[[ -list ]]`/`[[ -record ]]` shape guards, typed membership
- a real `test` builtin evaluating POSIX conditions **through the VFS** — where kaibo's no-subprocess sandbox left the old `/usr/bin/test` shell-out dead, `test -f path` now works
- `fromjson`/`tojson`/`fromjsonl`/`tojsonl` bridges, real `jq -s` slurp, redirects inside `$(...)`
- importance-ranked onboarding (leads with the load-bearing rules, steers enumeration at `$(keys …)`/`$(values …)`); new `help regex` / `help collections` references

Functionally confirmed the new surface renders through kaibo's own path (`render_topic("collections")`, `help regex`, and `$(keys …)`/`$(values …)` now resident in the composed foundations).

## Also here: trimming obsolete shell coaching

0.11 makes BRE `\|` (and `\(…\)`, `\{n,m\}`) native operators **and** kaish now teaches it in its own error message — so kaibo's three hand-rolled "alternation takes `-E`" notes (a 0.10 workaround from #56) are dead weight. Dropped them from `KAISH_SANDBOX_ADDENDUM`, the cheatsheet, and the `run_kaish` instructions; models type `\|` / `-E 'a|b'` naturally and both work now. This is the "let models do what's natural, stop paying to instruct where the tool self-corrects" principle.

Kept — and sharpened — the `-r`-wants-a-directory note, which guards a **different, still-live** gotcha: `grep -r PATTERN FILE` (a file operand, not a dir) silently finds nothing on 0.11. Filed upstream as **tobert/kaish#105**; the fix there retires this last note too.

## Under the hood / invariants

- chumsky migrated `1.0.0-alpha.8 → 0.13`; two stale transitive regex crates dropped from the tree.
- **TLS invariant intact** — `cargo tree -i aws-lc-rs` is empty; no aws-lc / openssl pulled in.
- Docs updated: `AGENTS.md` kaish-pin note and a user-facing `CHANGELOG.md` entry.

_Cross-family review notes and follow-ups are in the comments below._

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -249,10 +249,19 @@ even for a one-line doc fix.
   `v*` tag. Before tagging: confirm the `kaish-kernel` pin is current (next bullet),
   re-run `docs/sandbox-probes.md` and stamp its "Last run" line, and verify
   `cargo tree -i aws-lc-rs` is empty and the musl binary is `not a dynamic executable`.
-- **kaish pin.** Currently `kaish-kernel = "0.10.0"`. The bump from `0.9.0` was
-  API-compatible — no kaibo call sites changed, identical builtin/tool set; under the
-  hood kaish gained bignum arithmetic (`num-bigint`) and a new regex engine
-  (`regex-bites`). Offline suite green (462 tests), boundary tests still have teeth.
-  The prior `0.8.4 → 0.9.0` step had renamed the `mcp()` config constructors
-  (`IgnoreConfig`/`OutputLimitConfig`/`KernelConfig`) to `agent()` in `sandbox.rs`.
-  Keep this current per the **Working here** kaish-bump discipline before cutting.
+- **kaish pin.** Currently `kaish-kernel = "0.11.0"`. The `0.10.0 → 0.11.0` bump was
+  API-compatible — **zero** kaibo call sites changed. It looks scary in the kaish
+  changelog (four **BREAKING** entries) but none reach kaibo: the new `ToolSchema.raw_argv`
+  and `ExecResult.latch` fields are additive (kaibo builds both via constructors, never
+  struct literals), the `kaish-help` `Fragment` `rank` field only bites code that
+  constructs `Fragment` literals (kaibo doesn't), the latched-`--json` envelope change
+  never fires (kaibo is read-only, so no destructive op ever latches), and the
+  scatter/gather + sed-BRE changes are runtime behavior kaibo has no hand-rolled docs for.
+  The whole agent-facing surface (native collections, `keys`/`values`/`typeof`, the `test`
+  builtin, `help regex`/`help collections`, JSON/JSONL bridges, importance-ranked
+  onboarding) flows in *for free* because kaibo single-sources it from `kaish_kernel::help`
+  — the bump updates the sourced text, no kaibo edits. Under the hood chumsky migrated
+  `1.0.0-alpha.8 → 0.13` (two stale regex crates dropped). Offline suite green (525 tests),
+  boundary tests still have teeth. Precedent that a bump *can* move call sites: `0.8.4 →
+  0.9.0` renamed the `mcp()` config constructors to `agent()` in `sandbox.rs`. Keep this
+  current per the **Working here** kaish-bump discipline before cutting.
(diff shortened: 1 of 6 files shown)


tobert:
### Cross-family review (dogfood)

Per our branch discipline, ran a cross-family pass — a different lineage than wrote the change. Used kaibo's own `consult` with the `deepseek` cast, pointed at the worktree so it verified kaibo's **actual construction sites** rather than trusting the PR summary.

**Verdict: all four kaish 0.11 BREAKING items are zero-impact**, each confirmed against `file:line`:

1. **`ExecResult.latch`** — kaibo touches `ExecResult` at three sites in `src/sandbox.rs` (`::failure()` at :71, pass-through at :303, and `from_result` reading only `code`/`err`/`out_bytes()`/`text_out()` at :337–362). Never reads `.latch`. Double-locked by read-only: no destructive op ever latches.
2. **`ToolSchema.raw_argv`** — constructed only via `::new()`; only field read is `.name` (`sandbox.rs:296`). An added field can't break constructor-only use.
3. **`Fragment.rank`** — `Fragment` appears nowhere in the tree; help is composed via `Recipe` / `SchemaContent` / `get_help`.
4. **scatter/gather + sed BRE** — no kaibo-authored text hard-codes the removed flags; `sed` examples are line-range only.

**One forward-looking finding (not blocking):** the exhaustive `match` on `IgnoreScope` at `src/server/config_resource.rs:411` is the single spot coupling kaibo to kaish's enum surface. A future kaish release adding a variant would be a **compile error** — the *right* (loud) failure mode, not a silent bug — but worth remembering at the next bump. Tracked in the follow-ups comment.

tobert:
### Follow-ups (none block this merge)

1. **[tobert/kaish#105](https://github.com/tobert/kaish/issues/105)** — `grep -r PATTERN FILE` (file operand) silently finds nothing; an `is_file(root)` fast-path fixes it. When it lands and we bump again, kaibo's remaining `-r`-wants-a-directory note can go too.
2. **Next kaish bump:** watch the exhaustive `IgnoreScope` match at `src/server/config_resource.rs:411` (from the review) — a new enum variant is a loud compile error there, easy to adapt.
3. **Before the next kaibo release** (not this PR): re-run `docs/sandbox-probes.md` and stamp its "Last run" line, per the release checklist.
4. **Memory** `kaish-grep-gotchas` already updated to 0.11 reality: BRE `\|` alternation now works (was the reason that memory existed); `-r`-on-a-file remains a live, silent gotcha pending #105.

PR #60, opened 2026-07-05
ci: SHA-pin actions + least-privilege permissions

tobert:
Bootstrapping the release build — dial-in slice (a) of the plan in `docs/releases.md` (PR #25): every `uses:` pinned to a commit digest with a version comment (a tag can be moved, a digest can't — the tj-actions lesson), and permissions go least-privilege per job (`{}` top-level, build reads, publish writes). Bumping the checkout/upload-artifact majors also clears the Node 20 deprecation warnings the baseline run surfaced.

Every SHA↔tag mapping was verified independently against the GitHub API. Validation dispatch from this branch: https://github.com/tobert/kaibo/actions/runs/28746367820 (the publish-job pins — download-artifact, gh-release — stay unexercised until a real `v*` tag; the build job proves the rest). Cross-family review (DeepSeek) in flight; findings will land here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,10 @@
 # fully static with nothing but a Rust toolchain plus zig (for ring's small C/asm).
 # cargo-zigbuild uses `zig cc` as the cross C compiler/linker, which keeps the Linux
 # cross-builds painless and lets one Linux runner emit both arches.
+#
+# Actions are pinned by commit SHA, not tag — a tag can be moved, a digest can't
+# (supply-chain). Each pin carries a `# vX.Y.Z` comment naming the release it points
+# at, for humans and Dependabot.
 name: release
 
 on:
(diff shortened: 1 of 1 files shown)


tobert:
**Cross-family review (DeepSeek, via kaibo `consult`): clean** — every check a confirmation, no findings. Confirmed: upload-artifact v7 immutable-name semantics are a no-op here (unique name per matrix leg); `merge-multiple: true` + `files: dist/*` is the canonical matrix→release pattern under download-artifact v8; gh-release v3's no-tag-autocreate change can't fire (tag-push trigger guarantees the tag); `toolchain: stable` + `targets:` is the documented invocation for a SHA-pinned dtolnay/rust-toolchain; `permissions: {}` breaks nothing (artifact/cache services don't consume GITHUB_TOKEN scopes). All six SHA↔tag pins were also verified independently against the GitHub API before the review.

**Validation caught a real one:** the first dispatch from this branch failed at packaging — `ref_name` = `ci/sha-pin-actions` puts a `/` in the dist name and `mkdir` fails. Exactly the latent footgun the pre-flight review flagged; fixed here (`${GITHUB_REF_NAME//\//-}`) since dispatch-from-branch is how every dial-in slice gets validated. Re-run: https://github.com/tobert/kaibo/actions/runs/28746465173

PR #61, opened 2026-07-05
ci: aarch64 leg on arm64 runners + per-leg smoke runs

tobert:
Bootstrapping the release build — dial-in slice (b) of the plan in `docs/releases.md`: linking isn't running, so every leg now executes its binary (`--version`) on its own hardware — the aarch64 leg moves to `ubuntu-24.04-arm` (free for public repos) so its binary runs natively, `x86_64-apple-darwin` smokes under Rosetta 2, and the musl legs assert the binary is fully static (`ldd`).

**Cross-family review (DeepSeek, via kaibo): clean, all findings HIGH-confidence confirmations** — ziglang PyPI ships aarch64 manylinux wheels; zigbuild same-ISA→musl is the easy case (zig brings its own sysroot), and the new smoke steps are exactly the mitigation for that less-traveled path; glibc's `ldd` "not a dynamic executable" message is architecture-independent; Rosetta 2 is preinstalled on GitHub's arm64 macOS images; the herestring step is Linux-gated so Git Bash never sees it.

Validation dispatch: https://github.com/tobert/kaibo/actions/runs/28746627856 — 4/5 legs green at PR time (both macOS smokes + x86 musl static-assert already passed); the arm leg is mid-build on a cold cache. Slice (c)'s prebuilt-zigbuild item will cut that cold-start.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,8 +33,10 @@ jobs:
           # Linux: fully static via musl. zigbuild handles the cross C toolchain.
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
+          # arm64 hardware (not cross-compiled from x86_64) so the smoke step below
+          # can execute this leg's binary natively, not just link it.
           - target: aarch64-unknown-linux-musl
-            os: ubuntu-latest
+            os: ubuntu-24.04-arm
           # macOS: not "static" (Apple forbids static libSystem) but self-contained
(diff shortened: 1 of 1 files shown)


tobert:
Validation complete: **all five legs green** (run 28746465173 → 28746627856). The aarch64 leg's log shows the point of the slice — `kaibo 0.1.0` printed by the binary executing natively on `ubuntu-24.04-arm`, then the `ldd` static assert passing. Ready to merge.

PR #62, opened 2026-07-05
ci: reproducible archives + prebuilt zigbuild + precise C claim

tobert:
Bootstrapping the release build — dial-in slice (c), the last of the PR-2 hardening from `docs/releases.md`:

- **Reproducible archives.** `SOURCE_DATE_EPOCH` comes from the commit date; the tarball is built with sorted entries/zeroed ownership/clamped mtimes through `gzip -n` (gtar selected on macOS — bsdtar lacks the determinism flags); the zip gets its mtimes pinned, directory entry included. Teeth-tested locally: two runs at different times produce byte-identical tarballs.
- **Prebuilt `cargo-zigbuild`** via SHA-pinned `taiki-e/install-action` v2.82.9 (pin verified against the GitHub API; its manifest ships aarch64-linux) instead of compiling it from source every release; `pip` → `python3 -m pip` while touching the line. Arm leg 4m22s → 3m40s (cache warmth confounds; the structural win is not building a tool at release time).
- **Precise C claim** in AGENTS.md Build & release: the tree is free of *cmake/autotools* C (no aws-lc, no OpenSSL) — ring's small C/asm still compiles, via zig or the platform cc.

**Cross-family review (DeepSeek, via kaibo):** five HIGH-confidence confirmations (`command -v` under `set -e`, shallow-clone `git log -1`, `tar|gzip` under `pipefail` fails correctly, taiki-e arm64 + checksum posture, AGENTS.md wording) and **one actionable finding, closed in a review commit**: 7z has no `--sort=name`, so zip entry order rests on filesystem enumeration — deterministic today only because the archive holds a single file; the boundary is now stated at the site so a second file triggers the sorted-list fix.

Validation dispatches: https://github.com/tobert/kaibo/actions/runs/28746908084 (all five legs green, prebuilt zigbuild proven on amd64+arm64) and https://github.com/tobert/kaibo/actions/runs/28747039327 (tip, comment-only delta, running at PR time).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -132,6 +132,10 @@ jobs:
           cp "$bin" "$dist"/
           # Fix mtimes before zipping so the archive is byte-stable for this commit —
           # including the directory itself (7z stores the dir entry's mtime too).
+          # Entry ORDER is the other determinism axis: 7z enumerates the filesystem
+          # (no --sort=name equivalent), which is only guaranteed stable here because
+          # the archive holds a single file. If this zip ever gains more files, feed
+          # 7z an explicitly sorted list.
           touch -d "@$SOURCE_DATE_EPOCH" "$dist" "$dist"/*
           7z a "$dist.zip" "$dist" >/dev/null
           sha256sum "$dist.zip" > "$dist.zip.sha256"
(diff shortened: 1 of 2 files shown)


tobert:
Tip validation complete: run 28747039327 **all five legs green**. Ready for review.

PR #64, opened 2026-07-05
ci: first CI workflow — offline suite, clippy -D warnings, TLS tripwire

tobert:
Bootstrapping the build: until now the 528-test suite ran only on our machines and the rustls+ring TLS invariant had no automated teeth — release.yml was the repo's only workflow. Every PR and push to main now runs the offline suite, clippy with `-D warnings`, and a `cargo tree` tripwire that fails if aws-lc/openssl/native-tls ever re-enter the dependency tree (asserted on the error text — `cargo tree -i`'s exit code is not reliable for an absent package; teeth proven locally by pointing the guard at `ring`, which is in the tree, and watching it fail).

Rides with the five outstanding clippy lints fixed so `-D warnings` holds from the first run (plus a review commit rewording a doc line instead of indenting mid-sentence prose into an accidental rustdoc bullet). Same SHA pins as release.yml, reused verbatim. This PR's own checks are the live validation of the workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,58 @@
+# The offline test suite, clippy as errors, and the rustls+ring TLS-invariant guard —
+# the fast correctness gate that runs on every pull request and every push to main.
+# release.yml is the cross-platform build/smoke story (tagged releases); this is quick
+# feedback on every change so nothing lands with a red suite, a new lint, or a C
+# crypto/TLS dependency that snuck back into the tree.
+name: ci
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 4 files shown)


tobert:
**Cross-family review (DeepSeek, via kaibo): no blockers.** Confirmed: `.then_some` swap is behaviorally identical (value is a static literal, no side effects); `OtelLayer<S>` alias is crate-private with one pattern-destructuring call site — zero API impact; the suite is genuinely offline (all 11 live tests `#[ignore]`d, the TLS-telemetry test hits an unroutable localhost by design); no feature axes exist to matrix. On the tripwire's error-text assertion: acknowledged fragile-but-reasoned — cargo's wording has been stable for years, the workflow comment documents the trade-off, and the release build's static-link failure is the backstop if cargo ever rewords it. `cargo tree -i` is the right scope (transitive, includes dev-deps). One low-value absence noted (`cargo doc` step) — deferred until the doc surface grows.

PR #65, opened 2026-07-05
feat(context): steer citations to shell reads; opt-in cat -n numbering for house rules

tobert:
## Why

The first all-local `explore` (zorak cast, 2026-07-05) surfaced a citation-honesty hole: the gemma4-e4b explorer answered in one completion, zero tool calls, by quoting the `[context]`-inlined `AGENTS.md` verbatim — under **invented line numbers** (`AGENTS.md:153-164`; real text, wrong lines). The inlined house-rules block looks citable, but it carried no line numbers, so any `file:line` against it was a guess dressed as evidence.

## Decisions

- **House rules are guidance, not source** (Amy's call). Asking a model to quote `AGENTS.md` accurately by line is unfair, and `cat -n` gutters on prose directives are clutter. The **default stays plain prose**; the house-rules framing carries the fix instead — it now tells the model to ground every `file:line` it cites in a file it read through the shell, including a guidance file if the answer needs to cite one precisely. In practice that re-read should be rare for kaibo.
- **`[context] numbered = true`** is the opt-in for operators whose context files *are* the kind a consult should cite directly. It reuses `attach.rs::number_lines` (now `pub(crate)`) so the numbered form has one source of truth with attachment inlining.
- Files the sibling finding from the same probe (provenance footer credits the explorer even when zero sweeps ran) as an open `docs/issues.md` entry — separate fix.

## Tests

- `context.rs`: default-plain and opt-in-numbered assertions (the numbered one was written failing-first against the un-numbered `section()`); all existing literals updated for the new field.
- `tests/config.rs`: `numbered` parses and defaults off.
- Full offline suite green (21 binaries).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -357,6 +357,17 @@ the git log. Each later release appends a new section at the top.
   against the canonical names kaibo advertises. The default is now resolved before
   comparison, so the right cast is flagged however you named it.
 
+- **Models are steered away from inventing line numbers for the house-rules
+  context.** Observed live (2026-07-05): a local explorer answered by quoting the
+  inlined `AGENTS.md` verbatim under made-up line numbers — an answer that *looked*
+  grounded but wasn't checkable. The house-rules framing now tells the model to
+  ground every `file:line` it cites in a file it read through the shell, including
+  a guidance file it wants to cite precisely. For operators whose context files
+  *are* the kind a consult should cite directly, a new `[context] numbered = true`
+  knob inlines the sections with `cat -n`-style line numbers (the same form
+  attachments take); the default stays plain prose — house rules are guidance,
+  not source.
+
 ### Security
 
 - **Read-only is structural, not best-effort.** kaibo compiles in only kaish's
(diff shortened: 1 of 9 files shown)


tobert:
Cross-family review (DeepSeek v4-pro via kaibo `consult`, whole-file read of the worktree, no diff): **clean, "ship it."**

Checked and confirmed: `merge_context` is the only live construction path (no bypass of the new field); `deny_unknown_fields` makes a typo'd `numberd` a loud load error; file-only knob matches the `[prompts]` precedent (env block touches only the two file lists); `trim_end`-then-`number_lines` keeps numbers exact to the on-disk file (`split_inclusive` adds no phantom tail — same ordering contract as `Attachment::wrapped_text`); `unwrap_or(false)` means existing configs are byte-identical in behavior.

One subtlety it raised and judged correctly traded: the framing sentence is static, so on the `numbered = true` path a model may still cite the inlined block directly — which is fine, because the numbers are real there; the steer matters on the default path where they don't exist. A config-conditional preamble variant wasn't worth the complexity.

tobert:
This isn't worth the churn. It was a small local model being asked an unfair question.

PR #69, opened 2026-07-13
feat(release): born-signed releases — cosign keyless + SLSA provenance + SPDX SBOM (plan PR 3)

tobert:
## Why

Plan PR 3 of [`docs/releases.md`](docs/releases.md): the transparency payoff. v0.2.0-rc.1 proved the tag→release leg precisely so the real v0.2.0 could be **born signed** — this PR is that signing layer, all GitHub-native, no middleman, no long-lived key anywhere.

## What

Everything lands in the **tag-gated publish job** — a `workflow_dispatch` smoke run never mints an OIDC identity or a signature. Build legs keep `contents: read`; the publish job adds `id-token: write` + `attestations: write`.

Layout (decided with Amy 2026-07-13):

- **One signed aggregate `checksums.txt`** (cosign keyless): verify the manifest once and `sha256sum -c` covers any file it lists. Ships **both signature shapes** — `.sig`/`.pem` (works on cosign v2 and v3) and a self-contained `.sigstore.json` protobuf bundle (offline `verify-blob --bundle` on v3). The per-artifact `.sha256` sidecars stay; the README download one-liner uses them.
- **Per-artifact SLSA build provenance** via `actions/attest-build-provenance`, stored in GitHub's attestation store — `gh attestation verify <file> -R tobert/kaibo` is the one-command path, zero extra release assets.
- **One SPDX SBOM from `Cargo.lock`** — a bare Rust binary carries no dependency metadata for syft to read; `cargo-auditable` per-binary SBOMs are a tracked follow-up in `docs/issues.md` (touches all five build legs incl. zigbuild, deserves its own validation).
- **README "Verify a download"** with the exact invocations — including the identity flags keyless verification *requires* (Gemini's plan-review finding: a bare `verify-blob` cannot work for keyless; signing nobody can verify is theater) and `--ignore-missing` so a single-file download checks clean.
- AGENTS.md cutting-a-release checklist gains the post-publish step: run the README's own verification commands against a fresh asset.

## Diligence

- New action pins (`cosign-installer` v4.1.2, `attest-build-provenance` v4.1.1, `sbom-action` v0.24.0) digest-verified through **two independent paths** (GH API + `git ls-remote` peeled tags); input names verified against each `action.yml` at the pinned SHA.
- cosign-installer defaults to **cosign v3.0.6**, which changed sign-blob defaults to the new bundle format — confirmed against `sign_blob.go` at that tag that `--output-signature`/`--output-certificate` still write files alongside `--bundle`, so the legacy pair and the bundle coexist in one invocation.
- `actionlint` clean, YAML parses.

## Validation plan

The signing steps only run on a tag, so a branch dispatch can't exercise them. Plan: tag an rc (e.g. `v0.2.0-rc.2`) to fire the signing path end-to-end, then verify the assets **with the README's own commands**, the way a user would — before the real v0.2.0.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,11 @@
 # Actions are pinned by commit SHA, not tag — a tag can be moved, a digest can't
 # (supply-chain). Each pin carries a `# vX.Y.Z` comment naming the release it points
 # at, for humans and Dependabot.
+#
+# Releases are born signed: the publish job cosign-signs an aggregated
+# checksums.txt (keyless — the identity IS this workflow at this tag), attaches
+# SLSA build provenance per artifact (`gh attestation verify`), and ships an SPDX
+# SBOM from Cargo.lock. Verification commands live in the README.
 name: release
 
 on:
@@ -148,6 +153,8 @@ jobs:
             kaibo-*.tar.gz*
             kaibo-*.zip*
 
+  # Signing, provenance, and the SBOM all live here in the tag-gated publish job —
+  # a `workflow_dispatch` smoke run never mints an OIDC identity or a signature.
   release:
     name: publish release
     needs: build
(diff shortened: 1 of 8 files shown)


tobert:
## Cross-family review (DeepSeek `consult` via kaibo, holistic — whole files attached, no diff)

**Verdict: merge-ready, no blocking issues.** Answer carried the provenance footer (cast `deepseek`, explorer `deepseek-v4-flash`, synth `deepseek-v4-pro`).

Confirmed: step ordering sound (every step's inputs exist when it runs); checksum globs and attest subject globs mutually consistent — the per-artifact `.sha256` sidecars are correctly excluded from both, `checksums.txt` is attested but not self-hashed, and the cosign sidecars are correctly *not* attested (their authenticity rides the transparency log, not build provenance); permissions exactly minimal (no `actions:`/`packages:` creep, build legs untouched); all five action pins verified; the `--ignore-missing` and plain-filename details in the verify flow both check out; all six files cross-consistent.

**One real finding, fixed in `e6c824c`:** the README's cosign `--certificate-identity` hardcoded `v0.2.0` — a copy-paste trap for anyone verifying any other tag (loud identity-mismatch failure, but needless friction). Now `vX.Y.Z`, unmissable.

**Accepted as-is:** the CHANGELOG's "locked dependency tree the binaries were built from" phrasing (technically correct via `--locked`; the workflow comment and issues.md carry the Cargo.lock-vs-binary nuance).

**First-tag watch item it named (no workflow change needed):** the cosign sign step makes live Fulcio/Rekor calls — a transient Sigstore outage fails the publish job; the remedy is re-running the job, which the rc validation will implicitly rehearse.

PR #70, opened 2026-07-13
fix(release): bundle-only blob signature — cosign v3 ignores the legacy pair (rc.2 kink)

tobert:
## What rc.2 taught us

`v0.2.0-rc.2` fired the full signing path green first try, and **everything that shipped verifies**: `gh attestation verify` exit 0, `cosign verify-blob --bundle` Verified OK under the tag identity, `sha256sum -c --ignore-missing` OK — all run against the real assets, the way a user would.

The one kink: cosign v3 in new-bundle-format mode **warns that `--output-signature`/`--output-certificate` are deprecated and ignores them** — the planned `.sig`/`.pem` pair never shipped (the `sign_blob.go` write path we source-checked does write them; a CLI-layer guard nils the flags first in bundle mode — the [run log](https://github.com/tobert/kaibo/actions/runs/29284271510) has the receipts).

## The fix

Embrace bundle-only rather than pin cosign back to v2 for a shape it deprecates:

- Sign step emits just `checksums.txt.sigstore.json` — self-contained (cert + signature + Rekor entry), verifies **offline** with cosign ≥ 3.
- README's cosign section documents the bundle invocation as *the* command — the exact flow already proven against rc.2. Operators without cosign keep `gh attestation verify` as the no-cosign path.
- `docs/releases.md` records the rc.2 validation + kink story.
- Version → `0.2.0-rc.3`, which validates the bundle-only invocation end-to-end.

actionlint clean.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -202,17 +202,14 @@ jobs:
       # the Rekor transparency log — no long-lived key anywhere. Verification
       # REQUIRES the identity flags (a bare verify-blob cannot work for keyless);
       # the exact invocation lives in the README's "Verify a download" section.
-      # Both signature shapes ship: .sig/.pem verifies on cosign v2 and v3, and
-      # the .sigstore.json protobuf bundle is self-contained (cert + signature +
-      # Rekor entry) for a v3 `verify-blob --bundle` that works offline.
+      # One signature shape: the .sigstore.json protobuf bundle, self-contained
+      # (cert + signature + Rekor entry), so `verify-blob --bundle` works offline.
+      # cosign v3 IGNORES the legacy --output-signature/--output-certificate pair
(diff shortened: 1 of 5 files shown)


tobert:
## Cross-family review (DeepSeek `oneshot` via kaibo — glance-scaled to the fix)

Clean on all three questions: (1) the sign + verify invocations are the canonical cosign v3 surface, correct for a real operator; (2) dropping `.sig`/`.pem` is defensible — v3 ignores those flags in bundle mode, so the outputs were dead anyway, and `gh attestation verify` remains the no-cosign path; (3) nothing weakens the trust story — the reviewer called bundle-only **stronger** than the two-shape plan (signature + certificate + transparency entry in one self-contained file, fully offline verification, no live Rekor lookup).

PR #71, opened 2026-07-13
feat(release): ghcr container image — multiarch distroless, born signed, built to be COPY'd (plan PR 4)

tobert:
## Why

Plan PR 4 of [`docs/releases.md`](docs/releases.md), pulled forward: the ghcr image as a first-class distribution path, landing *after* signing so it's **born signed**. The headline design goal (Amy): the fully-static binary makes the image a **COPY source** — `COPY --from=ghcr.io/tobert/kaibo:latest /usr/local/bin/kaibo /usr/local/bin/kaibo` is a one-line kaibo install into any devcontainer or image, zero libc/package concerns.

## What

- **`Dockerfile`** — the image contract: `FROM gcr.io/distroless/static:nonroot` **pinned by manifest digest** (same discipline as action pins), `ARG TARGETARCH` COPY of the per-arch musl binary, `WORKDIR /work` as the canonical read-only project mount, exec-form stdio ENTRYPOINT, OCI source/description/license labels.
- **New `image` job** in release.yml: multiarch amd64+arm64 from the already-built musl artifacts (pure COPY — no QEMU), smoke-boots the amd64 image (`--version`) *before* any push, then tag-gated: ghcr login, metadata-action tags (full semver always; `major.minor` + `latest` only for stable — an rc never becomes someone's `latest`), push, `cosign sign` on the manifest digest, `attest-build-provenance` with `push-to-registry`. The job's build+smoke also runs on `workflow_dispatch`, so this branch's dispatch validates it; **no dispatch ever publishes or signs**.
- **README container section**: COPY recipe first, then the `docker run --rm -i -u … -v "$PWD:/work:ro"` MCP registration with the two documented footguns — `-i` is load-bearing for a stdio server (silent EOF-exit otherwise, and distroless has no shell to debug into), and UID mapping (`-u` / podman `--userns=keep-id`). Keys passed by name. Image verification via `gh attestation verify oci://…` or `cosign verify`.
- CHANGELOG (Added), releases.md PR-4 realization, issues.md: pipeline pointer advanced + a new **graceful no-stdin error** entry (it's a kaibo *code* change — in a shell-less image the entrypoint IS the binary — deferred out of pipeline YAML with a concrete fix shape).
- Version → **0.2.0-rc.4** (Amy: another rc validates the containers).

## Proven locally before CI

- podman build + run: `kaibo 0.2.0-rc.3`, uid 65532, `/work`, correct entrypoint.
- **COPY contract**: `COPY --from` into `ubuntu:24.04` → binary runs.
- **MCP flow**: `initialize` handshake through the container with `--userns=keep-id` + `:ro` mount answers correctly (honest "Setup needed" instructions with no key passed).

## Diligence

Four docker action pins digest-verified via two independent paths (GH API + `git ls-remote`); distroless digest fetched from gcr.io directly. actionlint clean. Branch `workflow_dispatch` validation in flight; rc.4 validates the publish path end-to-end after merge, including a live consult **through the pulled image** (the distroless TLS-cert proof).

**One-time operator step after the first push (Amy):** flip `ghcr.io/tobert/kaibo` package visibility to **public** (workflow-created packages default private) — anonymous `docker pull` and the COPY recipe need it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,9 @@
 # Releases are born signed: the publish job cosign-signs an aggregated
 # checksums.txt (keyless — the identity IS this workflow at this tag), attaches
 # SLSA build provenance per artifact (`gh attestation verify`), and ships an SPDX
-# SBOM from Cargo.lock. Verification commands live in the README.
+# SBOM from Cargo.lock. The image job ships the same binaries as a multiarch
+# ghcr.io/tobert/kaibo image (distroless, non-root), signed and attested by the
+# same machinery. Verification commands live in the README.
 name: release
 
 on:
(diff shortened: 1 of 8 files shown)


tobert:
## Cross-family review (DeepSeek `consult` via kaibo, holistic — whole files, no diff)

**Verdict: solid — staging logic, digest wiring, base-image choice, signing posture, and docs all coherent.** Findings and dispositions:

- **HIGH → resolved as non-bite by verify-first:** the reviewer couldn't confirm whether metadata-action suppresses the `{{version}}` tag for prereleases (which would have left rc.4 reachable only by sha tag). Read `meta.ts procSemver` at the pinned SHA: a prerelease **degrades every semver pattern to `{{version}}`** and holds `latest` false — rc.4 gets its version tag, and neither `0.2` nor `latest` can ever point at an rc. Workflow comment now states the verified mechanism (`e43ec75`).
- **MEDIUM:** ghcr packages created by a workflow default **private** — README commands fail until the one-time visibility flip (already flagged as Amy's post-first-push step in the PR body). Added the reviewer's suggested README line: pulling needs no login.
- **LOW, accepted:** job-level permissions are broader than a dispatch run uses (structural — the job must build+smoke on dispatch; the publishing steps are gated). Smoke rebuilds amd64 on tag runs (~seconds, COPY-only). Podman `--userns=keep-id` advice — our own local test ran the MCP handshake through the container with exactly that flag and a `:ro` mount, so it stands until a podman user reports otherwise.

**Branch dispatch validation: all six jobs green** — the image job built the multiarch manifest, smoke-booted the amd64 container, and correctly skipped login/sign/attest off-tag: https://github.com/tobert/kaibo/actions/runs/29288345880

PR #72, opened 2026-07-13
feat(release): per-tag get+verify block on every release page

tobert:
## Why

Two field reports from Amy reading the rc.4 release + package pages: the release page should link the ghcr package, and her first pull attempt grabbed a `sha256-*` tag — which is a **cosign signature artifact**, not an image (podman refused it with the sigstore-bundle mediaType error). Both are education-at-the-right-surface problems; the release page is the surface.

## What

- **`.github/release-body.md`** — a template the publish job renders with `envsubst` (TAG = `v0.2.0`, VERSION = `0.2.0`) and passes to softprops as `body_path`; softprops documents that a body **prepends** above `generate_release_notes`. The block: package link, `docker pull` + `COPY --from` lines, both `gh attestation verify` one-liners (file and `oci://`), the cosign bundle verify + `sha256sum -c` — **every command carrying that release's exact tag**, ready to copy (the README must keep `vX.Y.Z`; the release page never should). Plus the warning that `sha256-*` package tags are signatures riding alongside the image.
- CHANGELOG (Added), releases.md realization note.

## Validation

- Rendered locally with rc.4 values: zero unexpanded variables, actionlint clean.
- **The rendered block is live on [rc.4's release page](https://github.com/tobert/kaibo/releases/tag/v0.2.0-rc.4) right now** (via `gh release edit`) — real markdown on a real page, exactly what the workflow will produce. The workflow path itself gets exercised by the next tag (rc.5 or v0.2.0).

### DIFF
--- a/.github/release-body.md
+++ b/.github/release-body.md
@@ -1,0 +1,46 @@
+<!--
+Rendered onto every release page by release.yml, prepended above GitHub's
+auto-generated notes. envsubst fills TAG (the git tag, v-prefixed) and VERSION
+(the same without the v — how the image is tagged). This is the one surface
+where every placeholder is concrete — the commands below carry the exact tag,
+ready to copy — so keep it that way: no angle-bracket stand-ins. HTML comments
+don't render on GitHub, so this block ships invisibly (envsubst reaches inside
+it, which is why it names the variables in prose).
+-->
+## Get it
+
+Grab your platform's archive below ([install + MCP registration](https://github.com/tobert/kaibo#installation)) — or take the [container image](https://github.com/tobert/kaibo/pkgs/container/kaibo):
+
+```sh
+docker pull ghcr.io/tobert/kaibo:${VERSION}
+```
+
+```dockerfile
+# one-line install into a devcontainer or any image — the binary is fully static
+COPY --from=ghcr.io/tobert/kaibo:${VERSION} /usr/local/bin/kaibo /usr/local/bin/kaibo
+```
+
+(Pull version tags like `${VERSION}` — the `sha256-*` tags on the package page are
+cosign signature/attestation artifacts riding alongside the image, not images.)
+
+## Verify it
+
+Every artifact carries SLSA build provenance — any downloaded file, one command:
+
+```sh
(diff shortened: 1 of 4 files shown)


tobert:
## Cross-family review (DeepSeek `oneshot` via kaibo — glance-scaled)

Clean: the render step's env scoping, line continuation, and envsubst explicit-var-list are all idiomatic and injection-safe (only the two named vars substitute; anything else stays literal). One suggestion applied in `c75ea65`: the verify commands use bare filenames, so the page now says to run them from the download folder. Edge cases it raised (a bare `v` tag, a `v-`-prefixed tag producing an invalid image ref) are legal-but-unrealistic tag shapes we don't cut — noted, not guarded.

The rendered block is live on [rc.4's release page](https://github.com/tobert/kaibo/releases/tag/v0.2.0-rc.4) (refreshed to match this revision) — the page as it stands is byte-what the workflow will produce for the next tag.

tobert:
## Addendum: publish-order fix (rode this PR — `99994f1`)

Amy copy-pasted the package page's install box and got a mediaType refusal: GitHub advertises the **most recently published version**, and cosign's signature bundle (a `sha256-<image-digest>` tagged artifact in the same package) lands after the image. Deterministic fix, not a sort-key gamble: tag runs now push the image **by digest** (`push-by-digest`, the documented buildx pattern), sign + attest the digest, then apply the version tags as the job's **final act** via `buildx imagetools create` — a tag-only copy of the signed manifest list. Newest published version = always pullable. The metadata `sha` tag is now dispatch-only, so releases publish clean version tags.

**Cross-family glance (DeepSeek):** endorses all four load-bearing claims — imagetools create points tags at the same index digest without mutation (per-platform + attestation manifests intact); the digest-only window has no practical GC risk on ghcr and the signature already references the digest; signing before tags exist weakens nothing (both verifiers resolve tag→digest at verify time); `outputs.digest` is identical under push-by-digest vs tagged push.

Branch dispatch validating the new build path (outputs line runs with `push=false` off-tag): https://github.com/tobert/kaibo/actions/runs/29290860024

PR #79, opened 2026-07-17
fix(batch): flag truncated batch answers instead of passing them off as clean (#75)

### LINKED ISSUE #75: gemini batch result surfaced truncated chain-of-thought as the answer (no verdict, no truncation signal)
Seen 2026-07-16 while running the kaish GH #95 lexer-rewrite review pair.

## What happened

Submitted a 1-prompt batch on cast `gemini-batch` (model `gemini-pro-latest`) with two attached files (~4,400-line `lexer.rs` + ~500-line test file) and a long code-review prompt asking for line-cited findings ranked by severity and a ship/no-ship verdict.

Handle: `gemini/batches/laaxe5t0oa9181hhc1y65282fow9fksnpz6i` (durable — should still be inspectable).

`job_get` returned a "Batch complete — 1 result(s)" payload whose item text:

- **starts mid-sentence**: `has_bracket_pair` is true. `!has_star_or_question` is true. …`
- **reads as raw reasoning**, not a final answer ("Wait! The DFA tracks EXACTLY whether…", "Let's look at Token::GlobWord again."),
- **ends mid-analysis** ("The parser will see `Int(1)`, `MinusAlone`.") with none of the requested structure — no findings list, no severities, no verdict,
- carried **no error, warning, or truncation marker** anywhere in the result.

The same review shape on the `deepseek` consult leg completed normally, so the prompt itself was answerable.

## Why this is worth a fix

The failure is silent: the result *looks* like a (rambly) review, so a caller skimming for findings can mistake a truncated thought-stream for "the model found little." A review pipeline treating this as a completed review gets a false negative on the whole surface.

Two things seem to be going on:

1. **Truncation wasn't surfaced.** If the response hit `MAX_TOKENS` (or any non-STOP finish reason), the result trailer should say so loudly rather than presenting the text as a normal completion.
2. **Thought parts leaked as the answer.** Gemini responses separate `thought` parts from answer parts; text that both starts and ends mid-sentence with no conclusion smells like the answer part was empty (budget exhausted during thinking) and reassembly fell back to — or failed to filter — thought content. If the answer part is empty, saying "model produced no answer (finish_reason=…)" beats emitting the thinking.

## Suggestions (take or leave)

- Propagate `finish_reason` per batch item and flag anything ≠ STOP in the rendered result.
- Filter or clearly label `thought` parts in batch result assembly; never present them as the item's answer.
- Possibly: raise/expose `max_output_tokens` for batch review workloads, since attached-file reviews at max thinking are exactly the long-output case.

Happy to re-run the same batch for forensics if useful — the prompt is reproducible.

tobert:
Closes #75.

## The bug, ground-truthed

A `gemini-batch` code review (gemini-pro-latest, ~4,400-line `lexer.rs` attached) came back as a rambly chain-of-thought fragment — starts mid-sentence, ends mid-analysis, no findings, no verdict — carrying **no error or truncation marker**. A caller skimming for findings could mistake it for "the model found little."

I fetched the durable handle from the issue (`gemini/batches/laaxe5t0oa9181hhc1y65282fow9fksnpz6i`). It shows exactly why:

- `finishReason: **MAX_TOKENS**` — the response was truncated at the output-token budget.
- `usageMetadata`: `thoughtsTokenCount: 31455` + `candidatesTokenCount: 1309` ≈ the 32768 `BATCH_MAX_TOKENS_FLOOR` — **thinking consumed essentially the whole budget**.
- The candidate's parts: part[0] is `thought:true` (correctly filtered today); part[1] has **no `thought` marker**, carries a `thoughtSignature`, and is trailing reasoning. The old `gemini_message_text` keeps any non-thought text part and returned it as the answer, reading `finishReason` nowhere.

The old comment — *"a final answer part may still carry a `thoughtSignature`… so it stays"* — is the exact assumption that broke: under `MAX_TOKENS`, a signatured non-thought part is trailing reasoning, not an answer.

## The fix (decisions, with Amy)

- **Gate on the finish reason, both providers.** A shared `finish_gated_answer` composes each item from its text + a normalized finish signal (Gemini `finishReason`, Anthropic `stop_reason` — the Anthropic path had the **identical latent bug**). Clean finish → unchanged; truncated / policy-halted → partial text kept under a loud `⚠️ INCOMPLETE` banner naming the reason; truncated-with-no-answer → an honest per-item failure instead of a blank success. Partial text is kept, not dropped, so a genuinely-clipped real answer stays useful — but it announces itself.
- **The signal is the finish reason, not the parts.** A `thoughtSignature` part is a real answer under `STOP` but trailing reasoning under `MAX_TOKENS`; they're structurally indistinguishable, so the gate is the finish reason.
- **Leave the 32K floor; steer the prompt.** Per-model output ceilings differ, so a blind global bump risks 400s. Instead the batch preamble now orders the written answer ahead of the depth (reasoning and answer share one output budget), and `docs/config.example.toml` documents raising a batch cast's `max_tokens` per-slot for long reviews.

## Tests & verification

- Ground-truthed offline tests: Gemini leaked-reasoning (the real payload's shape), only-thoughts→failure, clean `STOP`; Anthropic `max_tokens`→banner and empty→failure; the pure gate/compose contract (incl. the policy-halt and clean-empty paths). **Verified they fail on the pre-fix behavior.**
- Drove the real 303 KB payload through `parse_gemini_poll` → `render_poll` and confirmed a caller now sees the `⚠️ INCOMPLETE — finishReason=MAX_TOKENS` banner with the fragment kept below.
- Full offline suite green; clippy clean.

## Review

Cross-family (DeepSeek via kaibo's own `consult`) review in flight; will fold in findings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -434,6 +434,20 @@ the git log. Each later release appends a new section at the top.
 
 ### Fixed
 
+- **A truncated batch answer no longer masquerades as a finished one.** When a batch
+  item hit its output-token budget mid-response — most often a big attached-file review
+  where max-effort thinking spends the whole budget before the answer is written — kaibo
+  presented whatever text came back as a clean result, with no signal it was cut off. A
+  caller skimming for findings could mistake a truncated reasoning fragment (no verdict,
+  no structure) for "the model found little." kaibo now reads each item's finish reason
+  (Gemini `finishReason`, Anthropic `stop_reason`): a clean finish is unchanged, but a
+  truncated or policy-halted one is flagged — the partial text is kept under a loud
+  `⚠️ INCOMPLETE` banner naming the reason, and an item that produced no answer at all
+  becomes an honest per-item failure instead of a blank success. The batch preamble also
+  now steers the model to write its conclusion first (reasoning and answer share one
+  output budget), and the example config documents raising a batch cast's `max_tokens`
+  for long reviews. (GH #75)
+
 - **A stalled backend can no longer hang a call overnight.** An interactive
   `consult` / `explore` / `oneshot` now runs under a whole-call wall-clock deadline
   (`call_deadline_secs`, default 1 hour; env `KAIBO_CALL_DEADLINE_SECS`), independent
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -381,6 +381,16 @@ batch    = true
 synth    = "gemini/gemini-pro-latest"        # batch-capable synth (Anthropic | Gemini); Pro is batch-only
 # Equivalent, spelled directly on the slot (no cast-level sugar):
 #   synth = { backend = "gemini", id = "gemini-pro-latest", lane = "batch" }
+#
+# OUTPUT BUDGET. Batch maxes the knobs: high effort + a 32768-token output-token FLOOR (it
+# never skimps on the cheap async lane). Reasoning bills against that same output budget, so a
+# big attached-file review at max thinking can spend the budget thinking and truncate the
+# answer — kaibo flags such a result INCOMPLETE with its finishReason rather than passing off
+# the cut-off fragment as a clean answer. Give batch more answer headroom by raising
+# `max_tokens` on the synth slot; it's a floor, so a higher value wins, and mind each model's
+# output ceiling (Gemini Pro is generous; some backends cap lower). `effort` / `thinking_budget`
+# ride the same slot if you want to tune reasoning depth:
+#   synth = { backend = "gemini", id = "gemini-pro-latest", lane = "batch", max_tokens = 65536 }
 
 # --- DELIBERATE casts: an interactive explorer PLUS an offline synth. `deliberate` runs
 #     the explorer live to build a cited dossier, then hands dossier+question to the offline
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -375,6 +375,64 @@ fn message_text(message: &Value) -> String {
         .unwrap_or_default()
 }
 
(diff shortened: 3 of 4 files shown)


tobert:
## Cross-family review — DeepSeek (via kaibo's own `consult`)

Ran a holistic, no-diff review with cast `deepseek` (explorer `deepseek-v4-flash` / synth `deepseek-v4-pro`), the whole `src/batch.rs` attached. **Verdict: correct, consistent across both provider paths, covers the edge cases, fits conventions. No gaps found.** Highlights it verified:

- **Clean-set correctness.** Anthropic's five `stop_reason` values (`end_turn`, `max_tokens`, `stop_sequence`, `tool_use`, `refusal`) — the two clean ones are right for a toolless item; `tool_use`/`refusal`/future values correctly fall to the abnormal catch-all. Gemini's `FINISH_REASON_UNSPECIFIED`/`OTHER`/`MALFORMED_FUNCTION_CALL` are genuinely abnormal and correctly flagged.
- **Empty→Err is a strict improvement** over the old `Ok("")` (which rendered as an empty `## [N]` a caller could misread as "no findings").
- **Provider-error paths correctly bypass the gate** (a provider error means the model never ran — no partial text to banner-wrap), and the cancel-with-partials recovery composes orthogonally with per-item gating.
- **Banner is operator-facing**, so naming the problem directly is the right tone (the positive-framing rule is for model-facing text).

One optional idea it raised — treat an *absent* Anthropic `stop_reason` as abnormal (the API always sends it) — it also called the current behavior "defensible": the empty-text check is the safety net, and a transient upstream quirk that drops the field shouldn't sink a valid answer. Leaving as-is.

PR #80, opened 2026-07-18
fix(shaping): route every Gemini id to thinkingLevel so the effort lever reaches Gemini

tobert:
## What & why

kaibo's Gemini casts all name 3.x-line models (default `gemini` synth `gemini-3.5-flash`, `gemini-batch` Pro synth `gemini-pro-latest`, Flash-Lite explorer `gemini-flash-lite-latest`), but the thinking-knob classifier `is_gemini3_level` matched only the bare `gemini-3` / `gemini-3-*` form. So the `-latest` aliases (no `gemini-3` substring at all) and the dotted 3.x ids (`gemini-3.1-pro-preview` — dot after the 3) fell through to the retired 2.5-era `thinkingBudget` path, which pins depth to a fixed number and **silently drops the per-role `effort` lever**. Setting `effort = "low"`/`"high"`/`"max"` on an interactive Gemini slot had no effect — no crash, since both knobs are accepted on the wire, just the wrong one sent.

## Decision

Google's [current thinking doc](https://ai.google.dev/gemini-api/docs/thinking) lists `thinkingLevel` (`minimal|low|medium|high`) across the whole 3-line — 3-pro, 3-flash, **3.5-flash**, 3.1-pro — and no longer documents `thinkingBudget`; that's the legacy 2.5-era knob. The prior "keep 3.5-flash on budget" pin rested on a probe that only showed budget *works*, never that level *fails* — and if both work, level is strictly better because it carries the effort lever.

So the fix isn't a wider classifier, it's a **collapse** (Amy's call): kaibo targets the Gemini **3-line and newer**, stated plainly. `ProviderKind::Gemini` always resolves to `GeminiLevel`; `is_gemini3_level` and the now-dead `GeminiBudget` `ThinkingStyle` variant are removed. A pre-3.x id emits a level and fails loud at the provider — the honest "unsupported" signal — over silently mis-shaping.

## Commits

1. **Collapse the classifier.** TDD: a failing test pinning the configured ids' `effort → thinkingLevel` came first (RED on `gemini-flash-lite-latest`), then the collapse turned it green. Dependent tests using 2.5 ids as fixtures moved to 3.x; the engine per-arm straddle test — which leaned on the now-defunct *intra-Gemini* budget/level split — was re-aimed at a stronger cross-*provider* straddle (Anthropic-adaptive synth vs Gemini-level explorer, structurally disjoint param trees).
2. **Fix the validation fallout** (caught by the GPT cross-family review): the `thinking_budget < max_tokens` load check was gated on `matches!(kind, Anthropic | Gemini)`. With Gemini no longer sending a budget, that gate falsely refused a valid Gemini slot (e.g. `max_tokens=4096` under the default `thinking_budget=8192`), and it had always wrongly refused Anthropic *adaptive* slots too (budget equally inert). Now gated on the resolved shape's `sinks_thinking_budget()` at both sites — only a style that actually sends a budget can starve the answer. Plus the doc drift the review flagged (effort-taker lists omitting Gemini; the batch example claiming `effort` tunes depth when batch forces it).
3. **Track** a cosmetic render-accuracy nit surfaced by the focused review (doc-only).

## Review (dogfooded, cross-family)

- **DeepSeek** (deep, commit 1): clean — verified all five ids route to `thinkingLevel`, zero dead-code hits, render helpers correct, tests have teeth; independently confirmed `includeThoughts: true` is **not** stale (the batch #75 truncation detector depends on those thought parts arriving).
- **GPT / OpenRouter** (deep, commit 1): surfaced the two "Medium" follow-ups above — the stale budget validation (fixed in commit 2) and the batch-render lane-blindness (pre-existing, tracked).
- **DeepSeek** (focused, commit 2): confirmed the `sinks_thinking_budget()` gating is correct and complete — no 400/starve path, both sites consistent, batch clamps safely, no missed site.
- Gemini cast was 503-ing throughout (provider overload), hence GPT stood in as the second family.

## Not in scope (tracked in `docs/issues.md`)

The `kaibo://config` inert-tunables render has a cluster of pre-existing display-only gaps (lane-blindness to a batch slot's forced effort; shape resolution that skips the `defaults.thinking_style` fallback). All cosmetic, no wire effect — tracked for a follow-up that reconciles the render's shape resolution with `slot.tunables`.

## Verification

`cargo test` (595 pass), `cargo clippy --all-targets` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -440,11 +440,21 @@ the git log. Each later release appends a new section at the top.
   the bare `gemini-3-…` form — so the `-latest` aliases and the dotted 3.x ids (e.g.
   `gemini-3.1-pro-preview`) fell through to the retired 2.5-era `thinkingBudget` path,
   which pins depth to a fixed number and silently drops the per-role `effort` lever.
-  Setting `effort = "low"` (or `"high"`, `"max"`, …) on a Gemini slot had no effect. kaibo
-  now routes every Gemini id to `thinkingLevel` — the knob Google's current API documents
-  across the whole 3-line — so the effort you configure is the reasoning depth Gemini
-  runs at. (kaibo targets the Gemini 3-line and newer; the legacy 2.5 budget knob is no
-  longer modeled — a pre-3.x id fails loud rather than silently mis-shaping.)
+  Setting `effort = "low"` (or `"high"`, `"max"`, …) on an interactive Gemini slot had no
+  effect. kaibo now routes every Gemini id to `thinkingLevel` — the knob Google's current
+  API documents across the whole 3-line — so an interactive Gemini slot's configured
+  effort is the reasoning depth it runs at. (The batch lane still runs at its own forced
+  max effort, by design. kaibo targets the Gemini 3-line and newer; the legacy 2.5 budget
+  knob is no longer modeled — a pre-3.x id fails loud rather than silently mis-shaping.)
+
+- **A Gemini or Anthropic-adaptive slot with a large `thinking_budget` no longer fails to
+  load.** The `thinking_budget < max_tokens` load check (Anthropic 400s on an inverted
+  pair) was gated on the provider *kind*, so it also rejected slots whose model sends no
+  budget at all — a Gemini slot (takes a `thinkingLevel`) or an Anthropic *adaptive* slot
+  (takes an `output_config.effort`) with, say, `max_tokens = 4096` and the default
+  `thinking_budget = 8192` was refused even though that budget never reaches the wire. The
+  check now gates on whether the resolved model shape actually *sends* a budget, so an
+  inert `thinking_budget` no longer blocks a valid config.
 
 - **A truncated batch answer no longer masquerades as a finished one.** When a batch
   item hit its output-token budget mid-response — most often a big attached-file review
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -434,6 +434,18 @@ the git log. Each later release appends a new section at the top.
 
 ### Fixed
 
+- **A Gemini slot's reasoning-effort setting now actually reaches Gemini.** kaibo's
+  Gemini casts (the default `gemini` synth, the `gemini-batch` Pro synth, the Flash-Lite
+  explorer) all name 3.x-line models, but the thinking-knob classifier recognized only
+  the bare `gemini-3-…` form — so the `-latest` aliases and the dotted 3.x ids (e.g.
+  `gemini-3.1-pro-preview`) fell through to the retired 2.5-era `thinkingBudget` path,
+  which pins depth to a fixed number and silently drops the per-role `effort` lever.
+  Setting `effort = "low"` (or `"high"`, `"max"`, …) on a Gemini slot had no effect. kaibo
+  now routes every Gemini id to `thinkingLevel` — the knob Google's current API documents
+  across the whole 3-line — so the effort you configure is the reasoning depth Gemini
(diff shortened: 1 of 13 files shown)


tobert:
## Follow-up: the fix is now verifiable in a trace (+ live Gemini validation)

Added a small observability commit (`6ab0d41`) so this fix can be *seen on the wire*, not just inferred — and used it to dogfood a live end-to-end validation of the Gemini family's thinking mode.

### What the commit adds
- **`run_phase` span now records `gen_ai.request.thinking`** (`src/consult/engine.rs`) — the exact `additional_params` blob a phase ships (thinkingLevel/budget + effort + sampling). Inert with no exporter; a toggle-less (`None`) provider records nothing.
- **Span-capture flake-fix centralized** — `CAPTURE_SERIAL` / `force_multi_dispatcher` / `serialized_capture` moved from `tool_span.rs` into the shared `test_support` module so **one** serial guard covers the whole test binary (a `run_phase` capture and a `tool` capture must serialize against each other, and it's a *third* no-subscriber test that poisons a callsite).
- **New discriminating test** `run_phase_span_carries_the_thinking_params` — asserts the field on both edges (exact blob with params, nothing without). Teeth-proven: neutered the `record` call and watched the positive branch go red (`[None, None]`), then restored.

### Live wire proof — the whole ladder rides `thinkingLevel`
Ran flash-lite → 3.5-flash → Pro against this branch and pulled the `run_phase` spans from OTLP:

| rung | model | slot | `thinkingLevel` | on `main` would've sent |
|---|---|---|---|---|
| `explore` | `gemini-flash-lite-latest` | explorer | **high** | `thinkingBudget: 8192` |
| `consult` | `gemini-3.5-flash` | synth | **high** | `thinkingBudget: 8192` |
| Pro (interactive) | `gemini-pro-latest` | synth | **high** | `thinkingBudget: 8192` |

All three miss the old `gemini-3`/`gemini-3-*` prefix (a `-latest` alias or a `.5`/`.` breaks it), so every one was misrouted to `thinkingBudget` before this fix. Each request carried the correct per-role sampling nested in `generationConfig` (explorer `temperature: 0.1`, synth `0.3`), and none 400'd on the level — so **the effort lever now reaches Gemini across the family**.

Answering the question that kicked this off — *were Pro batches thinking-off before?* No: `gemini-pro-latest` fell to `thinkingBudget: 8192`, so thinking ran, but at a **fixed budget with the per-role effort lever silently dropped** (and since the batch lane *forces* effort, a Pro batch ignored that forcing and ran flat). After the fix it's `thinkingLevel` = the forced `BATCH_EFFORT`.

### Cross-family review of the fold-in
Reviewed holistically by **`gemini-pro-latest`** (a different family than authored it), reading the actual worktree code:
> No effort settings are silently dropped, no valid configs are wrongly refused, and the observability changes are safe and robustly tested. The branch is good to go.

It independently verified: the span field is inert when telemetry is off or `thinking` is `None`; sharing one `CAPTURE_SERIAL` strictly serializes the capture tests (improving the flake guarantee); and the new test has real teeth on both the positive and negative cases.

### Not covered / follow-up
This span lives in `run_phase`, so it covers the **interactive** lanes. The **batch** lane (`batch.rs::gemini_batch_body`) assembles its own request and emits no `run_phase` span — its `thinkingLevel`-at-`BATCH_EFFORT` shape is proven by offline tests (`gemini_3_line_uses_thinking_level_at_batch_effort`, `gemini_shaping_nests_thinking_in_generation_config`), not a live span. A span in the batch request builder is a reasonable follow-up (recorded in `docs/devlog.md`; issues.md stays open-work-only).

Full offline suite green (379 lib tests incl. the refactored `tool_span` captures), `cargo clippy --all-targets` clean.

Issue #82, opened 2026-07-18
CLI: wire up consult_submit/job_wait and deliberate's direct lane


tobert:
The CLI front door (`kaibo consult`/`oneshot`/`explore`/`kaish`/`batch`/`config`, `src/cli.rs`) covers every *synchronous* MCP tool, but two async paths are still MCP-only:

- `consult_submit` + `job_wait`/`job_get`/`job_list`/`job_cancel` (background consults)
- `deliberate`'s **direct** lane (one long completion on a big local model, session-scoped `job-N`)

Both return a handle backed by an in-memory `JobStore` (`src/jobs.rs`) tied to a `tokio::spawn`ed task on the long-running MCP server process (stdio lifetime = caller session). A one-shot CLI invocation has no such long-running process to hold the job open — it would submit, exit, and take the job with it before a second invocation could ever collect it.

This is the same shape as the "restart survival is intentionally out of scope" note already in `docs/issues.md` (Async consult follow-ups): making these durable across process boundaries is a kaibo-as-daemon question that hasn't been decided.

`batch_submit` sidesteps this today because its handle (`backend/provider-id`) is durable *at the provider*, not in kaibo's memory — `kaibo batch submit`, `get`, and `list` all work as independent one-shot CLI invocations. That's why batch got a CLI story and consult_submit/deliberate-direct didn't.

Candidate directions, not yet decided:
- Give `deliberate`'s direct lane a durable dossier/session-scoped path via the persistence store (`src/store.rs`), similar to how batch handles persist — this one seems tractable without a daemon.
- `consult_submit`'s value is mostly "keep working in the same long MCP session while a deep consult runs in the background" — that doesn't map cleanly onto independent CLI processes at all unless kaibo grows a daemon + control channel, which is a bigger decision.

Noted while documenting the CLI in the README (2026-07-18); flagged as a candidate for the next RC.

PR #83, opened 2026-07-18
docs(readme): document the CLI front door

tobert:
## What & why

`kaibo consult`/`oneshot`/`explore`/`kaish`/`batch`/`config` shipped in #77/#78
but the README never got a CLI section — a reader had no way to discover the
CLI existed at all, only the changelog and `src/cli.rs`'s own doc-comment.

## Changes

- New **`## CLI`** section: MCP-tool → CLI-subcommand table, the
  stdout-is-the-answer/stderr-is-everything-else + `--json` + exit-code
  contract, and where `--session`/batch-handle state actually lives on disk
  (default XDG path, what's stored, `--state-db`/`--no-persistence` escape
  hatches).
- Introduction now mentions the CLI as a second front door.
- The container-image subsection gets a `docker run … kaibo consult …`
  example (drop `-i`, append a subcommand — no stdio transport to hold open
  for a one-shot run).
- Filed #82 for the one gap: `consult_submit`/`job_wait` and `deliberate`'s
  direct lane aren't CLI-reachable — their job handles are in-memory on a
  long-running server process, so a one-shot CLI invocation would exit and
  take the job with it before collection. README links it instead of
  re-explaining the architecture inline.
- CHANGELOG entry under `### Changed`.

Reviewed with `kaibo consult` (cast `deepseek`) against `src/cli.rs`,
`src/main.rs`, `Dockerfile`, and `docs/config.md`'s persistence section —
every factual claim (subcommand names, flags, exit codes, JSON envelope
shape, the `docs/config.md#persistence-persistence` anchor, the docker
entrypoint behavior) checked out. One wording nit it flagged ("every
synchronous tool" being imprecise since `batch` isn't synchronous) is fixed
in this diff.

## Test plan

- [x] `cargo build --bin kaibo`
- [x] Live-ran `kaibo --help`, `kaibo config`, `kaibo oneshot … --cast openrouter`
      (real answer + provenance footer) to confirm the documented behavior
      matches reality
- [x] Verified the `docs/config.md#persistence-persistence` anchor slug
- [x] Cross-model review via `kaibo consult` (cast `deepseek`)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -105,7 +105,7 @@ project and cannot run external commands.
 ## Working here
 
 - **TDD.** Tests that can and will fail. The sandbox boundary gets failing-first
-  tests — and we prove they have teeth (mount the project writable with
+  tests — and we prove they actually work (e.g. mount the project writable with
   `LocalFs::new` instead of `read_only`, watch the write-denial tests fail).
 - **`docs/sandbox-probes.md` is the read-only/containment audit runbook.** The
   `cargo test` suites are the continuous guard; that doc is how we *live-test* the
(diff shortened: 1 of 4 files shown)


tobert:
Follow-up commit (`36b3502`) from review: `kaibo --help` (long form) now carries the `EXIT CODES` table directly — previously it only lived in `cli.rs`'s doc-comment and this README section, so a script author running `--help` cold had no way to find it. `-h` stays terse (clap `after_long_help`); a new test (`long_help_documents_exit_codes`) asserts the split holds.

Also dropped the "have teeth" idiom (README's exit-code paragraph + two spots in AGENTS.md) for plain "have defined behavior" / "actually work" — flagged as hard to parse for a reader seeing it cold.

Full `cargo test` (598 passed) + clippy clean on top of this commit.

PR #93, opened 2026-07-26
OpenAI joins the batch lane — file-based, Responses-only

tobert:
## What this is

The third provider on kaibo's offline batch lane. A hosted OpenAI Platform backend can now
staff a `lane = "batch"` synth, so `batch_submit` / `job_get` / `job_cancel` / `job_list`
and `deliberate` all reach GPT — same kaibo semantics as the Anthropic and Gemini lanes:
toolless items, max thinking, a durable `backend/provider-id` handle.

## The decisions behind the shape

**Batch capability stopped being a property of a provider *kind*.** `batch_supported(kind)`
was fine while the answer was "anthropic and gemini" — keyed kinds, one endpoint, one
capability. `openai` isn't: the kind is *anything speaking the OpenAI wire* (Platform,
llama.cpp, Ollama, LM Studio, gateways), and `/v1/batches` belongs to Platform alone. A
kind-keyed check would have handed a local Gemma server a batch lane it cannot serve, and
the user would have found out as a 404 at submit — after config load told them the cast was
fine. So it now takes `&Backend` and reuses the exact-base-URL seam (`is_hosted_openai`)
that #92 built for interactive Responses routing. A `lane = "batch"` synth on a local
endpoint is a **load error naming the fix**.

**Responses-only.** Every item is a `POST /v1/responses` — the same endpoint the interactive
hosted arm uses, so batch and interactive share one body shape, one reasoning knob
(`reasoning.effort` at `BATCH_EFFORT`), one answer parser. The Chat Completions fallback
`docs/openai-api-plan.md` left open was rejected: a second body builder *and* a second
answer parser for no model kaibo can reach today. Older `gpt-4*` families run on Responses
too — they take sampling instead of reasoning, which the model-aware classifier already
knows.

**There is no local file.** Amy's question reframed the retention issue better than the plan
had: *why is it a file anyway — could we do it in memory?* It is in memory. The file is
never ours. Anthropic carries its requests inline in one POST and Gemini nests them inline;
OpenAI is the odd one out, requiring them as a file resource **in the caller's own OpenAI
account**. kaibo builds the JSONL in a `String` and streams it into the multipart upload;
results come back as strings. The read-only invariant is untouched by this lane and no new
`std::fs` call rides in — the blessed-line count in `tests/no_write_path.rs` doesn't move.

**kaibo deletes nothing, deliberately.** That input file is the *only* surviving record of
what was submitted (the store keeps the handle and label, never the prompts); deleting an
output on poll would break re-polling a finished handle; and OpenAI expires batch outputs
itself after 30 days. The docs say plainly that files accrete in your account. An opt-in
cleanup flag is tracked in `docs/issues.md` — easy to add later, impossible to undo.

## Two things the other providers didn't need

- OpenAI splits succeeded and failed items across **separate files**, so a poll reads both
  and merges — a failure that lives only in the error file must not vanish from the answer
  set — then re-sorts by kaibo's own index `custom_id`s, since merge order is otherwise an
  artifact of which file an item landed in.
- OpenAI dates batches in **epoch seconds** where the others send RFC3339, so the list view
  converts. The test pins the round trip through `rfc3339_to_epoch` that the 24-hour
  recency filter depends on.

## Validation

- `cargo test` — 700 pass, 0 fail. `cargo clippy --all-targets` clean.
- **Live probe** (`gpt-5.6-sol`, two items): submitted → `validating` → `in_progress`, with
  `request_counts` parsing into the progress line correctly. `kaibo batch list` swept the
  new `gpt` backend beside the Anthropic and Gemini batches in one view, with the epoch
  timestamp rendered into the shared RFC3339 field. *The completed-results leg was still in
  flight when this PR opened; I'll post the finished output here.*
- New offline tests pin: the JSONL wire shape, attachments as native Responses parts, the
  effort/token flooring, the whole status lifecycle (including an unknown status being
  loud and an empty-string `error_file_id` reading as absent), `output_text` extraction
  with reasoning items skipped, the incomplete/empty/HTTP-error/request-error per-item
  failure paths, index re-ordering, validation-failure reasons, and the list timestamp
  round trip. Plus a config test proving the same kind + same model loads on the hosted
  backend and is refused on a local one.

## Review

Cross-family, both on this branch's code, no diff supplied (holistic reads):

- **DeepSeek** (`consult`, interactive over the worktree): *"No correctness gaps found. The
  gating is exhaustive, the per-item failure surface is complete, the tests are real, and
  the in-memory JSONL stays cleanly off the filesystem… Ship it."* Verified independently
  that `src/batch.rs` contains zero filesystem-mutating calls and the blessed-site count
  stays pinned at 4.
- **Gemini Pro** (`batch` deliberate, whole files attached): *"highly robust, correct, and
  defensively engineered"* — walked the four failure modes (progress stalls, silent
  per-item drops, batch-level validation failure, misreported truncation) and found each
  disarmed; called the per-backend routing airtight across `batch_supported`,
  `submitter`/`poller`, and `require_hosted`. Its one stated limit: it could not verify the
  `/v1/responses` body schema keys from documentation alone — which is what the live probe
  covers.

Reviewed-by: DeepSeek v4-pro via kaibo `deepseek` cast
Reviewed-by: Gemini Pro via kaibo `gemini-deliberate` batch cast

https://claude.ai/code/session_019cLSYtttNJjNeAeYkemdaZ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,20 @@ the git log. Each later release appends a new section at the top.
   A `kind = "openai"` backend pointed exactly at `https://api.openai.com/v1` can run
   current GPT-5.6 models through `oneshot` and `consult`, including image attachment and
   `view_image`; generic/local OpenAI-compatible backends stay on the Chat Completions path.
+- **OpenAI joins the offline batch lane.** A hosted OpenAI Platform backend can now staff a
+  `lane = "batch"` synth, so `batch_submit`/`job_get`/`job_cancel`/`job_list` and
+  `deliberate` all reach GPT — same kaibo semantics as the Anthropic and Gemini lanes
+  (toolless, max thinking, half price, a durable `backend/provider-id` handle). OpenAI's
+  batch protocol is the file-based one: kaibo builds the requests as JSONL **in memory** and
+  uploads them as a file in *your* OpenAI account, then downloads the results — it writes
+  nothing locally, and deletes nothing (that input file is the only record of what you
+  submitted, and OpenAI expires batch outputs itself after 30 days). Batch capability is now
+  judged per **backend** rather than per provider kind, because only OpenAI Platform serves
+  `/v1/batches` — a `lane = "batch"` synth on a local OpenAI-compatible server is refused at
+  config load with the fix named, instead of 404ing at submit time.
 - `docs/openai-api-plan.md`, a living plan for making hosted OpenAI a first-class kaibo
   backend and clarifying that kaibo's OpenAI model calls use OpenAI Platform API keys, not
-  Codex subscription entitlement. It also tracks OpenAI's `/v1/batch` support and the
-  kaibo provider adapter still needed before Sol can run on kaibo's offline batch lane.
+  Codex subscription entitlement.
 - **`consult`** — the headline tool: ask a model *outside your own family* about a
   codebase and get a grounded, cited answer. A capable model reads precise spans
   directly and delegates broad sweeps to a cheap explorer sub-agent, then synthesizes
--- a/README.md
+++ b/README.md
@@ -378,6 +378,9 @@ background and hands back a job handle immediately, so your agent keeps working.
 provider's batch lane — offline, maximum thinking, roughly half the interactive
 price. The built-in `anthropic-batch` and `gemini-batch` casts put Claude Opus or
 Gemini Pro on your hardest questions cheaply; handles are durable across restarts.
+A hosted OpenAI Platform backend can staff the lane too — declare `lane = "batch"`
+on its synth slot (`docs/config.example.toml`), and GPT joins the same handles,
+polling, and `deliberate` flow.
 
 ### `deliberate` — deep offline reasoning over a dossier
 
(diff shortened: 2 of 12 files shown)


tobert:
## Live probe: complete ✅

The results leg landed — the part Gemini's review said it couldn't verify from documentation
(the `/v1/responses` body schema on the way out) is now confirmed on the wire.

```
$ kaibo batch get gpt/batch_6a660ee3336881908117f9194432d6d5
Batch complete — 2 result(s):

## [0]
kaibo-openai-batch-item-A

## [1]
kaibo-openai-batch-item-B

———
kaibo · batch · gpt · batch_6a660ee3336881908117f9194432d6d5
```

Full lifecycle observed on a real two-item batch against `gpt-5.6-sol`: `validating` (0/0) →
`in_progress` (0/2 → 1/2) → `completed` (2/2). What that exercises end to end:

- the multipart upload of the in-memory JSONL, and OpenAI accepting the request bodies —
  so `instructions` / `input` / `max_output_tokens` / `reasoning.effort` are the right keys;
- `request_counts` parsing into the progress line at each stage;
- the output-file download and `output_text` extraction, with reasoning items skipped;
- `order_by_index` — both answers came back under their submit indices, in order;
- the empty-string `error_file_id` on a clean batch reading as *absent* rather than as a
  file to fetch (a `/files//content` 404 would have sunk the whole poll).

**Re-polling a finished handle is idempotent**, which is the property the delete-nothing
decision rests on — the second `batch get` above returned the identical result set, because
nothing was cleaned up behind it.

`batch list` on the new backend:

```
- `gpt/batch_6a660ee3336881908117f9194432d6d5` — completed, 2/2 done (created 2026-07-26T13:42:59+00:00)
```

— epoch seconds converted into the shared RFC3339 field, which is what the 24-hour recency
filter reads back.

PR #96, opened 2026-07-26
The explorer learns to attach: bytes route past the model that found them

tobert:
# The explorer learns to attach: bytes route past the model that found them

## What

The delegated explorer sub-agent inside `consult` and `deliberate` gains an `attach`
tool (new module `src/sweep_attach.rs`). It is a **routing verb, not a reading verb**:
the explorer names workspace paths; kaibo resolves them through the existing attach
machinery (canonicalize → containment under the project root → TOCTOU-capped read via
the read-only kaish worker → `classify` → `cat -n` numbering → escaped `<file>`
wrapper) and the full bytes ride *alongside* the report to the sweep's consumer. The
explorer only ever receives a receipt (path, line count, size, budget remaining) —
the bytes never enter its window. A 3,000-line file costs the explorer ~20 tokens to
deliver.

## Why

The 2026-07-03 reframe ("attach means the model sees the bytes") covered only the
operator→down direction. The explorer's one channel *up* was its report string:
evidence had to be transcribed out of the cheap model's own `max_tokens`, lossily.
This was worst exactly where the architecture promises the most — `deliberate`'s
offline synth is toolless, so the frontier model was reasoning over a cheap
explorer's paraphrase of the code. Now the dossier carries the real numbered file.

## Decisions (from the design conversation with Amy)

- **Consumer-shaped delivery.** consult: stitched onto the explore tool result.
  deliberate: stitched into the dossier. Top-level `explore` sits out v1 (its caller
  reads files itself); one `None` suppresses both the tool and its preamble
  paragraph, so toolset and prompt can't drift.
- **Dedupe asymmetry is deliberate.** consult seeds the sink with caller-attached
  paths (their bytes already reach the driver); deliberate seeds it EMPTY — there, a
  caller attachment is only a read-WHOLE directive to the explorer, so the explorer
  re-routing it is the delivery, not a duplicate.
- **Demotions are consumer-shaped and loud.** consult's over-cap tail says "read it
  yourself: `cat -n …`"; deliberate's says **NOT INCLUDED — you cannot fetch it**,
  because a silent gap in front of a toolless synth is the failure mode.
- **Images are in scope** (Amy's call): a blind explorer can staple a PNG to a
  vision-capable consumer. Gated on the *receiving arm's* vision cap, refused loudly
  in the receipt. Mechanically: `RunExplore::Output` became `serde_json::Value` —
  bare `Value::String` when textual (wire-identical to the old `String`), rig's
  hybrid `{"response","parts"}` envelope when images ride (verified against
  rig-core 0.38.2 source). The view_image-specific break→rewrite→resume path for
  transports without tool-result images is generalized to key on *any*
  image-carrying tool result; the next image-bearing tool needs no edit.
- **`max_attachments` is a count, set high on purpose** (default 32/sweep, `0` =
  tool not injected; file/env/`--max-attachments`, surfaced in `kaibo://config`).
  Amy wants to observe problematic patterns in the wild before clamping — every
  accepted attach emits a `PhaseEvent::Attached` progress beat, and `classify`'s
  existing per-file/cumulative byte caps bound the worst case. No per-call MCP arg:
  resident schema text under the 2048-char budget is the wrong price for an
  operator guardrail.

## Invariants

Pure read path — no new `std::fs` write sites (`tests/no_write_path.rs` still pins
exactly 4 blessed lines); containment canonicalizes into the project root with the
same discipline as `view_image`; the kaish kernel never crosses an `.await` (only
the `Send` worker handle does; `tests/explore_send.rs` still compiles). The sink's
reserve/commit/release protocol makes the count cap atomic under rig's concurrent
tool calls.

## Tests

~37 new (module 16, prompts 4, scripted-loop 14, containment 2, config 1); 735
total green, clippy clean, `cargo tree -i aws-lc-rs` / `-i mimalloc` empty.
Mutation teeth-check on the two load-bearing properties: leaking file bytes into
the receipt → red; disabling the containment check → 4 tests red across two
binaries.

## Review

Process: Opus architect plan (every seam pre-anchored `file:line`), Sonnet
implementation, Fable orchestration. Cross-family review per house style — whole
files, no diff:

- **DeepSeek** (interactive consult over the branch): verified the sink protocol,
  `Value` envelope wire, break/rewrite generalization, containment, and both
  deliberate lanes clean with `file:line` anchors. Two MEDIUM test gaps, both
  addressed: the two-images-in-one-turn rewrite test is added; the server-level
  deliberate-image integration test is recorded in `docs/issues.md`.
- **Gemini Pro** (offline `deliberate`, whole-file dossier): raised five risks.
  Three HIGHs were verified against the code and refuted — the break already
  parses and confirms `parts[]` (now pinned by a test so it stays that way),
  refusal wording mirrors `view_image`'s established shape, and kaibo never
  truncates the dossier so its orphaned-image scenario has no trigger. Its
  slot-leak-on-cancel concern is unreachable (the sink is per-sweep and dies
  with the sweep).
- **Convergent and adopted**: both reviewers flagged that the explorer preamble
  must say attach returns a receipt, never the contents — a weak explorer would
  otherwise attach what it meant to read and hallucinate. The directive now says
  "delivering, not reading," pinned in its test.

Post-review: 738 tests green, clippy clean, `no_write_path` blessed-line count
unchanged at 4.

https://claude.ai/code/session_012KFFmSMYYCFcBDzGx3J8be

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,20 @@ record. Each later release appends a new section at the top.
 
 ## [Unreleased]
 
+### Added
+
+- **The explorer can now hand whole files to whoever reads its report.** Inside
+  `consult` and `deliberate`, the delegated investigator gets an `attach` tool: when a
+  whole file is the evidence, it routes the file's real bytes (numbered, `cat -n`
+  style) alongside its report — into the consult driver's context, or into the
+  `deliberate` dossier the offline model reasons over — instead of transcribing spans
+  through its own small budget. Images ride too: an explorer that can't see a PNG can
+  still staple it to a vision-capable answering model, and is told plainly when the
+  reader is text-only. Governed by `[defaults] max_attachments` (files per sweep;
+  default 32, `0` disables the tool), also settable via `KAIBO_MAX_ATTACHMENTS` and
+  `--max-attachments`; every routed file surfaces as a progress beat so you can watch
+  what the explorer chose.
+
 ### Fixed
 
 - **State-db-collides-with-project-tree error now names `--root`/`--allow-path`.**
--- a/docs/config.md
+++ b/docs/config.md
@@ -270,6 +270,14 @@ The `[defaults]` knobs themselves:
   bounds resident prompt cost, not just one request. The tool-less tools
   (`oneshot`/`batch_submit`) are unaffected — with no shell to fall back on, they
   keep their own hard per-file/per-call caps.
+- **`max_attachments`** (32; `0` is legal): cap on how many files ONE explorer sweep
+  may route with its `attach` tool — the bytes ride ALONGSIDE the sweep's report to
+  whoever reads it (the `consult` driver, or `deliberate`'s offline synth) without
+  ever entering the explorer's own context. Distinct from `inline_attach_budget`
+  (which bounds inlining the *caller's* attachments into the driver prompt): this
+  bounds a sweep's own routing, and it's a behavioral guard, not a memory one — the
+  per-file/cumulative byte caps (`attach.rs`) already bound worst case. `0` turns the
+  tool off entirely.
 
 ### Built-in registry (the defaults)
 
@@ -414,6 +422,7 @@ role the cast doesn't carry). The naming rule for everything else is mechanical:
 | session cache size | `defaults.session_capacity` *(must be > 0)* | `KAIBO_SESSION_CAPACITY` | — |
 | async job cache size | `defaults.job_capacity` *(must be > 0; default 64)* | `KAIBO_JOB_CAPACITY` | — |
 | attach inline budget (bytes) | `defaults.inline_attach_budget` *(0 = never inline; default 262144)* | `KAIBO_INLINE_ATTACH_BUDGET` | — |
+| explorer attach cap (count) | `defaults.max_attachments` *(0 = attach tool off; default 32)* | `KAIBO_MAX_ATTACHMENTS` | `--max-attachments N` |
 | exec timeout (s) | `sandbox.exec_timeout_secs` | `KAIBO_EXEC_TIMEOUT_SECS` | — |
 | output cap (bytes) | `sandbox.output_limit_bytes` | `KAIBO_OUTPUT_LIMIT_BYTES` | — |
 | scratch cap (bytes) | `sandbox.scratch_limit_bytes` *(must be > 0; default 64 MB)* | `KAIBO_SCRATCH_LIMIT_BYTES` | — |
--- a/docs/devlog.md
+++ b/docs/devlog.md
@@ -34,6 +34,69 @@ asks before opening XDG paths, names Codex's `network_access` / `writable_roots`
 offers a Codex-only state db or CAS dir as a cleaner multi-agent default when sharing is not
 the goal.
 
+## 2026-07-26 — the explorer learns to attach: bytes route past the model that found them
+
+Amy asked "what if we gave explorers an attach tool that would work like the other
+attachments kaibo does" — and the 2026-07-03 reframe ("attach means the model sees the
+bytes") turned out to be only half-finished. That work made the caller's attachments
+reach every reader as real bytes, operator→down. But the explorer's only channel *up*
+was its report string: any evidence it wanted its consumer to hold, it had to
+transcribe — out of its own `max_tokens`, through its own window, lossily. The cheap
+explorer was the narrowest pipe in the system and we were pushing whole files through
+it by hand.
+
+The fix is a **routing verb, not a reading verb**: `attach` joins `run_kaish` in the
+sweep toolset. The explorer names paths; kaibo resolves them through the same
+containment + TOCTOU-capped VFS read + `classify` + `number_lines` machinery the
+caller's attach uses, and the bytes ride *alongside* the report to whoever reads it.
+The explorer gets back a receipt — path, line count, size, budget remaining — and
+never the contents. A 3,000-line file costs the explorer ~20 tokens to deliver. The
+receipt-never-carries-bytes property is the module's load-bearing test, and it was
+mutation-checked (leak the body into the receipt → red; disable the `starts_with`
+containment guard → four tests red across two binaries).
+
+Where the bytes land is consumer-shaped, and that shape carried the design:
+
+- **`deliberate` is the case the feature exists for.** Its offline synth is toolless —
+  everything it reasons over must already be in the dossier, and until now "in the
+  dossier" meant "transcribed by the explorer." Now the dossier carries the real
+  numbered file. Consequently deliberate's sink does **no** caller-file dedupe: a
+  caller attachment there is only a read-WHOLE directive to the explorer, so the
+  explorer re-routing it is the delivery, not a duplicate. `consult`'s sink dedupes
+  the same paths, because its driver already holds them inlined. One concept — "seed
+  the sink with paths whose bytes already reach the consumer by another channel" —
+  two seedings. Over-cap demotion splits the same way: consult's says "read it
+  yourself: `cat -n …`", deliberate's says **NOT INCLUDED — you cannot fetch it**,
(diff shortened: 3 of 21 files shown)


tobert:
## Cross-family review — DeepSeek (`deepseek-v4-pro` synth / `deepseek-v4-flash` explorer, via kaibo `consult`)

Reviewed holistically against the rebased worktree, no diff supplied. **No findings requiring action** — the boundary, the budget, and the concurrency model all hold up. Recording the substance because it's the first review this PR has had.

### Containment — no widening, and stricter in one dimension

`sweep_attach.rs:332-350` canonicalizes then checks `canon.starts_with(&self.root)`, where `root` already passed `Resolver::resolve_root`. That makes `attach` **stricter** than `resolve_attachments`: the explorer can only attach under the one project root it's sweeping, not every allowed tree. The read at `sweep_attach.rs:386` goes through `worker.read_file_capped` — the same read-only VFS `run_kaish` uses — so a symlink swapped in after host-side canonicalization is refused at the mount layer structurally, not by racing a re-check. The sink holds only in-memory state; nothing touches disk.

Noted and endorsed: worktree-following doesn't apply to `attach` paths, so an explorer sweeping a worktree can't `attach` from a sibling worktree (it can still `cat -n` it). The reviewer judged this the right call — `attach` routes full bytes, and widening it to sibling worktrees is a separate design decision, not something to inherit by reaching for the broadest check. Agreed.

### Budget — accurate, atomic, no leak

Per-**sweep**, not per-consult (a fresh `SweepAttachSink` per `RunExplore::call`, `engine.rs:1160-1165`). The receipt's "budget remaining" matches enforcement because `usage()` counts only `Slot::Ready`; `Released` slots free their line while `seen` still prevents re-claiming the same bad path. `reserve` counts `Pending`+`Ready` under the mutex, so concurrent calls can't oversubscribe — pinned by `concurrent_attach_calls_share_one_budget`. `read_file_capped`'s `+1` overflow byte catches a file that grew between stat and read, so there's no OOM window.

### Concurrency

One `Mutex<State>`, never held across an `.await` — the async work (read, classify) happens *between* `reserve` and `commit`/`release`. The `Pending → Ready/Released` state machine is what makes this correct under rig's `buffer_unordered`. `drain` is idempotent and in practice called once, after the sweep returns, so it can't race a `reserve`.

### Failure modes — all loud

The reviewer traced all ten paths in `attach_one` (`sweep_attach.rs:325-433`): missing path, outside root, vanished, not a regular file, duplicate, already-delivered, over budget, read failure, classify refusal, image-to-blind-reader. Every non-trivial failure releases its slot so the budget isn't spent on a bad file, and every one produces a receipt line. Nothing is swallowed.

### Complexity

`RunExplore::Output` as `serde_json::Value` was judged **justified** rather than over-built: the two wire shapes are exactly what rig's `from_tool_output` parses, `view_image` already uses the same hybrid envelope, and the no-image path is byte-for-byte identical to the pre-attach `String` output. `SweepConsumer` is right-sized at two variants driving two behaviors. No dead code found.

The one seam noted: the tool description (`sweep_attach.rs:468-481`) and the preamble directive (`prompts.rs:520-535`) carry similar framing, so changing the `attach` narrative means touching both. Defensible — different audiences, different token budgets — and the code already points at the seam.

---

A `gemini-deliberate` pass is also running for a second family; I'll add it when it lands. (A `fable` run was attempted first and failed on Anthropic credit balance, not on the code.)

tobert:
## Cross-family review 2 — Gemini Pro (`gemini-pro-latest`, batch lane, via kaibo `deliberate`)

A second family over the same surface, and it earned its keep: **it found a real bug DeepSeek missed.**

### Finding 1 (accepted, fixed in 7ca2a5f): the consult dedupe seed wasn't canonicalized

`consult` seeded `already_delivered` with a plain `root.join(a.path())`, while the sink compares against a **canonicalized** path (`reserve` is handed `canon`, `sweep_attach.rs:170`). So any caller attachment not already in canonical form — a `./` segment, a `..`, a symlinked file or ancestor — failed to match. The dedupe missed and the explorer re-routed bytes the driver already had.

Verified against the code before acting. It fails quietly and costs tokens rather than correctness, which is exactly the kind of defect that survives a long time unnoticed.

Fixed by making the seed a named function, `delivered_seed`, that resolves paths the same way `attach_one` does — a silent-mismatch hazard between two resolution sites deserves one home and a test, not a `.map()` at the call site. The existing dedupe test couldn't catch this because it seeds an already-canonical path by hand; the bug lived in the production seed. New tests drive the helper with `./README.md`, `sub/../README.md`, and a symlink, and each was confirmed to fail against a join-only seed by reverting the canonicalize and watching it go red.

### Finding 2 (reviewed, deliberately not changed): transient budget-receipt drift

A receipt's "budget remaining" is a point-in-time snapshot, so a concurrent sibling's RAII refund can make one receipt report less budget than is actually free.

Left alone on purpose: it's inherent to reporting a number from inside a concurrent turn, it errs *conservative* (the model under-uses rather than over-spends), and enforcement itself stays atomic under the one mutex. Gemini's own recommendation was only "don't let the prompt penalize the model for it", which the current framing doesn't.

### Where the two reviews agreed

Containment secure (canonicalize + `starts_with` + the read-only worker, no new write path); budget enforcement atomic under `buffer_unordered`; the consult/deliberate asymmetries correctly implemented as described; every failure mode loud with the slot refunded; and no unnecessary complexity — both independently judged `Option<Arc<SweepAttachSink>>` and the `SweepConsumer` split the right shapes, and `RunExplore::Output` as `Value` justified rather than over-built.

Gemini added one framing I liked: the design "separates the control plane (the explorer's receipt) from the data plane (the direct-to-consumer payload)."

---

Both reviews are cross-family (neither wrote this code). With the seed fix in, I have no outstanding concerns on this PR.

tobert:
**Brought up to date with main (rig 0.41 era) + the overdue cross-family review.**

**Merge `ea7c305`** — the textual conflicts were the small part; the branch predated rig 0.41 (#113) and the direct single-shot lanes (#116), so its new code needed porting, not just merging:

- `SweepAttach` and the generalized `ToolImageBreakHook` moved to rig-agent 0.41's `Tool`/`AgentHook` traits. The hook now keys on **typed `ToolOutput` image parts** instead of sniffing serialized JSON for `"type":"image"` — the false positive the old test pinned (a grep hit quoting envelope JSON) is now impossible by construction, and the test proves that instead.
- `RunExplore` emits declared `ToolOutput` blocks (text + `Image`), the contract `view_image` moved to on main, because rig 0.41 only routes images a tool *declares*. This also dissolved the tracked "report arrives JSON-quoted to the driver" issue — that quoting came from the `Value` envelope — so its `docs/issues.md` entry is deleted.
- `deliberate_direct` keeps #116's direct-completion shape and gains this branch's sweep-routed images (same `user_turn_with_attachments` prompt, one upstream request).
- `max_attachments` documented in the rewritten `docs/config.md` `[defaults]` section and the shipped template, keeping its "every option with its default" promise true.

838 tests green, clippy silent, no new rustfmt drift.

**Cross-family review (DeepSeek, agentic consult over the whole worktree, no diff handed over)** — verdict **ready to merge**, all five axes clean:

1. *Budget/dedupe/containment under concurrency*: sound — one mutex, never held across the async read, slot claimed at `reserve` so `buffer_unordered` races can't oversubscribe (pinned by the concurrent-attach test).
2. *rig-0.41 port fidelity*: typed-content contract consistent across `view_image`, `RunExplore::call`, `tool_output_carries_image`, and `rewrite_tool_image_history`; no path for a silent image drop or double delivery.
3. *Deliberate lane*: sweep-routed image survives dossier drain → lane dispatch → the synth's single turn, both batch and direct, traced end to end.
4. *Explorer-facing prompt text*: positively framed, budget-transparent, names its reader.
5. *Read-only/containment invariants*: unweakened — `attach_one` mirrors the `view_image` canonicalize-then-`starts_with` discipline, symlink escape tested.

Its one non-blocking nit (tool description took the sink lock to render a constant cap) is folded in `2c5b965`.

PR #99, opened 2026-07-27
Add per-backend wire = "responses" | "chat" for openai-kind backends

tobert:
## Problem

kaibo only speaks OpenAI's Responses API when an `openai`-kind backend's `base_url`
is exactly `api.openai.com` (`Backend::is_hosted_openai`, `src/config.rs`). Every
other `openai`-kind backend gets rig's Chat Completions client, which sends
`max_tokens` — and current GPT-5.x reasoning models reject that outright
(`unsupported_parameter`; they want `max_completion_tokens`/Responses). So GPT-5.x
behind an OpenAI-compatible gateway/proxy that implements OpenAI's Responses API at
`/v1/responses` (verified against a real gateway) was unusable, even though nothing
about the wire protocol required that.

## Design decisions

- **New optional per-backend field**: `wire = "responses" | "chat"` on an
  `openai`-kind backend (`RawBackend`/`Backend` in `src/config.rs`). Validated in
  the existing loud style — a load error on any non-openai kind (naming the fix),
  and a load error naming both legal values on an unknown one.

- **Split the predicate in two, rather than growing `is_hosted_openai` a `wire`
  branch.** `is_hosted_openai` keeps its endpoint-exact meaning and stays the
  **batch/platform-lane gate** (`src/batch.rs`'s `batch_supported`/`require_hosted`)
  — a gateway that faithfully proxies `/v1/responses` does not necessarily proxy
  `/v1/batches` or the Files API, so `wire = "responses"` must never grant batch
  eligibility. The new `Backend::uses_responses_wire()` is the interactive-only
  question: an explicit `wire` wins in either direction (a gateway can opt in, or
  OpenAI Platform's own endpoint can opt out to Chat Completions); unset falls back
  to the same endpoint-exact heuristic `is_hosted_openai` uses, preserving today's
  behavior for anyone who hasn't set the knob. Both predicates share a private
  `is_openai_platform_endpoint` helper so the endpoint check itself isn't
  duplicated.

- **Interactive wire selection** (`Arm::from_slot` in `src/consult/engine.rs`) now
  reads `uses_responses_wire` instead of `is_hosted_openai`. `kaibo://config`'s
  rendered `wire` and its inert-tunables computation (`src/server/config_resource.rs`)
  follow the same swap, since both describe the interactive request shape, not
  batch eligibility. `src/batch.rs`'s `batch_supported`/`require_hosted` are
  untouched — still `is_hosted_openai`.

- **Docs**: `docs/config.example.toml` gets a `[backends.gateway]` example (a
  `wire = "responses"` gateway), `docs/config.md` documents the knob in prose and
  notes that none of the built-in backends set it. `CHANGELOG.md` gets one `Added`
  entry. `docs/issues.md` notes the deliberate follow-up: if a responses-wire
  gateway ever does faithfully proxy `/v1/batches`, it needs its own opt-in rather
  than riding in on `wire = "responses"`.

## Test coverage (written failing-first)

- **Config parse**: the knob is accepted on the `openai` kind and sticks as the
  typed enum; rejected loudly on `anthropic`/`deepseek` kinds naming the fix; an
  unknown value is rejected naming both legal spellings.
- **`uses_responses_wire` precedence**: unset + `api.openai.com` (with/without
  trailing slash) → true; unset + a local/other URL → false; `wire = "responses"`
  on a gateway URL → true; `wire = "chat"` on `api.openai.com` → false; a
  non-openai kind is always false. Mirrors the existing
  `hosted_openai_detection_is_endpoint_exact` test. Also proves `is_hosted_openai`
  does **not** follow the override in either direction.
- **Arm construction**: a gateway backend (`base_url =
  "https://llm-gateway.example.internal/v1"`, `wire = "responses"`) builds the
  Responses-shaped arm — mirrors `hosted_openai_arm_uses_responses_shape`.
- **Batch**: that same `wire = "responses"` gateway backend is still refused by
  `batch_supported`/`OpenaiBatch::from_backend`/`poller`.
- **Config render**: the *resolved* wire (not the raw config value) appears per
  openai-kind backend in `kaibo://config`, absent on every other kind.

## Quality gates

- `cargo test`: 465 passed in the lib suite, all integration test binaries green,
  0 failures anywhere.
- `cargo clippy --all-targets`: clean.
- `cargo tree -i aws-lc-rs` / `cargo tree -i mimalloc`: both empty (untouched, as
  expected — this change is config/consult/batch logic only).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,15 @@ the git log. Each later release appends a new section at the top.
 
 ### Added
 
+- **GPT-5.x is now usable behind OpenAI-compatible gateways via `wire = "responses"`.**
+  A new per-backend `[backends.<name>]` knob (`openai` kind only) picks the interactive
+  request shape explicitly: `"responses"` for rig's Responses client, `"chat"` for
+  OpenAI-compatible Chat Completions. It's optional — unset still infers the shape from
+  the endpoint, exactly as before — and exists for a gateway/proxy that implements the
+  Responses API at `/v1/responses` (verified against a real gateway) without sitting at
+  OpenAI Platform's own URL: without it, current GPT-5.x reasoning models reject Chat
+  Completions' `max_tokens` outright. `wire` never affects batch eligibility, which stays
+  Platform-only.
 - **Hosted OpenAI Platform backends now use the Responses API for interactive GPT calls.**
   A `kind = "openai"` backend pointed exactly at `https://api.openai.com/v1` can run
   current GPT-5.6 models through `oneshot` and `consult`, including image attachment and
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -289,6 +289,22 @@ base_url = "http://localhost:8080/v1"
 key_optional = true
 request_timeout_secs = 1800
 
+# An OpenAI-compatible gateway/proxy that implements OpenAI's Responses API at
+# /v1/responses (verified against a real gateway) — not OpenAI Platform's own
+# endpoint, so kaibo can't infer this from the URL. `wire = "responses"` opts it
+# into rig's Responses client for interactive calls, the shape current GPT-5.x
+# reasoning models require (they reject Chat Completions' `max_tokens` outright).
+# `wire = "chat"` is the symmetric opt-out, including on OpenAI Platform's own
+# endpoint. Either way, batch stays Platform-only: a gateway proxying /v1/responses
+# doesn't necessarily proxy /v1/batches or the Files API, so `wire` never affects
+# `batch_submit` eligibility — only `[backends.gpt]` above (Platform's exact URL)
+# can staff a batch synth.
+[backends.gateway]
+kind = "openai"
+base_url = "https://llm-gateway.example.internal/v1"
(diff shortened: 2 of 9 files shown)


tobert:
Cross-family review record (per the PR discipline): two independent `kaibo consult` passes over this branch, both from non-Claude lineages since the patch was Claude-written.

- **deepseek cast** (DeepSeek-V4-Pro synth): traced every `is_hosted_openai` / `uses_responses_wire` call site — all use the correct predicate, none missed. **No correctness bugs.** Two non-blocking notes: the `hosted_openai_*` shaping helpers in `src/consult/shaping.rs` are now reached by any responses-wire backend, so their "hosted OpenAI"-flavored names/doc comments have drifted (logic is model-id-keyed and correct — rename or doc touch-up at leisure); plus a minor test-coverage suggestion.
- **nemotron cast** (Nemotron-3-Super-120B synth): independently drove the shell over the same sites; confirmed the interactive/batch predicate split holds everywhere. **No issues found.**

Reviewer verdict matches my own read of the diff: the wire/batch split is the load-bearing decision and it's implemented and tested correctly. Ready to merge from my side.

PR #100, opened 2026-07-27
Allow a base_url on the gemini backend kind

tobert:
## Problem

kaibo's `gemini` kind rejected an explicit `base_url` at config load (`src/config.rs`: "only the `openai` and `anthropic` kinds have a configurable endpoint"), so a Gemini-API-compatible gateway/proxy in front of the real backend couldn't be dialed. rig-core 0.38's Gemini `ClientBuilder` is the same generic builder as Anthropic's and already exposes `.base_url(...)` — the restriction was entirely kaibo's own validation.

## Design decisions

1. **Relax validation.** `kind = "gemini"` now accepts `base_url`, the same way `kind = "anthropic"` already does. Error message becomes "only the `openai`, `anthropic`, and `gemini` kinds have a configurable endpoint".
2. **Plumb it into the live client.** `Arm::from_slot`'s `ProviderKind::Gemini` arm (`src/consult/engine.rs`) now conditionally calls `.base_url(...)` on rig's builder, mirroring the existing Anthropic arm.
3. **The host-root reconciliation (the one real decision).** I verified against the vendored rig-core 0.38.2 source (`providers/gemini/client.rs`) that rig's live Gemini client takes a **HOST ROOT** (`https://generativelanguage.googleapis.com`) and appends its own versioned path (`/v1beta/models/{model}:generateContent`) itself — it does not accept a pre-versioned `base_url`.

   kaibo's batch lane has its own fallback constant, `GEMINI_API_BASE = "https://generativelanguage.googleapis.com/v1beta"`, already carrying the version segment (the batch calls append relative paths straight onto it). Before this change, `GeminiBatch::base_url` used a configured `backend.base_url` **verbatim**, which would have been fine only because `base_url` could never actually be set on a gemini backend. Making `base_url` configurable meant that verbatim use would now dial a gateway with no `/v1beta` segment — a real bug this PR had to close, not just plumb around.

   The contract I settled on: **`base_url` is always a HOST ROOT**, matching rig's own builder and the anthropic kind's existing style, so one configured address serves both the interactive and batch paths. `GeminiBatch::base_url` versions a configured value with `/v1beta` itself — trimming a trailing slash and a pre-existing `/v1beta` suffix first (borrowing the spirit of rig's own `normalize_anthropic_base_url`, which strips a user-supplied `/v1`/`/messages` suffix), so a reasonable config never double-versions. The unset fallback (`GEMINI_API_BASE`) is untouched.

## Test coverage (written failing-first)

- `src/config.rs`: `gemini_backend_accepts_a_base_url` (mirrors `anthropic_backend_accepts_a_base_url`); flipped `backend_stanza_validation_is_loud`'s rejected example from gemini to deepseek since gemini is no longer in the rejected set, and updated the asserted message text.
- `tests/config.rs`: same flip in `base_url_on_a_keyed_backend_is_rejected`, plus a new `gemini_backend_accepts_a_base_url` integration test.
- `src/consult/engine.rs`: `arm_from_slot_builds_a_gemini_arm_with_a_custom_base_url`, mirroring the existing Anthropic-arm test — proves a keyless gemini arm with a custom `base_url` builds offline, and that an unset `base_url` still builds.
- `src/batch.rs`: `gemini_batch_base_url_versions_a_configured_host_root` — a configured host root gets `/v1beta` appended; a trailing slash is trimmed; a base_url pre-versioned with `/v1beta` (with or without a trailing slash) is not double-versioned; the unset fallback is unchanged.

All of the above were confirmed failing against the pre-change code before implementing the fix.

## Docs

`docs/config.md` (backend table + surrounding prose) and `docs/config.example.toml` (`[backends.gemini]` gains a commented `base_url` line mirroring the anthropic one) updated to describe the new, optional, host-root contract. `CHANGELOG.md` gets an Added entry under the unreleased section.

## Quality gates

- `cargo test`: all green (461 lib tests + full integration suite, 0 failed)
- `cargo clippy --all-targets`: clean
- `cargo tree -i aws-lc-rs` / `-i mimalloc`: both empty (untouched, as expected)
- Ran a cross-family review pass (kaibo's own `consult`, `anthropic` cast — the local `openai-local`/Gemma backend wasn't reachable in this environment) over the diff before opening this PR. It confirmed the host-root reconciliation is correct for all realistic inputs and found no other read-site in the codebase that assumes a gemini backend never has a `base_url`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,13 @@ the git log. Each later release appends a new section at the top.
   judged per **backend** rather than per provider kind, because only OpenAI Platform serves
   `/v1/batches` — a `lane = "batch"` synth on a local OpenAI-compatible server is refused at
   config load with the fix named, instead of 404ing at submit time.
+- **The gemini backend can now dial a Gemini-API-compatible gateway/proxy.** Setting
+  `base_url` on a `kind = "gemini"` backend used to be a load error; it now works the
+  same way it already does on the anthropic kind — unset still resolves to Google's
+  own endpoint, set it points both interactive (`consult`/`oneshot`) and batch calls
+  at your gateway instead. The value is a host root (matching rig's own client and
+  the anthropic kind's convention), not a versioned path — kaibo appends the
+  `/v1beta` version segment the batch lane needs on its own.
 - `docs/openai-api-plan.md`, a living plan for making hosted OpenAI a first-class kaibo
   backend and clarifying that kaibo's OpenAI model calls use OpenAI Platform API keys, not
   Codex subscription entitlement.
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -206,10 +206,11 @@ scratch_limit_bytes = 67108864
 # Keys are referenced, never inlined: api_key_env names an env var, api_key_file a
 # path (~ ok). env wins over file. key_optional=true falls back to a placeholder
 # bearer token (keyless local servers). base_url is legal on the openai kind
-# (required for a new backend) and the anthropic kind (optional — unset still
-# dials rig's built-in api.anthropic.com; set it to point at an Anthropic-
-# Messages-API-compatible gateway/proxy instead). Every other keyed kind rejects
-# base_url at load (rig fixes those endpoints).
+# (required for a new backend) and the anthropic and gemini kinds (optional —
+# unset still dials rig's built-in api.anthropic.com / generativelanguage.googleapis.com;
+# set it to point at an Anthropic/Gemini-API-compatible gateway/proxy instead —
+# a HOST ROOT in both cases, since rig appends its own versioned path). Every
+# other keyed kind rejects base_url at load (rig fixes those endpoints).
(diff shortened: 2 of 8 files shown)


tobert:
Cross-family review record: the pre-PR review in the body ran on an anthropic cast, which is the same lineage that wrote the patch — so here's the required outside pass, via `kaibo consult` on the **deepseek cast** (DeepSeek-V4-Pro synth) over this branch.

- **One finding, cosmetic:** a stale test doc comment still said "base_url is openai/anthropic-kind only" — fixed in 8e01bf0.
- `/v1beta` normalization edge cases (case-sensitivity, double-versioned input, mid-path segment) reviewed and judged out-of-contract for a host-root `base_url`; they fail loudly rather than silently, so left as-is.
- Checked every `ProviderKind::Gemini` site: none still assume a gemini backend has no `base_url`; the batch submitter/poller dispatch routes correctly.

No correctness bugs. Ready to merge from my side.

PR #106, opened 2026-07-29
chore: bump turso 0.7.0 -> 0.7.1

tobert:
## Summary

- turso is exact-pinned (`.db-tshm` format + API declared unstable between
  releases), so a bump is always a deliberate, tested port — never a silent
  float. Checked upstream's `v0.7.0...v0.7.1` compare before touching the pin:
  9 commits, and the only change touching the embedded engine (`core/translate`)
  is two query-planner bugfixes (IN-LIST cost model, a missing-index-after-UPSERT
  case). Everything else is JS/serverless bindings and other-language SDK
  packages kaibo's embedded, `default-features = false`, no-`sync` build never
  touches.
- Part of the prep for cutting real v0.2.0 (turso was flagged stale during the
  release checklist pass).

## Test plan

- [x] `cargo build --all-targets` clean
- [x] `cargo test` — 502 lib + full integration suite green, including
      `tests/store.rs`'s `survives_reopen` / `reopen_does_not_re_run_migration`
      (the tests that would catch a `.db-tshm` format drift)
- [x] `cargo clippy --all-targets` clean
- [x] `cargo tree -i aws-lc-rs` / `-i mimalloc` still empty

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3938,9 +3938,9 @@ checksum = "e421abadd41a4225275504ea4d6566923418b7f05506fbc9c0fe86ba7396114b"
 
 [[package]]
 name = "turso"
-version = "0.7.0"
+version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cac8d131650c1bf6a2721202208b3dfba218be25c975f48bebda766173fbdc3b"
+checksum = "6b6c49aecd4abae3ffa88ab1e28b3de7a7497a9732be828e3a6f1855b6ea80fd"
 dependencies = [
  "thiserror",
  "tracing",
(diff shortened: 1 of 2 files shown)


tobert:
## Cross-family review (kaibo `consult`, cast `deepseek`)

Verdict: **safe, merge.** Independent read grounded in the code:

- `src/store.rs`'s turso usage (open/connect/execute/query/pragma, `turso::Error`/`turso::Value`) is the standard embedded surface — simple point queries and small aggregations, nothing that touches the two upstream `core/translate` fixes (IN-LIST cost model, missing-index-after-UPSERT). Those fixes make already-correct queries no less correct.
- Cargo.lock shows a clean bump — all 9 turso subcrates land on 0.7.1, no 0.7.0 remnants.
- `tests/store.rs::survives_reopen` and `::reopen_does_not_re_run_migration` are exactly the tests that would catch a WAL-format regression (reopen + `user_version` pragma survival) — both green.
- Invariants re-verified against the lockfile: `mimalloc` absent (default-features=false held), `turso_sync_engine`/`turso_sync_sdk_kit` present only as unused deps with no HTTP transport in their own dep tree (`sync` feature not enabled), zero `aws-lc-rs`/`aws-lc-sys`/`openssl-sys`/`cmake` anywhere.

Full review: https://github.com/tobert/kaibo/pull/106#issuecomment (posted inline above)

🤖 via kaibo `consult` (cast `deepseek`, explorer `deepseek-v4-flash`, synth `deepseek-v4-pro`)

PR #108, opened 2026-07-29
fix: name --root/--allow-path in the state-db-collides-with-project-tree error

tobert:
## Summary

Amy hit this today running `kaibo` with no `--root`/`--allow-path` while her cwd was her home directory:

```
Error: failed to open the persistence state db at /Users/atobey/.local/state/kaibo/state.db: state db path must live outside every allowed project tree, but /Users/atobey/.local/state/kaibo/state.db is inside one
```

**Root cause (unchanged, not weakened here):** with no `--root`/`--allow-path`, kaibo's default allowed tree is the launch directory (`src/server/resolver.rs`'s cwd-inference branch). The default state db path is `$XDG_STATE_HOME/kaibo/state.db`, defaulting to `~/.local/state/kaibo/state.db` — a subdirectory of `~`. Running kaibo unscoped from `~` makes the default state db collide with the default allowed tree, and `SessionStore::open`'s containment guard (`src/store.rs`) correctly refuses to open it: that guard is what keeps a model driving kaish from ever reaching another project's session history through a read-only project mount. This PR does not touch that check, `resolver.rs`'s cwd-inference, or `default_state_db_path`'s XDG resolution.

**What was actually wrong:** the error message. `main.rs` already special-cases `StoreError::SingleProcessLocked` with a tailored explanation, but every other open error — including this one — fell through a generic catch-all that suggested only `--no-persistence`/`--state-db`, never mentioning `--root`/`--allow-path`, which is usually the more natural fix for *this specific* collision (narrow what's mounted, rather than moving the db or giving up persistence for the run).

## Changes

- `src/store.rs`: extracts the hint text into a pure `path_in_allowed_tree_hint(state_db_path: &Path) -> String` function next to `StoreError`, doc-commented in the file's existing style. It names all three real fixes (`--root`/`--allow-path`, `--state-db`/`KAIBO_STATE_DB`/`[persistence] path`, `--no-persistence`/`KAIBO_NO_PERSISTENCE`/`[persistence] enabled = false`) and keeps the existing reassurance that kaibo never deletes the file.
- `src/main.rs`: adds a second special-cased match arm (parallel to the existing `SingleProcessLocked` arm) that matches `StoreError::PathInAllowedTree` specifically and calls the new hint function. Still a hard, loud `Err` — no silent fallback, per the project's "crash over corrupt/silent" stance. Every other open error still falls through the original generic arm unchanged.
- `src/store.rs`: a failing-first unit test (`path_in_allowed_tree_hint_names_every_real_fix`) that asserts the hint mentions `--root`, `--allow-path`, `--state-db`, and `--no-persistence` together, so a future edit can't silently drop one. I red/green'd this for real: the function first contained the *old* generic message verbatim (missing `--root`/`--allow-path`), confirmed the test failed, then updated the message content and confirmed it passed.
- `CHANGELOG.md`: a `### Fixed` entry under `[Unreleased]` (the `[0.2.0]` section is untouched).

## New message (reproduced end-to-end against the exact scenario)

```
Error: failed to open the persistence state db at /home/.../state.db: state db path must live outside every allowed project tree, but /home/.../state.db is inside one
This usually means kaibo ran with no --root/--allow-path, so its default allowed (read-only, model-reachable) tree is the launch directory — and that directory contains the state db's default XDG location. Three real fixes: narrow what kaibo mounts (--root <project-dir> or --allow-path <dir>), move the state db elsewhere (--state-db <path>, KAIBO_STATE_DB, or [persistence] path in config.toml), or skip persistence for this run (--no-persistence, KAIBO_NO_PERSISTENCE, or [persistence] enabled = false in config.toml). If the db really is corrupt, back it up and move it aside by hand — kaibo creates a fresh one on the next start. kaibo never deletes /home/.../state.db on its own: it holds your session history, which is model output you paid for.
```

## Test plan

- [x] `cargo build` clean
- [x] `cargo test` — full suite green (all existing `PathInAllowedTree` containment tests in `tests/store.rs` still pass unchanged, proving the guard itself is untouched)
- [x] `cargo clippy --all-targets` clean
- [x] `rustfmt --check --edition 2021` on the two touched files (no repo-wide `cargo fmt`, per repo convention)
- [x] Manually reproduced Amy's exact scenario (`HOME` set to an empty dir, no `--root`/`--allow-path`, bare `kaibo`) and confirmed the new message renders as shown above
- [x] New unit test red→green: failed with the old generic-message content, passed once the hint text was updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,20 @@ record. Each later release appends a new section at the top.
 
 ## [Unreleased]
 
+### Fixed
+
+- **State-db-collides-with-project-tree error now names `--root`/`--allow-path`.**
+  The state db's default path (`~/.local/state/kaibo/state.db`) can land inside
+  kaibo's default allowed tree when kaibo runs with no `--root`/`--allow-path`
+  and the launch directory is the home directory (or another ancestor of the
+  state dir) — startup then refuses to open the db, correctly, since a model
+  must never be able to reach kaibo's cross-project session store through a
+  read-only project mount. The refusal itself is unchanged and stays as strict
+  as ever; only the error message improves; it previously suggested only
+  `--state-db`/`--no-persistence`, leaving out the fix that's usually the right
+  one here — narrowing the allowed tree with `--root <project-dir>` or
+  `--allow-path <dir>` so it no longer contains the state db.
+
 ## [0.2.0] — 2026-07-29
 
 ### Changed
--- a/src/main.rs
+++ b/src/main.rs
@@ -224,6 +224,20 @@ async fn serve(common: CommonArgs, gates: ServeGates) -> Result<()> {
                 );
                 handler
             }
+            // A special-cased arm, like `SingleProcessLocked` above: the generic message
+            // below never mentions --root/--allow-path, which is actually the most natural
+            // fix for THIS error — the default allowed tree collides with the state db's
+            // default XDG path only when kaibo runs unscoped from (an ancestor of) the
+            // launch directory, so narrowing the mount is usually the right move, not
+            // relocating the db or giving up persistence. Still a hard, loud `Err` — no
+            // silent fallback here, unlike the one carve-out above.
+            Err(e @ kaibo::store::StoreError::PathInAllowedTree(_)) => {
+                return Err(anyhow::anyhow!(
+                    "failed to open the persistence state db at {}: {e}\n{}",
+                    path.display(),
+                    kaibo::store::path_in_allowed_tree_hint(&path)
+                ));
+            }
             Err(e) => {
                 return Err(anyhow::anyhow!(
                     "failed to open the persistence state db at {}: {e}\n\
--- a/src/store.rs
+++ b/src/store.rs
@@ -177,13 +177,14 @@ pub fn path_in_allowed_tree_hint(state_db_path: &Path) -> String {
     format!(
         "This usually means kaibo ran with no --root/--allow-path, so its default allowed \
          (read-only, model-reachable) tree is the launch directory — and that directory \
-         contains the state db's default XDG location. Three real fixes: narrow what kaibo \
-         mounts (--root <project-dir> or --allow-path <dir>), move the state db elsewhere \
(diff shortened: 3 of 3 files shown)


tobert:
## Cross-family review (kaibo `consult`, cast `deepseek`)

Ran kaibo's own `consult` tool (cast `deepseek`, explorer `deepseek-v4-flash` / synth `deepseek-v4-pro`) against `src/store.rs`, `src/main.rs`, and `CHANGELOG.md` on this branch, specifically asking it to confirm the containment guard in `SessionStore::open` is unchanged and to sanity-check the new hint function, its test, and the new `main.rs` match arm.

**Verdict: containment guard confirmed unchanged.** Quoting the review:

> The containment logic in `SessionStore::open` (lines 263–274 of `store.rs`) is the same strict block that shipped before this PR... The PR adds zero lines inside `open`... The guard stays exactly as strict — `PathInAllowedTree` is still a hard `Err`, and the resolution/comparison logic is byte-for-byte what it was.

It also checked the new `main.rs` match arm's ordering (specific-before-general, same pattern as the existing `SingleProcessLocked` carve-out), confirmed it still returns a hard `Err` (no silent fallback), and confirmed the match is exhaustive.

**One real finding, since fixed (67161be):** the hint text originally included "If the db really is corrupt, back it up and move it aside by hand" — but `SessionStore::open` checks containment *before touching disk at all* (see its own comment), so a `PathInAllowedTree` failure never means the db is corrupt; it means the db was never opened. That line has been dropped from the hint and replaced with a sentence that says the db itself is fine — kaibo just refused to open it because the path lands inside a project it must keep read-only.

Full review transcript available on request; summarizing the rest as minor/non-blocking:
- Noted the resolved path appears twice in the final composed message (`main.rs`'s `{}: {e}` prefix plus `StoreError::PathInAllowedTree`'s own `Display` already containing the path) — harmless redundancy, left as-is.
- Noted the hint's "This usually means kaibo ran with no --root/--allow-path" framing is a *likely cause*, not a guarantee (e.g. it can also fire if `--root`/`--state-db` are both set but happen to nest) — but the three fixes it names are still the right ones regardless, so left as-is.

Tokens: 43498 in / 4670 out / 9216 cached.

PR #109, opened 2026-07-30
docs: pull the 解剖 kanji back to the README's Name section

tobert:
## What

The 解剖 kanji comes out of the 11 *identity* positions it had accreted into, and stays in exactly one place: the README's `## Name` section.

Removed from: the MCP server instructions lead + `with_title`, the CLI `--help` banner, three crate/module doc-comments (`main.rs`, `lib.rs`, `cli.rs`), `AGENTS.md`'s title, the `docs/issues.md` and `docs/devlog.md` headers, `docs/config.example.toml`'s header, and the README banner's `alt` text.

Kept in: the README `## Name` section — plus two new sentences making the situation explicit, so the absence everywhere else reads as a decision rather than drift.

## Why

Amy's call while cutting v0.2.0: *"the kaibou kanji is more happy coincidence than actual name."* The name was built as kai + aibo, to rhyme with gpal/dpal/cpal; that it also reads as 解剖 (dissection, autopsy) was discovered afterward. Stamping `kaibo (解剖)` onto every banner and header quietly asserted the opposite — that the kanji *is* the name.

Deliberately **not** a scrub. Amy asked to de-emphasize, not erase, and the etymology is genuinely good README material.

Two small side benefits: the MCP instructions lead is the resident, 2048-char-budgeted client-facing string, so this buys back a few characters of the budget the `instructions_*` tests enforce; and `Cargo.toml`'s `description` was already done during the v0.2.0 cut, so this finishes that thread.

## Risk

Docs and display strings only — no behavior, flags, or tool names change. The one test that pinned the handshake lead string (`kaish_syntax.rs`) follows the text it pins. Full `cargo test` green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,4 +1,4 @@
-# AGENTS.md — kaibo (解剖)
+# AGENTS.md — kaibo
 
 Kaibo is a stdio MCP server that provides an assistant agent **for other agents**.
 It augments a calling agent (Codex, Claude, Gemini, local agents, etc.) with a team of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,15 @@ record. Each later release appends a new section at the top.
 
 ## [Unreleased]
 
+### Changed
+
+- **The tool is plain "kaibo" now.** The 解剖 kanji no longer rides along with the
+  name in the MCP handshake, the CLI `--help` banner, the example config, or the
+  docs — that reading is a coincidence of how the name was built (kai + aibo), not
+  the name itself, and repeating it everywhere implied otherwise. The README's
+  `## Name` section keeps the story and is now the one place it appears. No
+  behavior, flags, or tool names changed.
+
 ### Fixed
 
 - **State-db-collides-with-project-tree error now names `--root`/`--allow-path`.**
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
   <!-- Kaile, the kaibo mascot. Brand assets + generator live in docs/brand/. -->
-  <img src="docs/brand/banner-teal.png" alt="kaibo（解剖）" width="820">
+  <img src="docs/brand/banner-teal.png" alt="kaibo" width="820">
 </p>
 
 <p align="center">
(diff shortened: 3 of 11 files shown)


tobert:
## Cross-family review — GPT (`~openai/gpt-latest` synth / `~openai/gpt-mini-latest` explorer, via kaibo `consult`)

Reviewed holistically against the worktree. It caught something I'd missed, and it matters.

### Finding 1 (High, accepted): the brand wordmark still reads `kaibo 解剖`

I grepped only `*.rs`/`*.md`/`*.toml`, so I missed **the most prominent identity position in the project** — the README's own hero banner:

- `docs/brand/generate.py:52` hard-codes the wordmark as `kaibo 解剖`
- `docs/brand/banner-teal.svg`, `banner-invert.svg`, `banner-green.svg` all carry it
- `docs/brand/banner-teal.png` — the image `README.md:3` actually displays — renders it
- the generator only emits `banner-*.svg`, so the PNG needs a separate re-render

That made the README's "this is the one place it appears" plainly false — as the reviewer also noted, `CHANGELOG.md` falsifies it too.

**Fixed the claim, left the art.** The section now says the true and more useful thing: the kanji doesn't ride along with the *name* (kaibo introduces itself as plain kaibo in the handshake, `--help`, and the docs), and it stays in two places where it isn't a claim about what the tool is called — the `## Name` section, where it's the story being told, and the brand wordmark, where it's deliberate lettering.

**Whether the wordmark itself should change is @tobert's call**, not a side effect of a docs PR — it needs the generator edited, three SVGs regenerated, and the PNG re-rendered with Noto Sans CJK. Flagged, not done. My read: a logo is exactly where a piece of wordplay is *earned*, and "de-emphasize, don't scrub" argues for leaving it.

### Finding 3 (Low, accepted): the order test was weaker than it read

```rust
let lead_at = text.find("kaibo").expect("opens with the lead");
```

The reviewer correctly judged this non-vacuous (the first later `kaibo` is inside the Casts section's `kaibo://config`, so deleting the lead would still fail the ordering) but noted it doesn't prove the text *starts* with the lead. Now asserts `text.starts_with(kaibo_lead())`, which is what "the lead" actually means — and it's the resident pitch plus the tool-search retrieval index, so verbatim is the right bar.

### Finding 4 (Low, accepted): broken intra-doc link

`src/kaish_syntax.rs:257` linked `[kaibo_instructions]`, a symbol that no longer exists (the real one is `kaibo_instructions_with_scope`). Pre-existing, but in a doc-comment this PR touches. Rewritten to describe the function directly.

### Also taken

README's "kai + aibo" reads like it should produce "kaiaibo" — now notes the overlapping `ai`.

### Confirmed clean

Crate/binary docs, CLI help, MCP `Implementation` title, and the server instructions all use plain `kaibo`. The one remaining `解` in `src/progress.rs:263` is a Unicode-truncation test, not a name usage — correctly left alone.

PR #110, opened 2026-07-30
Don't advertise a tool no configured cast can staff

tobert:
## What

A tool now clears **two** gates to be advertised: its `--no-<tool>` flag, and a configured cast that can actually staff it. A tool nothing can staff has its route removed rather than shipping unusable.

## Why

`deliberate` was **dead on arrival** for anyone running stock kaibo. It needs an explorer paired with an offline synth; the two built-in offline casts (`anthropic-batch`, `gemini-batch`) are synth-only, so `cast_can_deliberate` matched nothing — every call failed `cast "…" has no explorer slot`, while the tool still billed resident tokens every session and its own description pointed at casts that exist only in `docs/config.example.toml`.

The machinery was already there, pointed one step short. `CAST_ENUM_RULES` computed per-tool cast eligibility from live config, but only fed the `cast` *enum*. `inject_cast_enum` documented the near-miss itself: given an empty roster it **skipped** injecting, because an empty `enum` reads as "no valid value" to a strict client — leaving the tool advertised and broken. Making an empty roster remove the route generalizes for free to `explore`, `batch_submit`, `consult`/`oneshot`, and the collect verbs.

Amy: *"deliberate should be disabled without a configured cast, so it doesn't take up tool space. Tools should not be presented if they are not usable as a default policy."*

## Decisions

- **One source, not two.** `live_tools()` folds flags + staffing into the surviving set; the router filters through it and `kaibo://config` reports it. The CLI renders that resource with no router in existence, so a second derivation is precisely how the resource would come to describe a surface the server doesn't serve.
- **Collect verbs follow *effective* liveness.** A producer nothing can staff mints no handles, so it can't justify keeping `job_*` resident any more than a flag-disabled one can.
- **The reason is said twice.** Vanishing is right for the model's tool list and wrong for the operator: a startup warning names the cast *shape* that would bring the tool back, and `[runtime]` gains `advertised_tools` + `unstaffable_tools`. That requirement text lives in the `CAST_ENUM_RULES` row beside the predicate it describes, so it can't drift.
- **"You disabled it" ≠ "nothing can run it."** `unstaffable_tools` deliberately omits flag-disabled tools; `[tools]` already reports those.

## A test-hygiene problem this surfaced

Handler tests built from a fixture TOML were really testing *"my fixture plus whatever API keys are on this machine"* — the built-in cast registry merges under every config, and a built-in cast counts as usable the moment its backend resolves a credential from env **or a key file on disk**. That's why the first cut appeared to gate `deliberate` but not `explore`/`batch_submit`.

Fixed with `new_with_env` (injecting the credential lookup — the closure pattern `config.rs` already uses, deliberately not `set_var`) plus fixtures that neutralize the built-in roster explicitly. `tests/gating.rs` now separates concerns: a `FULLY_STAFFED` fixture so the flag tests measure flags, and one test pinning the DOA bug itself.

## Docs

`[server.tools]` gains the two-gates explanation (it was also missing `explore` and `deliberate` entirely), and the DELIBERATE casts section says plainly that it's what turns the tool on — all three hosted batch providers work there, as does a big local model on the `direct` lane.

## Note for the media CAS

An image tool inherits this gate by adding one `CAST_ENUM_RULES` row, so a user who configures no image cast pays zero resident tokens for it. Amy's CAS shape decisions (image as a cast slot; warn-don't-refuse on ephemeral storage; return digest+path, not bytes) are recorded in `docs/issues.md`.

## Testing

Full `cargo test` green, `clippy --all-targets` clean. Seven new staffing tests plus two `kaibo://config` tests; the shipped example config still parses.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,10 +37,14 @@ loop. Each consultation tool is that loop wearing different clothes:
 Both model-driven tools name their cast + answering model(s) in a provenance footer
 (`with_provenance` in `server.rs`), so a cross-model study sees which model answered.
 
-Each tool is independently gated by a `--no-<tool>` flag (all on by default; the
-all-off server is refused at startup). Multi-provider over `rig-core`: a
-**`ProviderKind`** is the wire protocol (keyed Anthropic / DeepSeek / Gemini, plus
-**`openai`** for any OpenAI-compatible endpoint). A **`[backends.<name>]`**
+Seven `--no-<tool>` capability flags gate the surface (all on by default; `consult` also
+gates `consult_submit`, `batch` gates `batch_submit`, and the `job_*` verbs follow their
+live producers). A tool also needs a cast that can **staff** it, or its route is dropped;
+a server left with nothing advertised is refused at startup, by either road. See
+`CAST_ENUM_RULES` and `live_tools` in `server.rs`, and "Tool gating" in `docs/config.md`.
+Multi-provider over `rig-core`: a **`ProviderKind`** is the wire protocol (keyed
+Anthropic / DeepSeek / Gemini / OpenRouter, plus **`openai`** for any OpenAI-compatible
+endpoint). A **`[backends.<name>]`**
 (`config.rs`) is a *named connection* of a kind with its own base URL and key source —
 so two `openai` backends (hosted GPT and a local Gemma/llama.cpp server, say) can be
 live at once. A **`[casts.<name>]`** is a model team mapping each reasoning role
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -216,6 +216,20 @@ Two audiences are optimized differently:
   of thousands to millions of tokens. Here verbosity is *licensed where it shapes
   behavior*: say it a few ways, frame positively, be explicit (see **Driving the
   models**). Verbose to install behavior, never verbose by default.
+- **Operator-facing docs** — `docs/config.example.toml`, `docs/config.md`, `README.md`.
+  The first two are **embedded in the binary** (`include_str!`) and served as
+  `kaibo://config/example` and `kaibo://config/guide`, so a model reads them as often as
+  a person does. They are shipped product, not repo notes. Write them as **technical
+  reference**: state the rule, name the default, show the shape. Declarative sentences
+  over conversational asides; a table or labelled list over a paragraph that buries three
+  facts in a clause chain. No em-dash pile-ups, no rhetorical questions, no voice.
+  Someone skimming for one key should find it without reading the prose around it, and a
+  non-native English reader should not have to parse an idiom to get a default value.
+  Split the two by job: the **template** says what to type (a knob, its default, one line
+  on what it does, a pointer for the rest); the **guide** explains semantics and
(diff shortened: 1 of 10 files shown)


tobert:
## Cross-family review — DeepSeek (`deepseek-v4-pro` synth / `deepseek-v4-flash` explorer, via kaibo `consult`)

Reviewed holistically against the worktree, not as a diff. Verdict: staffing logic correct, `live_tools()` genuinely single-source, `ToolGating::enabled` mapping complete, new tests fail on regression and the hermetic fixtures are deterministic. **One real finding**, now fixed in 09b00b6.

### Finding (accepted, fixed): the zero-tool startup refusal didn't cover staffing

`ToolGating::all_disabled()` runs before the handler exists and reads only the `--no-<tool>` flags, so it couldn't see the second road to an empty surface: every cast-taking tool left ON but unstaffable, with `run_kaish` and `list_models` disabled. That server started and served an empty tool list in silence.

Reproduced by hand before fixing — all four staffing warnings fire, tool list empty, kaibo proceeds to serve. Narrow to reach (the keyless `openai-local` cast survives by default, so it takes an explicit `key_optional = false` plus an unreachable key file to kill it), but a real hole in the crash-over-silently-useless posture, so it now gets the same loud refusal the flag case already had.

**The test for it is worth calling out.** Written first, it *passed before the fix* — for the wrong reason. It runs the binary with stdin closed, and a kaibo that reaches `serve()` exits non-zero anyway (`connection closed: initialize request`), so a bare exit-status assertion couldn't distinguish the guard from the transport giving up. It now pins the guard's own message, and I re-verified by removing the guard and watching it fail. A test that can't fail is worse than no test.

### Confirmations worth recording

- **Two-phase gate is sound.** `require_deliberate_cast` checks only the synth lane; the explorer requirement is caught at arm-resolve. Since the `CAST_ENUM_RULES` predicate already filtered to explorer-bearing casts, an advertised cast always resolves its explorer arm. `explore`'s gate is deliberately `true` — loose at the gate, tight at the enum.
- **`live_tools()` callers can't disagree.** Both feed it `usable_casts` off the same env-access pattern (`new_with_env` gets `|k| std::env::var(k).ok()` from `new`), and the environment is process-stable between startup and a later resource read.
- **The stock-install invariant is credential-independent.** Adding keys makes more built-in casts *usable* but never creates one with both an explorer and an offline synth — so `a_stock_install_does_not_advertise_deliberate` holds with a full keyring or an empty one.
- **`_ => true` in `ToolGating::enabled` is correct for the collect verbs.** They have no flag; liveness comes from `jobs_live`. They also never reach `unstaffable_tools`, since they're absent from `CAST_ENUM_RULES`.

### Noted, not acted on

`tests/gating.rs` has no test proving `FULLY_STAFFED`'s `deep` cast staffs `deliberate` directly — it's implied by `default_advertises_all_tools` asserting against the complete `ALL_TOOLS` list, and covered explicitly by `deliberate_is_advertised_when_a_cast_can_staff_it` in the server tests. The reviewer judged the split fine (staffing half in server tests, flag half here, with the full-list assertion as the seam); I agree.

PR #115, opened 2026-08-02
fix: an empty consult answer errors or recovers, never returns as success

tobert:
A `consult` came back with a body containing nothing but the provenance footer. No
error, no partial, no signal that the review had not happened. The caller merges on
that, which makes it the worst failure shape available.

## What actually happened

Reported as: *"we've had that a couple times recently with deepseek."* Two runs, same
question, same path, both stateless — one on `deepseek` returned an empty body, one on
`or-kimi` returned a full cited review.

The OTLP trace settled it. The driver ran **16 turns of a 200 cap** — no exhaustion. Its
**terminal turn produced 14 output tokens and no text**, and its last act was
`grep -rn "fn active_id"`. It stopped mid-investigation without ever writing an answer.

The mechanism, reproduced offline with no key and no network: DeepSeek returns a
reasoning block with empty text; rig drops the empty text block, keeps the reasoning
block, sees no tool calls, takes the clean-finish branch, and extracts text by filtering
to `Text` only. Result: `Ok("")`. **rig pins that as intended behavior in its own test
suite**, so this check cannot be delegated upstream. kaibo then dressed the empty string
in a footer with real token counts and returned success.

There was no empty-answer guard anywhere on the interactive lane. The batch lane has had
one since GH #75 — `finish_gated_answer`, "a completed-but-empty answer is no answer."
The interactive lanes just never grew it.

A fourth path fell out of the same root: an empty **explorer report** reached the driver
as a blank tool result, so it reasoned over an empty sweep and never learned it failed.

## The fix: recover when it can, fail loudly when it can't

Checked at the one seam every phase runs through, so `consult`, `explore`, `oneshot`, and
`deliberate` are covered together.

**Evidence present → ask once, and only once.** If the model had already gathered
evidence and simply stopped before writing, kaibo replays the transcript with
`ToolChoice::None` and its own note ("you stopped without writing your answer", distinct
from the turn-cap wording). This usually recovers the answer that was already paid for —
in the reported run the driver held everything it needed. Both attempts' tokens are
counted; usage is **added, not substituted**.

**No evidence → error, and do not ask again.** This half is load-bearing, not a fallback.
Pressing a model with no evidence to "answer anyway" is how you get a confident review
with nothing behind it — strictly worse than an error on a tool whose entire product is
grounded citation. So the case where a retry is least likely to succeed is also the case
where succeeding is dangerous. That reasoning is in the code comment, because otherwise
someone will later "simplify" it into an unconditional retry.

The error names what it cost — turns used against the cap, input/output/cached tokens —
and says plainly that the provider's finish reason was **not available** on this wire,
rather than implying it was checked.

## Two places the implementation diverged from its brief, both deliberate

- **No `finalized` flag.** At-most-once is structural: the empty-answer arm returns on
  every path, so it cannot re-enter the loop. The compiler confirmed it, warning the flag
  was never read. A shape that forbids stacking beats a bool a later edit could reset.
- **The gate is "tool results present", not "successful tool results".** rig folds a
  *failed* tool call into a tool result string, so success and failure are only
  distinguishable by sniffing result text — the kind of silent heuristic this codebase
  refuses. The line still holds: a transcript of only tool *errors* yields an honest
  grounded answer ("I could not read X"), and what the gate excludes is the genuinely
  dangerous shape — zero tool interaction, pure generation, nothing to cite. Flagged
  because it is weaker than the words the fix was specified in.

## Tests

Four tests were written **red first**, before any fix, reproducing the empty body through
the real consult loop with a scripted client. The gate is proven **both ways**, including
asserting the *absence* of a second request in the no-evidence case — an implementation
that retried blind and got another empty would pass an error-only check while being
exactly wrong.

One red test was retargeted rather than kept: it scripted the reported run's exact shape,
which is precisely the evidence-present case, so the fix made it *recover*. Leaving it
error-expecting would have pinned the wrong behavior. A new test covers the slot it
vacated.

## Verified, not assumed

- `messages` is `Some` by construction on the `Ok` path (one return site, always
  `.with_messages`). `None` is still handled, as the **error** case — failing closed.
- Usage is `spent + retry_usage`, never replaced. `completion_calls.len()` is used only as
  a turn count for diagnostics, never for usage accounting.

## Still open, deliberately

`finish_reason` is genuinely unavailable here: rig's agent hook event carries `prompt`,
`content`, `usage`, `message_id` and no raw response. A follow-up branch wraps
`CompletionModel` to capture it — at which point this error stops having to say the field
was unavailable.

795 passed / 0 failed. `clippy --all-targets` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,31 +122,15 @@ record. Each later release appends a new section at the top.
 ### Fixed
 
 - **A consultation that produced no answer no longer comes back as a successful empty
-  one.** A `consult` could return a body containing nothing but the provenance footer —
-  no error, no partial, no signal that the review had not happened. Caught on a reasoning
-  model whose final turn came back as reasoning with no answer text: the underlying `rig`
-  client treats a textless final turn as a clean finish and hands kaibo an empty string,
-  which kaibo then dressed in a footer with real token counts and returned as success.
-  That is the worst failure shape there is, because the caller merges on it.
-
-  kaibo now checks every phase's answer at the one place they all run through, so
-  `consult`, `explore`, `oneshot`, and `deliberate` are covered together. What you see
-  depends on what the model had in hand:
-
-  - If it had already gathered evidence and simply stopped before writing, kaibo asks it
-    once — and only once — to write up what it found. This usually recovers the answer
-    you paid for, and the tokens from both attempts are counted in the footer.
-  - If it had gathered nothing, kaibo fails the call instead of asking again. Pressing a
-    model with no evidence to "answer anyway" is how you get a confident review with
-    nothing behind it, which is worse than an error on a tool whose whole product is
-    grounded citation.
-
-  Either way the failure names what it cost — turns used against the cap, input/output/
-  cached tokens — and says plainly that the provider's finish reason was *not* available
-  rather than implying it was checked. A delegated `explore` sweep that returns nothing
-  is now a visible failure to the model driving it, instead of a blank result it would
-  have quietly reasoned over. Batch results already worked this way; the interactive
-  lanes now match.
+  one.** A reasoning model's final turn can carry reasoning but no answer text, and
+  kaibo would dress that empty string in a provenance footer and return it as success —
+  no error, no signal that the review had not happened. Every phase (`consult`,
+  `explore`, `oneshot`, `deliberate`) now checks its answer: if the model had already
+  gathered evidence, kaibo asks it once to write up what it found, which usually
+  recovers the answer (both attempts' tokens counted in the footer); if it had gathered
+  nothing, the call fails with its diagnostics attached (turns used, token counts, and
+  the provider's own finish reason when it reported one) rather than pressing an
+  evidence-free model into an ungrounded answer.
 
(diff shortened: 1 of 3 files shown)


tobert:
Two updates since the overnight state, now that #116 is merged:

**Merge reconciliation (`e49556f` + `ea6cb5c`).** The merge of main was textually clean and semantically wrong, again: #116 moved `oneshot`/`deliberate_direct` off `run_phase` onto `Arm::complete`, which silently dropped the empty-answer guard from both toolless lanes, and `run_phase_loop` lost `model_name`. The guard now lives in `run_completion` too (toolless = no evidence = fail immediately, never re-ask), and — the convergence the signoff predicted — the empty-answer diagnostics now read the provider's `finish_reason` off #116's `CompletionLog` instead of claiming it's unavailable. `None` is reported as honest absence, since the OpenAI Responses wire reports nothing on a clean stop. The no-evidence test stamps a raw payload to pin the plumbing.

**Review feedback.** Both forced-finish notes now say that naming what you do not know is part of a complete answer; the "— right now" tail is gone; the CHANGELOG entry is compressed to what a user notices.

Full suite green (544 lib + integration), clippy silent, rustfmt diffs in `engine.rs` are all pre-existing from main.

PR #117, opened 2026-08-02
docs: adopt kaish's writing style guide by reference

tobert:
Closes the last item of the style-guide propagation: kaish `docs/style.md` (kaish #260) names kaibo and kaish-extras as adopters; extras adopted in extras#6, kaibo now matches.

The adoption layers **under** the existing Writing-for-models audience rules, which stay in charge — where the guide's "keep the why" collides with the 2048-character resident budget, the budget wins. Weight map: full weight on error/diagnostic strings and client-facing tool descriptions; Terms only on README and docs (kaish's Terms table binds where kaibo uses kaish's words); devlog/signoff exempt. Groom at the point of touch; there is no bulk pass.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -249,6 +249,23 @@ Two audiences are optimized differently:
   start to finish by whoever is configuring kaibo, so every line there is a line they pay
   for.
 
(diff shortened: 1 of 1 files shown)


tobert:
Parking this until the in-flight kaibo work settles — the branch stays; reopen when ready.

PR #118, opened 2026-08-02
chore: follow-ups from the merged attach/empty-answer/rig-0.41 work

tobert:
The four follow-ups agreed after the 2026-08-02 merge sweep (#114/#115/#96/#116), each small on its own:

**Template coverage guard.** `kaibo://config/example` promises "every option with its default", and history says the gap recurs — PR #110 found four missing knobs, then `max_attachments` shipped in #96 without a template line until the merge reconciliation added one by hand. The new test asks serde itself for every raw struct's field set (`deny_unknown_fields` bakes the list into the `Deserialize` impl, and the unknown-field error recites it), then requires each name to appear in the template — no hand-kept list to drift from the parser. Teeth proven: deleting `max_attachments` from the template fails the test naming `RawDefaults.max_attachments`.

**The deliberate image hand-off is pinned at the server seam.** The residual worry from the explorer-attach cross-family review: a refactor could drop the routed `images` between `sink.drain()` and lane dispatch with nothing failing. The drain-and-stitch step is now an extracted `stitch_dossier_delivery`, and a test drives the *real* `attach` tool over a real workspace (PNG + text file), stitches, and dispatches to a scripted synth — proving the routed image survives to the offline synth's single turn as a native image part.

**Stale "four built-ins" text.** `tests/config.rs` and the config-resource test now enumerate all five kinds (OpenRouter included), so silently dropping a built-in fails the suite; the `Lane::Direct` doc stops claiming no tool routes to it.

**`ModelTurnAction::Retry`: evaluated, rejected, recorded.** The plan was to convert #115's hand-rolled forced write-up turn to rig-agent's first-class retry. Reading the API says no: `Retry(Feedback)` cannot carry `ToolChoice::None` (the retried turn could burn more tool calls instead of writing), a hook retry consumes the run's turn budget (an empty answer at the cap would get no recovery, where today's path deliberately runs the write-up outside the budget), and `Stop(reason)` is a string that would flatten `empty_answer_error`'s typed diagnostics. The decision now lives as a comment on the recovery arm, with a pointer to re-evaluate if rig grows a constrained retry.

Internal-only (tests, comments, tracker) — no CHANGELOG entry. The three shipped `docs/issues.md` entries are deleted per the tracker discipline. 849 tests green, clippy silent, fmt drift only where authored.

### DIFF
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -334,6 +334,22 @@ DeepSeek CLI-subcommands review (2026-07-17).
 
 ## P3 — Infra, perf, polish
 
+### Batch results don't cross-check returned `custom_id`s against the submitted set
+From the 2026-08-02 batch error-propagation audit (DeepSeek agentic consult, key claims
+verified by hand — the audit's overall verdict was that per-item propagation is SOUND:
+loud per-item `Err`s on all three providers, `finish_gated_answer` on every succeeded
+path at `batch.rs:548/986/1625`, no error detail discarded). The one blind spot: none of
+the three result parsers (`parse_results_jsonl` `batch.rs:524`, `parse_gemini_inlined`
+`:975`, `parse_openai_output_jsonl` `:1579`) verifies that every submitted `custom_id`
+came back. A provider that silently drops an item yields N-1 results with no error about
+the missing one — "Batch complete — 9 result(s)" after submitting 10. Fix shape: carry
+the submitted count (or id set) on the handle and emit a synthetic per-item failure for
+each absentee at parse time. Two accepted-as-designed observations from the same audit,
+recorded so they aren't re-flagged: Anthropic has no batch-level `Failed` state ("ended"
+covers all outcomes; per-item errors live inside `Done`, and `job_list` is a triage view —
+`job_get` has the detail), and Gemini's errored-item summarizer probes only
+`error.message` before falling back to the full JSON (data preserved, label coarser).
+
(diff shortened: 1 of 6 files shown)


tobert:
**Cross-family review (Gemini, agentic consult over the branch worktree)** — all four items endorsed:

1. *Template coverage guard*: the serde error-message extraction judged sound (test-only, so a format change breaks a test, not production). **One real finding**: bare `contains()` would let short field names (`id`, `log`, `lane`, `kind`) pass as accidental substrings of prose like `guide`/`changelog`, so the guard could keep passing after a knob's documentation vanished. Fixed in `a7d4790` — the check now requires the name bounded by non-identifier characters, and the template still passes, meaning every field genuinely appears as a real token.
2. *`stitch_dossier_delivery`*: extraction confirmed behavior-identical; the pipeline test "exercises the exact seams claimed" (real PNG, real `KaishWorker`, real `attach` tool, scripted synth asserting the native image part).
3. *Five-built-ins loops*: verified the assertions hold for the `openrouter` built-in specifically (its vision pin is a capability, not a tunable, so the tunables fall through to defaults as asserted).
4. *`ModelTurnAction::Retry` rejection*: all three claims confirmed against rig-agent 0.41's hook design.

Also riding this PR: the batch error-propagation audit result (DeepSeek consult, key claims hand-verified) — propagation is **sound**; its one gap (no `custom_id` cross-check against the submitted set) is now a tracked `docs/issues.md` entry with a fix shape, alongside two accepted-as-designed observations recorded so they don't get re-flagged.

PR #119, opened 2026-08-02
fix: an empty-answer failure is a retryable model outcome, not a kaibo bug

tobert:
## What

The #115 empty-answer guard added a new error vocabulary (`"returned an EMPTY answer"`) that `classify_failure` (`src/server/render.rs`) never learned: its reached-a-model markers predate the guard, so every empty-answer failure fell through to `FailureKind::Internal` and the calling agent read *"This is a kaibo-side error (not the provider) — please report it"* wrapped around inner diagnostics that say *"Retry, or try the same question on a different cast."* Contradictory guidance, blaming kaibo for a model outcome.

Found live the same day the guard merged: a starved deepseek oneshot (256-token synth cap) spent all 256 output tokens on reasoning, returned empty content with `finish_reason "length"`, and the loud error came back labeled a kaibo bug.

## Decisions

- **Empty-answer gets its own `FailureKind`** rather than riding `Provider`, whose "retrying is unlikely to help" would contradict the guard's own diagnostics. The new guidance names the outcome (model ran, no answer text), invites a retry or a different cast, and points at the slot's `max_tokens` when the provider reported a length cutoff — the actual fix for the starved class.
- **The marker counts toward `reached_a_model` but classifies after the transient vocabulary**, so a forced write-up turn that died on an overload keeps the more specific transient framing.
- **The issues.md entry tracking the empty-answer recurrence** (the guard is a backstop, not a cure; characterize from telemetry before touching prompts) lands here. Its classifier-bug bullet ships in this same change, so per the delete-when-shipped convention it is not recorded.

## Tests

Written failing-first: the three empty paths (loop gate, single-shot, forced-turn wrapper) pin retry-inviting, non-kaibo-blaming guidance; the transient-during-recovery case pins transient framing. Full suite green, clippy clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -162,6 +162,14 @@ record. Each later release appends a new section at the top.
   the provider's own finish reason when it reported one) rather than pressing an
   evidence-free model into an ungrounded answer.
 
+- **An empty-answer failure now reads as what it is — a retryable model outcome, not a
+  kaibo bug.** The failure classifier predates the empty-answer guard above and did not
+  know its vocabulary, so every empty-answer error was wrapped in "This is a kaibo-side
+  error (not the provider) — please report it" around inner text that said "Retry" —
+  contradictory guidance blaming kaibo for the model's silence. The guidance now says the
+  model delivered no answer text, invites a retry or a different cast, and points at the
+  slot's `max_tokens` when the provider reported the answer was cut off by length.
+
 - **A vision model still sees the image after the `rig` 0.41 upgrade.** rig 0.41 stopped
   inspecting a tool's text output to discover rich content in it, which would have
   silently turned every `view_image` result into base64 text labelled JSON — the model
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -321,6 +321,39 @@ advice split. Distinct from the upstream rig retry/backoff entry above — that'
 level 429/503 backoff; this is a mid-loop model fumble kaibo can re-prompt around itself.
 Surfaced by the DeepSeek CLI-subcommands review (2026-07-17).
 
+### Empty answers keep happening — the guard is a backstop, not a cure
+The empty-answer guard shipped 2026-08-02 (#115, meeting #116's finish_reason plumbing and
+the attach work in #118): no lane returns an empty answer as success. The tool loop gets one
+evidence-gated forced write-up turn (`consult/engine.rs::run_phase`, `empty_answer_error`),
+the turn-cap finalize and the single-shot lanes fail loudly with diagnostics (turns, tokens,
+the provider's own `finish_reason`), and batch has long had `finish_gated_answer`. But the
+guard changes the *failure mode*, not the *frequency*: DeepSeek still sometimes produces the
+textless terminal turn (Amy, 2026-08-02, same day the guard merged), and other families
+likely do too. Experiments the same day: a starved oneshot (256-token synth cap) repros the
+`finish_reason "length"` class on demand — reasoning ate the whole completion budget, guard
+errored loudly — while three grep-heavy review consults on the default deepseek cast all
+completed clean, so the wild stopped-without-answering class is sporadic (or already
+lowered by #114's role-identity preambles; n=3 can't tell). Open work, in order:
+- **Characterize from telemetry before touching prompts.** The plumbing to answer "which
+  models, how often, which finish_reason, does the forced re-ask rescue it" is all in place:
+  `gen_ai.response.finish_reason` on every `run_phase` span (#116), the `warn` event
+  "phase returned an empty answer with evidence gathered — forcing one final write-up turn"
+  (`engine.rs`), and `empty_answer_error`'s token/turn diagnostics in the error text.
+  Collect real occurrences from OTLP over normal use; the interesting split is
+  rescued-by-forced-turn vs. errored-anyway, and `finish_reason "stop"` (model chose to
+  stop) vs. `"length"` (starved — a `max_tokens` problem, different fix).
+- **Prevention is prompt work, and #114 already aims at it.** The measured failure shape is
+  a driver ending a turn with a tool call as its last act and no text (the 2026-08-01
+  DeepSeek run: explorer report delivered, fifteen greps returned, then a 14-token terminal
+  turn). #114's role-identity preambles ("the synth's final turn *is* the answer") target
+  exactly this; measure whether incidence drops post-#114 before adding more prompt mass.
+- **If the forced re-ask often fails**, consider a second differently-worded re-ask or
+  per-model shaping — but only with data showing the single re-ask isn't enough.
+- **The un-catchable neighbor, recorded so it isn't mistaken for a gap:** a non-empty but
+  vacuous answer ("I will now analyze…" then stop) passes the trim gate by construction.
+  If that shape shows up in practice it's a different problem (answer-quality, not
+  emptiness) and wants its own thinking, not a heuristic bolted onto this gate.
+
 ### Dossier checkpointing via the persistence store
 `deliberate`'s dossier (the synchronous explorer sweep) is the expensive artifact — minutes
 of live exploration — yet it lives only in memory until the offline synth consumes it.
--- a/src/server/render.rs
+++ b/src/server/render.rs
@@ -47,6 +47,11 @@ enum FailureKind {
     TransientProvider,
     /// A non-transient model/provider error (auth, bad request). Retrying won't help.
     Provider,
+    /// A model ran and delivered no answer text (`consult/engine.rs::empty_answer_error`,
+    /// or the forced write-up turn's own failure wrapper). A model-side outcome that the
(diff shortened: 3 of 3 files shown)


tobert:
## kaibo cross-family review — deepseek cast (explorer, synth)

Run via `kaibo consult` over the branch worktree (holistic read, no diff supplied). Verdict: **sound, no changes requested.**

Here is the review, grounded in the code.

---

## 1. The marker heuristic is sound

The classifier at `src/server/render.rs:72-121` uses a two-gate design: first determine whether a model ran at all (the `reached_a_model` gate, lines 88-91), then decide how to frame the outcome. The markers are the four strings kaibo's engine stamps onto errors that originate after a model was invoked:

| Marker | Origin | engine.rs |
|---|---|---|
| `"model loop failed:"` | Any `PromptError` not otherwise caught | `1273` |
| `"model call failed:"` | `run_completion` (oneshot / deliberate direct) | `1445` |
| `"model used all"` | `finalize_after_max_turns` forced-turn failure | `1361` |
| `"returned an empty answer"` | `empty_answer_error` + forced-turn wrapper | `690`, `1225` |

Every site in `engine.rs` that produces one of these markers has genuinely reached a model (the agent loop ran, a direct completion was sent, a forced write-up turn was attempted). The absence of all four markers correctly sends an error to `Internal` (line 93), because the only other failure sites are `make_tools()?` calls at lines 1110 / 1127 / 1218 — kaish kernel builds that fail before the model is ever called. The `call_deadline` backstop (line 79) fires from *around* the loop and carries none of these markers, so its early return before the gate is correct.

No internal error path coincidentally emits any of the four markers, and no marker-bearing error can originate from before the model runs.

---

## 2. Misclassification risks

### (a) An empty-answer message matching transient vocabulary by accident

**Verdict: no false match in the standard text; one deliberate match is by design.**

The `empty_answer_error` body at `src/consult/engine.rs:689-699` reads (lowercased for the classifier at `render.rs:73`):

> model {model} returned an empty answer — {detail}. … diagnostics: … retry, or try the same question on a different cast.

Checked against every entry in `TRANSIENT` (`render.rs:97-113`):

- `"try again"` — close but not present; the text says `"try the same question"`, not `"try again"`. No match.
- None of `overload`, `rate limit`, `rate_limit`, `ratelimit`, `resource_exhausted`, `too many requests`, `timed out`, `timeout`, `connection reset`, `reset by peer`, `connection closed`, `broken pipe`, `temporarily`, `unavailable` appear anywhere in the `empty_answer_error` format string.

The *forced-turn failure wrapper* at `engine.rs:1224-1227` produces:

> model {model_name} returned an empty answer, and the forced final-answer turn also failed: {e}

where `{e}` is the error from `forced_finish_turn`, itself wrapped as `"model call failed: <provider error>"` at line 1445. If the provider error body contains `"overload"` (Anthropic 529), the full lowercased chain matches both `"returned an empty answer"` and `"overload"`, and the transient check fires first at line 114 — classifying it as `TransientProvider`. This is the exact scenario tested at `render.rs:756-767` (`an_empty_answer_whose_forced_turn_hit_a_transient_error_stays_retryable`), and it is correct: the forced turn hit an overload, so framing it as transient with overload-specific retry guidance beats the generic empty-answer message.

### (b) An internal error matching the empty marker

**Verdict: cannot happen with the current code.**

The string `"returned an empty answer"` (lowercased from the `EMPTY` in the format) appears in exactly two places in the production error path:

1. `empty_answer_error` at `engine.rs:689-699`
2. The forced-turn failure wrapper at `engine.rs:1224-1227`

Both are genuine empty-answer outcomes. No kaish kernel error, filesystem error, or any other internal failure path produces this string. The `{:#}` anyhow chain rendering (line 73) would carry the text from an inner error, but only if some code explicitly wrapped an internal error with `"returned an empty answer"` — and there is no such site.

The `tracing::warn!` at `engine.rs:1206` (`"phase returned an empty answer with evidence gathered"`) is a log line, not an error, and never reaches the classifier.

### (c) A provider rejection whose body contains "returned an empty answer"

**Verdict: theoretically possible but vanishingly unlikely; impact is mild.**

Provider error bodies are JSON structures: Anthropic errors are `{"type":"error","error":{"type":"overloaded_error","message":"…"}}`, OpenAI errors are `{"error":{"type":"…","message":"…"}}`. None of the standard provider error vocabularies include the phrase `"returned an empty answer"`. A provider would have to echo kaibo-specific phrasing to produce this false match.

If it did, the classifier would see `empty_answer = true` and `reached_a_model = true`. If the body also contains no transient vocabulary, it would classify as `EmptyAnswer` rather than `Provider`. The guidance for `EmptyAnswer` says *"The model ran but delivered no answer text"* — technically inaccurate for a provider rejection, but the actual error detail is preserved in the tool result (line 145: `"…{detail}. {guidance}"`), so the calling agent reads both the raw error and the guidance. The operational advice — *"You may retry, and a different cast often helps"* — is harmless even for a non-empty-answer provider rejection; worst case the caller wastes one retry.

### (d) Transient vocabulary matching a non-transient provider error

**Verdict: low risk; worst-case impact is one unnecessary retry.**

The transient list at `render.rs:97-113` is built from the providers' actual vocabulary. Most entries are unambiguous: `"timed out"`, `"connection reset"`, `"broken pipe"` only appear in transport failures. The weaker entries are:

- **`"unavailable"`** — could appear in a 403 body like `"model unavailable in your region"`, which is a configuration error, not transient. But such an error would rarely appear wrapped in `"model loop failed"` (it would surface at setup). And the retry guidance is mild: *"you may retry this call, or proceed without the consultation"* — the caller would retry once, see the same error, and proceed.
- **`"try again"`** — provider error messages sometimes say `"Please check your API key and try again"` in a 401 body. This would classify a permanent auth error as transient. Again, impact is one wasted retry, and the full error detail is visible to the caller.

None of these are introduced by the empty-answer change; they are inherent to the string-matching heuristic the doc comment at lines 60-71 already acknowledges.

---

## 3. Guidance text quality

All four guidance strings (`render.rs:147-165`) are clear, actionable, and correctly scoped:

| Kind | Guidance | Assessment |
|---|---|---|
| `TransientProvider` | *"This looks like a transient provider condition … you may retry this call, or proceed without the consultation."* | Names likely causes (overload, rate limit, timeout), states kaibo's no-auto-retry policy, gives two clear next steps. |
| `Provider` | *"The model or its provider rejected the request; retrying is unlikely to help — proceed without the consultation, or check the cast and config."* | Explicitly discourages retry, points to config review. |
| `EmptyAnswer` | *"The model ran but delivered no answer text — a model-side outcome, not a kaibo bug. You may retry, and a different cast often helps; if the diagnostics report finish_reason \"length\", raise that slot's max_tokens so reasoning cannot starve the answer."* | Names it as model-side (not a kaibo bug), gives retry + different-cast advice, adds the concrete `max_tokens` tip for the `finish_reason: "length"` case that produced the live repro (`render.rs:712-713`). |
| `Internal` | *"This is a kaibo-side error (not the provider) — please report it; you can still proceed without the consultation."* | Honest about being kaibo's fault, asks for a report, gives the proceed option. No provider-blaming. |

The `EmptyAnswer` guidance is the most detailed, which is appropriate: it is the least self-explanatory failure mode, and the `max_tokens` tip directly addresses the starvation scenario that triggered this work.

---

## 4. Callers of `consultation_failed` / `consultation_failure_text`

All callers route through the same classification logic:

| Caller | `file:line` | Uses |
|---|---|---|
| `consult` (sync) | `server/mod.rs:1403` | `consultation_failed` |
| `consult_submit` (async) | `server/mod.rs:1545` | `consultation_failure_text` (stores for `job_get`) |
| `explore` (sync) | `server/mod.rs:1622` | `consultation_failed` |
| `deliberate` dossier phase (sync) | `server/mod.rs:1736` | `consultation_failed` |
| `deliberate_submit` (async) | `server/mod.rs:1904` | `consultation_failure_text` |
| `oneshot` (sync) | `server/mod.rs:1968` | `consultation_failed` |
| CLI: consult, oneshot, explore, batch submit/get | `cli.rs:891` | `consultation_failure_text` via `fail_consultation` |

The split (`consultation_failed` for synchronous tools that return a `CallToolResult` immediately; `consultation_failure_text` for async paths that store the rendered string in a `JobState::Failed` for later `job_get` collection, `render.rs:202`) is consistent. Every caller gets the same classified guidance. No caller bypasses the classifier.

---

## Summary

The heuristic is sound. The empty-answer marker (`"returned an empty answer"`) cannot be produced by an internal error and does not contain transient vocabulary in its canonical form. The one case where empty-answer text *can* carry transient vocabulary — the forced-turn failure wrapper — is intentional and tested: the transient classification wins (correctly). The `reached_a_model` gate correctly separates model-side from kaibo-side failures, and every `make_tools()?` site correctly lands in `Internal`. The transient vocabulary list is well-chosen; the few theoretically ambiguous entries (`"unavailable"`, `"try again"`) carry negligible operational risk because the calling agent always sees the full error detail alongside the guidance. All callers share the same classification path. The four guidance texts are clear, actionable, and appropriately scoped.

PR #120, opened 2026-08-03
feat: surface model output ceilings and size synth budgets from them

tobert:
A configure-prompt-running agent today cannot see model output ceilings anywhere: `list_models` renders context+pricing only, while the ceiling sits unnormalized in `DiscoveredModel.raw`. Meanwhile reasoning bills into the same completion budget as the answer, and the 2026-08-02 budget probes showed the default 16384 is tight for thinking-on synths (deepseek spent half a 10.6K completion on reasoning; or-kimi hit 97% of cap).

Four parts, all agreed with Amy in-session ("yeah definitely do 1. 2 as well. 3 as well, can be one pr for those" + "yes on the max tokens pins"):

1. **Normalize `output_ceiling` on `DiscoveredModel`** (OpenRouter `top_provider.max_completion_tokens`, Gemini `outputTokenLimit`; Anthropic/plain-OpenAI honestly `None`) and render it beside context in `render_models`, so both the MCP `list_models` face and `kaibo models` inherit it. The JSON envelope carries the field too.
2. **Configure prompt**: `CONFIGURE_STEPS_CORE` step 3 now tells the configuring agent to read the ceiling from the model listing and set the synth slot's `max_tokens` *from* it (not *to* it — a 128K ceiling is not a recommendation).
3. **Guide**: the `max_tokens` entry points at `kaibo models` / `list_models` as the ceiling source.
4. **Built-in synth pins from measurement**: deepseek and anthropic synths pin `max_tokens = 32768` (2× the floor; both thinking-on by default). Unmeasured casts stay on the `[defaults]` 16384 floor rather than guessing. A table-pinning test makes future edits deliberate.

**Behavior change worth a reviewer's eye**: with the pins, `[defaults] max_tokens` and `KAIBO_MAX_TOKENS` no longer flow into the built-in deepseek/anthropic *synth* slots — a slot pin wins, same as in a user cast. Two inheritance tests were repointed at the unpinned gemini synth, plus a new assertion that the pin beats a file-set default.

The anthropic 32768 pin has no catalog backing (Anthropic's `/v1/models` reports no ceiling); it rests on the DeepSeek measurement plus thinking-on billing, and is flagged in `docs/issues.md` for a live probe. The measured table now lives in the new "Synth output budgets" issues entry.

854 tests green (+3 new, failing-first), clippy silent, no new config knob so the template-coverage guard is untouched.

Patch written by a Claude subagent from the in-session brief; cross-family review to follow as a PR comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,15 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **Model listings now show each model's output ceiling.** `kaibo models` and the
+  `list_models` tool render the provider's advertised max completion tokens beside the
+  context window (OpenRouter's `top_provider.max_completion_tokens`, Gemini's
+  `outputTokenLimit`; providers that don't publish one show nothing), and the `--json` /
+  structured face carries it as `output_ceiling`. This is the number to size a synth
+  slot's `max_tokens` from — reasoning bills into the same completion budget as the
+  answer — and the `configure` prompt and the `max_tokens` entry in
+  `kaibo://config/guide` now both say so.
+
 - **The explorer can now hand whole files to whoever reads its report.** Inside
   `consult` and `deliberate`, the delegated investigator gets an `attach` tool: when a
   whole file is the evidence, it routes the file's real bytes (numbered, `cat -n`
@@ -59,6 +68,12 @@ record. Each later release appends a new section at the top.
   all three.
 
 ### Changed
+
+- **The built-in deepseek and anthropic casts give their synth more room to answer.**
+  Both pin the synth slot's `max_tokens` to 32768, twice the 16384 `[defaults]` floor,
+  because a measured consult showed deepseek-v4-pro spending half its completion budget
+  on reasoning before the answer even started. The `[defaults]` floor itself is
+  unchanged, and every other built-in cast stays on it until measured.
 
 - **kaibo's models now work from a role, not a job description.** Every preamble opens
   on who the model is on kaibo's team — "You are the synthesis agent", "You are the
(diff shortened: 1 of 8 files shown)


tobert:
## Cross-family review — DeepSeek (kaibo `consult`, holistic over the worktree, no diff)

**Verdict: ready to merge, no blockers.**

- **Parsing & rendering correct**: all three catalog paths verified (`and_then` chains bottom out at `None` on missing/mistyped fields); all four context/ceiling render combinations handled; CLI and MCP share `render_models`/`models_json_envelope` so the faces can't drift.
- **Pin precedence sound**: the 32768 pins win through the same `unwrap_or` every slot pin uses; the precedence chain (pin > file `[defaults]` > env) is covered by three tests, and the repointed inheritance tests document why they moved.
- **Text clean** for both audiences: the configure sentence is action-first with the rationale installed; the guide line is reference-voiced and scoped.
- **Tests non-vacuous**: the pin table is exhaustive; the ceiling-only render test asserts the *absence* of context.

Five minor notes, none blocking. Two folded in `19b6fb0`:
1. The configure prompt directed the agent to read a ceiling some providers never publish (Anthropic's `/v1/models` reports none) — now says to fall back to the provider's own model documentation.
2. A string-form slot override (`synth = "anthropic/other-model"`) replaces the slot whole and silently drops the built-in 32768 pin — pre-existing whole-slot semantics, but `max_tokens` is the first pin where the loss matters, so the merge test now asserts it as deliberate.

Left as-is (trivial per the review): no context-only render test (symmetric path), `list_models` tool description doesn't name the ceiling (the configure prompt does, and description budget headroom is finite), MCP-side prompt needle (CLI needle covers the shared core).

tokens: 2.03M in / 14.1K out / 1.96M cached · explorer deepseek-v4-flash · synth deepseek-v4-pro

PR #121, opened 2026-08-03
feat: image generation through a media CAS — the generate tool, stages 3-4

tobert:
Completes the media-cas arc on the four decisions Amy made in-session (2026-08-02/03):

1. **Trait shape approved, widened first**: `MediaOutcome::Complete` and `MediaPollOutcome::Complete` carry `Vec<MediaArtifact>` — Amy: "pretty sure many of the image gen models do multiple returns". One digest + one provenance sidecar per artifact; an empty list is a provider bug to refuse, never a blessed shape.
2. **Retrieval is operator-surface only**: the new `kaibo://cas/{digest}` resource serves bytes (base64, correct mime) to the calling client. The inner model team never sees the CAS — no kaish mount, no cast-facing read path — so one project's team can't enumerate another project's artifacts. Cross-agent sharing is deliberately deferred. Amy's framing: "cas should be a way for the kaibo agents to get stuff to the client efficiently."
3. **Lifecycle follows persistence**: disk CAS while persistence is active; in-memory with a severe startup warning when it isn't (artifacts don't survive restart); `[cas] enabled = false` turns it off and un-advertises dependent tools. One derivation (`Config::cas_mode`) feeds main, `kaibo://config`, and the CLI.
4. **Lane split by declared shape**: sync operations answer in-call; provider-declared deferred operations return `job-N` on the existing job verbs. The tool returns digests + URIs (+ real path in disk mode), never inline bytes.

The `generate` tool is the eighth capability flag, staffed via the cast's `image` slot through `CAST_ENUM_RULES`, with a third liveness gate: the CAS must be on. Stock installs (no image cast) drop the route — pinned in `tests/gating.rs`.

**Write-surface discipline held**: `tests/no_write_path.rs` pinned count is 4 before and after — `MemoryCas` touches no filesystem and the tool writes only through `Cas::put`'s existing blessed sites. No destination-path parameter exists anywhere; `src/sandbox.rs` is untouched.

Judgment calls flagged for review (details in commits):
- A HOME-less environment (no resolvable CAS dir) lands on memory mode with the severe warn naming the reason, rather than a fatal error.
- Deferred polling reuses `[defaults] call_deadline` as its budget (10s cadence) instead of a new knob.
- AGENTS.md's "produces no output artifacts" opening was rewritten — the change `docs/issues.md` explicitly required to land with the first artifact-producing tool.

888 tests green (baseline 868 after merging main), clippy silent. Patch by a Claude subagent from the in-session decision brief; cross-family reviews (DeepSeek + Gemini) to follow as PR comments.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,10 +5,12 @@ It augments a calling agent (Codex, Claude, Gemini, local agents, etc.) with a t
 models, lending one kind of help — *consultation*: grounded, cited, read-only answers
 about a codebase. The team
 *perceives* what fuses into its reasoning (image input today — `view_image` and image
-attachments on the model-driven tools; more modalities as the models gain them), but
-kaibo produces **no output artifacts** — it reasons over code, it doesn't render or
-emit. (When it needs to *record* something, that's a specific mediated tool, not a
-general write path; see the read-only invariant.)
+attachments on the model-driven tools; more modalities as the models gain them), and —
+when a cast carries an `image` slot — *produces*: the `generate` tool writes artifacts
+into kaibo's own media CAS (content-addressed, provenance-recorded, retrieved by the
+operator via `kaibo://cas/<digest>`), **never into the project**. Recording and emitting
+are always a specific mediated tool over kaibo's own store, not a general write path;
+see the read-only invariant.
 
 ## For agents working here
 
@@ -37,9 +39,11 @@ loop. Each consultation tool is that loop wearing different clothes:
 Both model-driven tools name their cast + answering model(s) in a provenance footer
 (`with_provenance` in `server.rs`), so a cross-model study sees which model answered.
 
-Seven `--no-<tool>` capability flags gate the surface (all on by default; `consult` also
-gates `consult_submit`, `batch` gates `batch_submit`, and the `job_*` verbs follow their
-live producers). A tool also needs a cast that can **staff** it, or its route is dropped;
+Eight `--no-<tool>` capability flags gate the surface (all on by default; `consult` also
+gates `consult_submit`, `batch` gates `batch_submit`, `generate` additionally needs the
+media CAS on, and the `job_*` verbs follow their live producers — a deferred `generate`
+mints a `job-N` like the other producers). A tool also needs a cast that can **staff**
+it, or its route is dropped;
 a server left with nothing advertised is refused at startup, by either road. See
 `CAST_ENUM_RULES` and `live_tools` in `server.rs`, and "Tool gating" in `docs/config.md`.
 Multi-provider over `rig-core`: a **`ProviderKind`** is the wire protocol (keyed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,9 @@ record. Each later release appends a new section at the top.
   when the store is on disk. Provider-native options (`aspect_ratio`,
   `output_format`, `seed`, `negative_prompt`, ...) pass through a `fields` object
   verbatim. Operations the provider declares deferred return a `job-N` handle on the
-  existing `job_wait`/`job_get`/`job_list`/`job_cancel` verbs. The tool follows the
+  existing `job_wait`/`job_get`/`job_list`/`job_cancel` verbs (the lane ships
+  offline-tested; every Stability operation wired today is synchronous). The tool
+  follows the
   staffing discipline: no configured cast with an `image` slot (a stock install)
   means it is not advertised and costs nothing; `--no-generate` /
   `KAIBO_NO_GENERATE` / `[server.tools] generate = false` switch it off explicitly.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,36 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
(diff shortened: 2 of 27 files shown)


tobert:
## Cross-family review 1 — DeepSeek (kaibo `consult`, holistic over the worktree)

**Verdict: ready to merge.** All nine review axes clean, with hand-verified evidence:

- The three blessed CAS write lines each carry their own marker; the `no_write_path` pinned count (4) has proven teeth (unmarked writes in cas.rs still fail; markers don't cross-pollinate between files).
- No destination-path parameter anywhere: `Cas::put` takes bytes + a closed `Extension` enum + provenance; the object path derives from the content digest; `create_new` everywhere, no unlink/truncate/rename.
- The inner model team has zero CAS routes — verified module by module (explorer, prompts, attach, sandbox, view_image, engine).
- `Digest::from_hex` enforces exactly-64-lowercase-hex before any lookup; not-found / corrupt / unreadable stay distinct; get and dedup both re-verify whole-object SHA-256.
- `cas_mode` is a single derivation consumed by main, the config render, and the CLI; the `SingleProcessLocked` degrade cascades to memory correctly.
- Triple gate on `generate` (flag ∧ staffable ∧ CAS on) with distinct "you disabled it" vs "nothing can staff it" warnings.

Five minor test gaps flagged, none blocking (empty-artifact-list refusal, deferred deadline timeout, unknown-MIME refusal, MediaStore corrupt dispatch, one weak assertion).

tokens: 4.39M in / 15.4K out / 4.30M cached · explorer deepseek-v4-flash · synth deepseek-v4-pro

tobert:
## Cross-family review 2 — GPT (kaibo `consult` via or-gpt, holistic; Gemini 503'd, direct GPT rate-limited)

**Verdict: not ready to merge** — no invariant breach found, but seven findings. I hand-verified the three biggest against source; all real. A fix batch is in progress on the branch:

**Confirmed bugs:**
1. `job_list` renders the in-memory jobs section only when the raw consult/deliberate flags are on — a generate-only server mints a `job-N` that `job_list` won't show. The shared render also calls every job a "Consultation". (Same class as the #110 producer-liveness fix.)
2. Provenance can lie: `fields.prompt` / `fields.model` override the effective request while the sidecar records the outer prompt and configured model. Fix: reserved keys, loud rejection.
3. The disk cap admission check doesn't count the provenance sidecar written after it.
4. Concurrent identical disk puts: the `AlreadyExists` loser returns success without verification while the winner may still be mid-write (handler clones share one `Arc<MediaStore>`).
5. Partial multi-artifact failure stores artifact 1, then errors without returning its digest — prevalidate the vector; name already-stored digests on mid-loop failure.
6. `Cas::open` is a pure containment check, but the guide claims a CAS dir that "cannot open" fails startup — a file at the root passes startup and fails on first paid generation.
7. Docs coherence: the embedded guide/template still say "there are no output or production roles" and omit `stability` from the kind table (the name-based template guard can't catch semantic contradictions); README has no `generate` entry; the changelog reads as though deferred Stability generation is live when no deferred operation is wired yet.

Also flagged: the MCP `fields` input (BTreeMap, unique sorted keys) can't express the `MediaRequest.fields` documented contract (ordered, repeated-name override) — reconcile before the API freezes.

**What it confirmed solid:** no steerable destination path, the pinned write carve-out, the model team's lack of any CAS route, digest validation, and the three-state lifecycle coherence.

tokens: 8.71M in / 47.8K out / 8.18M cached · explorer ~openai/gpt-mini-latest · synth ~openai/gpt-latest

tobert:
## Fix batch for the GPT review — all seven findings addressed (`c130700`, `ae92ea7`)

No finding was disputed; all six confirmed bugs checked out against source. Highlights:

1. **job_list**: new `job_producer_live()` predicate (consult ∥ deliberate ∥ generate-route-liveness) is the single source behind both the listing section and the `job-N` collect guard; all job wording is now producer-neutral. Failing-first test: a generate-only server lists, collects, and cancels its deferred job.
2. **Provenance**: `fields.prompt` / `fields.model` rejected as reserved keys, error naming the right parameter.
3. **Cap**: sidecar serialized before admission and counted; exact-boundary test measures real directory size.
4. **Concurrency**: disk `put` serialized through a shared mutex (handler clones share one store); an `AlreadyExists` loser now reads back and verifies. Honest caveat: the true race isn't deterministically reproducible — the failing-first proof is a dangling-symlink test that reproduces the post-pre-check `EEXIST` shape (old code returned Ok with nothing readable); a 32-round two-thread test pins the serialized contract.
5. **Partial multi-artifact**: whole-vector MIME prevalidation before any write; a genuinely mid-loop failure names the already-stored digests as retrievable. Deferred lane now tested with two artifacts.
6. **`Cas::open`**: structural `NotADirectory` refusal (file at root or in ancestry) at startup, still zero side effects; guide wording narrowed to match.
7. **Docs**: guide/template/README caught up with the media spine (wire split, third role, stability tables, `generate` entry); deferred claims qualified everywhere ("the lane ships offline-tested; every Stability route today answers in-call"); `MediaRequest.fields` contract rewritten to what the MCP face can express.

Plus DeepSeek's two test gaps (deadline timeout naming the provider id; empty-vector refusal). 898 tests green (+10), clippy silent, `no_write_path` pinned count unchanged at 4.

PR #122, opened 2026-08-03
feat(media): openai-images — one kind for hosted gpt-image and a local sd-server

tobert:
The second media kind on the #121 spine, approved by Amy in-session ("let's get the openai-images PR rolling"). One kind covers two real targets because they speak the same endpoint: hosted OpenAI image generation (gpt-image-1 family) and a local stable-diffusion.cpp sd-server (`/v1/images/generations`).

Design decided in-session:

- **`ProviderKind::OpenAiImages`** (canonical `openai-images`), `class()` → `Media(MediaKind::OpenAiImages)`. The #121 class-funnel claim held structurally: after adding the variant, the only non-exhaustive match in the tree was `MediaArm::from_slot` — no completion path needed an arm.
- **Key posture mirrors the `openai` completion kind, not Stability**: `OPENAI_API_KEY` / `~/.openai-key`, `key_optional` seeds true (the keyless local sd-server case); hosted use sets `key_optional = false` + the key file explicitly, exactly like `[backends.gpt]`. `base_url` optional, defaulting to hosted; a host root — the client appends its own path.
- **Sync-only**: `generate` returns `Complete(Vec<MediaArtifact>)` — `n` makes multi-artifact the normal case, the first real exerciser of #121's Vec widening. `response_format = "b64_json"` is seeded for URL-defaulting families and omitted for gpt-image (which rejects the parameter and always returns b64); URL-only responses are refused loudly. `poll` bails — unreachable by construction, said in its doc.
- **Mime** from the request's `output_format` (png|jpeg|webp, default png) into the CAS's closed `Extension`; unknown format errors before the paid call. No seed on this wire → `None`.
- **Config validation generalized to class-based**: image-slot-on-wire and reasoning-slot-on-media both bail naming the kind's class, so the error text can't go stale as kinds accumulate (the docs-drift lesson from #121's review).

Agent judgment calls: scalar typing in `fields` (the Images API type-checks its body — `n` must be a JSON integer; bare numbers/bools go typed, documented as wire-shape translation, not an allowlist); the key file matches the `openai` kind's `.openai-key` so both kinds share one credential.

920 tests green (+22 offline, fixture-driven; +1 `#[ignore]`d hosted live probe mirroring `stability_live.rs`), clippy silent, `cargo tree -i aws-lc-rs` empty (client builds through `src/tls.rs`), `no_write_path` untouched. Mutation checks proved the b64-refusal and class-validation tests can fail.

Docs: guide kind/base_url/staffing tables, template with both hosted and keyless-local shapes, README, short CHANGELOG entry.

Patch by a Claude subagent from the in-session design brief; cross-family reviews to follow as PR comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,12 @@ record. Each later release appends a new section at the top.
 - **A second media kind: `openai-images`.** One backend kind covers hosted OpenAI
   image generation (gpt-image-1) *and* a local stable-diffusion.cpp `sd-server`,
   which speaks the same `/v1/images/generations` shape — `base_url` picks the
-  target (unset dials hosted OpenAI), keys ride the same `OPENAI_API_KEY` /
-  `~/.openai-key` sources as the `openai` kind, and the keyless local case works
-  with no credential at all. A single `generate` call can now return several
-  images (the `n` field), each stored under its own digest.
+  target (unset dials hosted OpenAI, so the key is required by default), keys ride
+  the same `OPENAI_API_KEY` / `~/.openai-key` sources as the `openai` kind, and a
+  keyless local server opts in with `key_optional = true`. A single `generate`
+  call can now return several images (the `n` field), each stored under its own
+  digest — and `fields` values keep their JSON types (`n` as a number, `user` as
+  a string) all the way to the provider, on every media kind.
 
 - **kaibo can now generate images.** A new `generate` tool turns a text prompt into
   artifacts through the cast's new `image` slot — a media backend (Stability's v2beta
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,14 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **A second media kind: `openai-images`.** One backend kind covers hosted OpenAI
+  image generation (gpt-image-1) *and* a local stable-diffusion.cpp `sd-server`,
+  which speaks the same `/v1/images/generations` shape — `base_url` picks the
+  target (unset dials hosted OpenAI), keys ride the same `OPENAI_API_KEY` /
+  `~/.openai-key` sources as the `openai` kind, and the keyless local case works
+  with no credential at all. A single `generate` call can now return several
+  images (the `n` field), each stored under its own digest.
+
 - **kaibo can now generate images.** A new `generate` tool turns a text prompt into
   artifacts through the cast's new `image` slot — a media backend (Stability's v2beta
   `core`/`ultra`/SD3.5 family) riding beside the reasoning slots, e.g.
--- a/README.md
+++ b/README.md
@@ -475,11 +475,12 @@ the box with environment variables and built-in defaults, so a missing config fi
 not an error. `$XDG_CONFIG_HOME/kaibo/config.toml` lets you wire your own roster. The
 config has three concepts for configuring models:
 
-- **backend** — a *connection*: which wire protocol (`anthropic` | `deepseek` |
-  `gemini` | `openrouter` | `openai`), base URL, and where its key comes from. Secrets
-  never live in the TOML — only the *name* of an env var or the path to a key file.
-  `openrouter` is a keyed gateway with a fixed endpoint — one key reaching every major
-  model family, reasoning on by default via its unified `effort` param.
+- **backend** — a *connection*: which wire protocol (the completion kinds `anthropic`
+  | `deepseek` | `gemini` | `openrouter` | `openai`, or the media kinds `stability` |
+  `openai-images` behind the `image` role), base URL, and where its key comes from.
+  Secrets never live in the TOML — only the *name* of an env var or the path to a key
+  file. `openrouter` is a keyed gateway with a fixed endpoint — one key reaching every
+  major model family, reasoning on by default via its unified `effort` param.
 - **role** — a *job* a model does: `explorer` (fast sweeps), `synth` (the voice that
   answers), and `image` (the media member behind `generate`). A reasoning slot that
   reads images carries a `vision` pin (see [`docs/casts.md`](docs/casts.md)).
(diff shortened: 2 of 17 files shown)


tobert:
## Cross-family review 1 — DeepSeek (kaibo `consult`, holistic over the worktree)

**Verdict: not ready as submitted — one merge blocker, everything else solid.** Fixed in `76196fb`.

**The blocker (real, now fixed):** `kaibo://config` rendered the backend `kind` via `format!("{:?}", kind).to_lowercase()` — every prior kind survived by being single-word, but `OpenAiImages` became `openaiimages`, a spelling that round-trips into a load error if copied. Now `canonical_name()`, with a failing-first test pinning `kind = "openai-images"`.

**What it verified clean:**
- Request/response handling against the real API: the per-family `response_format` split (seeded for URL-defaulting families, omitted for gpt-image which rejects it), URL-only refusal with exact error equality in tests, `n>1` order preservation, empty-data and bad-base64 rejection.
- The scalar-typing heuristic is sound for this API's parameter set — `size = "1024x1024"` can't parse as a bare scalar so it stays a string; a float `n` is faithfully passed and rejected by the API rather than silently corrected.
- Class-based validation covers all four class-mismatch cases with no stale kind lists — a new media kind is covered the moment it's classified.
- The `key_optional = true` posture (placeholder bearer → 401 mid-call on a misconfigured hosted backend) is the same documented posture as the `openai` completion kind, with the guide and template both saying to set `key_optional = false` for hosted use.
- TLS through `src/tls.rs`; no aws-lc/openssl in the tree. No new write path; mime maps into the existing closed `Extension` enum; the tool remains operator-surface only.

**Non-blocking notes taken:** `resolved_base_url`'s doc now states it's completion-wire only (its localhost fallback is the wrong default for a media backend — `MediaArm::from_slot` resolves those to hosted). Left for follow-up: a boundary test on the bare `"gpt-image"` prefix match and `field_value` edge cases (`"true"`, `"null"`, empty).

921 tests green after the fold, clippy silent.

tokens: 3.14M in / 16.4K out / 3.07M cached · explorer deepseek-v4-flash · synth deepseek-v4-pro

tobert:
## Cross-family review 2 — GPT (kaibo `consult` via or-gpt; Gemini still 503ing)

**Verdict: not ready to merge** — invariants all confirmed intact (write path, TLS, CAS isolation), but four must-change items. All verified real; a fix batch is in progress on the branch:

1. **The scalar-typing heuristic mistypes legitimate strings**: the Images API's `user` field is a string — `user = "123"` becomes JSON number `123` and a provider 400. Fix: carry caller intent end to end — the tool's `fields` accepts string|number|boolean, `MediaRequest` carries a typed `FieldValue`, openai-images serializes it verbatim (no guessing), stability stringifies losslessly.
2. **`key_optional = true` is unsafe for a hosted-default kind** — the sharpest catch of the review. The analogy to the `openai` completion kind fails at the endpoint: that kind's keyless default is coherent because its *default endpoint is localhost*; `openai-images` defaults to hosted, so a minimal stanza loads clean, advertises `generate`, then sends `Bearer no-auth` to api.openai.com and 401s on the first paid call. This design decision is reversed: the kind seeds key-required; local keyless sd-servers say `key_optional = true` explicitly.
3. **Empty `b64_json` decodes to a zero-byte artifact** (empty string is valid base64); plus `" PNG "` passes local validation but reaches the provider unnormalized, and a caller-supplied `response_format` can knowingly buy URL artifacts we then refuse. All three hardened.
4. **Stale operator-facing text**: the `list_models` refusal names Stability's routes for any media kind; `src/lib.rs` still claims kaibo has "no write path of any kind" (stale since #121); `media.rs` says Stability is the sole implementation; `kaibo://tools`, the template's first `image`-slot mention, and README's kind list are all Stability-only.

Plus a real test gap: nothing offline drives `OpenAiImagesClient::generate` itself — a mock-transport test is being added (path, bearer, body types, per-family `response_format`, multi-image propagation).

What it confirmed solid: the per-family b64 split implementation and its tests, ordered multi-image parsing, the class-based validation core, `tls::https_client` usage with a clean lockfile, the pinned write carve-out, and the operator-only CAS surface.

tokens: 12.09M in / 91.3K out / 10.89M cached · explorer ~openai/gpt-mini-latest · synth ~openai/gpt-latest

tobert:
## Fix batch for the GPT review — all five items addressed (`bcc3b6b`), no disputes

1. **Typed fields end to end**: `media::FieldValue` (Str/Num/Bool) rides `MediaRequest`; the tool's `fields` schema accepts string | number | boolean; openai-images serializes verbatim (the value-guessing `field_value` is deleted, not patched), stability stringifies losslessly. `n = 2` stays the integer `2`; `user: "123"` stays a string.
2. **Key posture reversed**: `openai-images` seeds `key_optional = false` — its default endpoint is hosted, so the old keyless seed meant `Bearer no-auth` to api.openai.com and a 401 on the first paid call. A keyless local sd-server now opts in explicitly beside its local `base_url`. Mutation check: reverting the posture fails exactly the new test.
3. **Response hardening**: empty `b64_json` refused (`EmptyArtifact { index }`); the parsed `OutputFormat`'s canonical lowercase reaches the wire (`" PNG "` → `png`); any caller-supplied `response_format` is refused before the call, naming the credits-then-refuse hazard.
4. **Stale-text sweep**: kind-aware `list_models` refusal (no more "core, ultra, sd3" for every media kind), `lib.rs`'s "no write path of any kind" claim (stale since #121) rewritten to the two-blessed-surfaces truth, `media.rs` names both live impls, `kaibo://tools` covers both kinds, template + README aligned, the png|jpeg|webp set stated where it was an ellipsis.
5. **Transport test**: a one-shot local capture server (mirroring `tests/llm_timeout.rs`'s precedent — rig's fake-HTTP seam doesn't apply to a non-rig client) asserts the exact path, bearer header, typed wire body, per-family `response_format` presence, ordered two-image collection, and a readable provider 401.

928 tests green (+8), clippy silent, `aws-lc-rs` absent, `no_write_path` pinned count unchanged at 4.

PR #125, opened 2026-08-05
feat(artifact): save_artifact — a one-way delivery tunnel from the model team to the caller

tobert:
An opt-in tool in the consult driver's inner toolset: the model hands kaibo bytes, gets back a CAS digest, and the caller's answer footer names each `kaibo://cas/<digest>` to read — or ignore. One way by construction: the model can write into the store and can never read, list, or probe it.

## Decisions (from the design sessions, 2026-08-04/05)

- **One input path: inline `content`. There is no `from`/path parameter, permanently.** A real-filesystem `from:` was designed and dropped after the hard shake: it could only reach existing project files (a retention risk with no value the client doesn't already have), not the model-authored bulk the tool exists for. The schema is pinned to exactly `{content, format, label}` by a test.
- **Two-key opt-in + CAS liveness.** Operator: `[artifacts] enabled = false` default (kaibo's first inverted capability — the only surface where a *model* decides bytes become durable), `KAIBO_ARTIFACTS_ENABLED`, `--allow-save-artifact`. Client: `save_artifacts: true` per call on `consult`/`consult_submit`. A caller that asks and can't be served gets a loud `invalid_params` naming the missing key — never a quiet consult that swallowed the bulk it asked to have stored.
- **Driver loop only** — delegated explorer sweeps never carry the tool (pinned end-to-end on a real nested sweep).
- **Caps refuse, never truncate** (1 MiB/artifact, 8 artifacts, 8 MiB/call); each refusal names the cap, the actual size, and the recovery, and stores nothing. Caps render into the tool description from live values.
- **Dedup is unobservable.** Same content in, same response out — new-vs-already-present is an existence oracle over every project this kaibo has served, so it is never revealed.
- **Authorship provenance.** Sidecars now distinguish model-authored (`tool: save_artifact`, cast, model, slot, session, the model's label) from provider renders (`generate` sidecars now carry `tool: generate`, additively, old sidecars still deserialize). Someone downstream may *execute* an artifact's contents; the sidecar is the trust record.
- **The delivery framing is probe-validated**: on DeepSeek, 2/2 runs with the framing kept bulk out of the answer (URI + summary + representative entries); 2/2 controls re-dumped the entire corpus. The validated paragraph ships in the tool description. Local-model leg pending (zorak down).

## Prerequisite (first commit)

`cas.rs`: the `<hex>.json` sidecar's `mime` is now the authority for locating and typing an object (one stat), with the extension probe surviving only as a fallback for objects missing a sidecar. This also fixed a live mislabel bug: content stored under two container names resolved by probe order. `Extension` grew `Txt`/`Jsonl`; `generate` now prevalidates on `Extension::is_image`, so widening the enum did not widen what the media lane accepts.

## Security verification

- `tests/no_write_path.rs` pinned blessed-write count unchanged at **4** — artifacts reach disk only through the existing `Cas::put`.
- `src/sandbox.rs`: zero-byte diff. No `cas`/`artifact` reference reachable from kaish.
- Guards proven by mutation (each mutation → named test fails → restored → green): operator-key bypass, sweep-toolset injection, truncate-instead-of-refuse, dedup-oracle leak, missing sidecar authorship.
- 928 → 959 tests, 0 failed; clippy `--all-targets` silent; `cargo tree -i` empty for `aws-lc-rs`, `mimalloc`, `openssl-sys`.

## Known non-blocking notes

- The per-call total cap (8 MiB) cannot fire with the shipped values (8 × 1 MiB = exactly 8 MiB, refusal on exceed only) — pinned by a test whose failure message explains; it's a backstop against future cap changes, not live enforcement today.
- CAS retention changes character with cheap minting; tracked in `docs/issues.md` (operator ls/du verbs, a GC story someday).

## Attribution

First-pass implementation by a Claude Opus subagent; design hard-shake, bulk-suppression probe, and verification by Claude Fable 5 working with Amy. Cross-family reviews (DeepSeek + GPT) to follow as PR comments before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,12 +22,12 @@ record. Each later release appends a new section at the top.
   write a generated corpus, a long inventory, or a fixture into kaibo's media store;
   the answer names each `kaibo://cas/<digest>` and you choose what to read. Off
   unless an operator turns it on — `[artifacts] enabled = true`,
-  `KAIBO_ARTIFACTS_ENABLED`, or `--allow-save-artifact` — which makes it kaibo's
-  first opt-in capability rather than a `--no-<tool>` gate. Limits are fixed (1 MiB
-  per artifact, 8 artifacts and 8 MiB per call) and a save past one is refused
-  rather than truncated, naming the limit it hit. Every saved artifact's provenance
-  sidecar records that a model wrote it, which cast and model, and the model's own
-  description, because the contents may be run.
+  `KAIBO_ARTIFACTS_ENABLED`, or `--allow-save-artifact` (serve only) — which makes it
+  kaibo's first opt-in capability rather than a `--no-<tool>` gate. Limits are fixed
+  (1 MiB per artifact, 8 artifacts and 8 MiB per call) and a save past one is refused
+  rather than truncated, naming the limit it hit. A consult that saved and then failed
+  still reports the digests, and a call with a `session_id` keeps them beside the
+  conversation.
 
 - **A second media kind: `openai-images`.** One backend kind covers hosted OpenAI
   image generation (gpt-image-1) *and* a local stable-diffusion.cpp `sd-server`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,18 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **A consult can hand you bulk output as an artifact instead of spending your
+  context on it.** Ask with `save_artifacts: true` and the investigating model can
+  write a generated corpus, a long inventory, or a fixture into kaibo's media store;
+  the answer names each `kaibo://cas/<digest>` and you choose what to read. Off
+  unless an operator turns it on — `[artifacts] enabled = true`,
+  `KAIBO_ARTIFACTS_ENABLED`, or `--allow-save-artifact` — which makes it kaibo's
+  first opt-in capability rather than a `--no-<tool>` gate. Limits are fixed (1 MiB
+  per artifact, 8 artifacts and 8 MiB per call) and a save past one is refused
+  rather than truncated, naming the limit it hit. Every saved artifact's provenance
+  sidecar records that a model wrote it, which cast and model, and the model's own
+  description, because the contents may be run.
+
 - **A second media kind: `openai-images`.** One backend kind covers hosted OpenAI
   image generation (gpt-image-1) *and* a local stable-diffusion.cpp `sd-server`,
   which speaks the same `/v1/images/generations` shape — `base_url` picks the
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -311,6 +311,30 @@ scratch_limit_bytes = 67108864
 #                                          # 0 is a load error, not "refuse everything".
 
 # ---------------------------------------------------------------------------
+# [artifacts] — may the model team save what it writes? OFF by default.
+#
+# kaibo's one inverted switch. Every knob in [tools] is on until you turn it off; this
+# one is off until you turn it on, because it is the only surface where a MODEL decides
+# that bytes become durable. Turning it on grants standing permission, not automatic
+# behavior: the call must also pass `save_artifacts`, and [cas] above must be live. All
+# three, or the model never sees the tool.
+#
+# What it buys: a consult producing bulk output (a generated corpus, a long inventory, a
+# fixture) writes it into the CAS and names the digest in its answer, instead of
+# spending your context window on material you may only want to keep. You read it at
+# kaibo://cas/<digest>, or you do not.
+#
+# Limits are fixed, not configurable: 1 MiB per artifact, 8 artifacts and 8 MiB per
+# call. A save past a limit is refused and stores nothing. Every saved artifact records
+# its author in the provenance sidecar (tool, cast, model, slot, the model's own label),
+# because the contents are model-written and something downstream may run them.
+# Env: KAIBO_ARTIFACTS_ENABLED. CLI: --allow-save-artifact (serve only).
+# ---------------------------------------------------------------------------
+
+# [artifacts]
+# enabled = false                          # true lets a consult that ASKS save artifacts
+
+# ---------------------------------------------------------------------------
 # [backends.<name>] — CONNECTIONS only: `kind` picks the wire and its client. The
 # completion wires (anthropic | deepseek | gemini | openrouter | openai) serve the
 # reasoning slots; the media kinds (stability | openai-images) serve `image` slots
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -289,9 +289,11 @@ scratch_limit_bytes = 67108864
 # not every operation has one). A backup tool should carry them.
 #
 # NO GC, on purpose. kaibo never prunes this store and never rewrites an object — the
-# address IS the content hash, so an "edit" is a new object. Every artifact gets a
-# <hex>.json sidecar (prompt, model, cast, timestamp, mime, seed) precisely so YOU can
-# prune intelligently: `find -mtime`, jq over the sidecars, restic. A TTL may come later.
+# address IS the content hash, so an "edit" is a new object and an address you hold never
+# goes stale. Every artifact gets a <hex>.json sidecar (prompt, model, cast, timestamp,
+# mime, seed, producing tool) beside it, so an object reached BY ITS ADDRESS describes
+# itself. That is the access pattern: by digest, one lookup. Sweeping the store to survey
+# it does not scale and is not what this is for.
 #
 # max_bytes is OPT-IN, and unset is not the same as a huge number. Enforcing a ceiling
 # means summing every file in the store, on the write path, for EVERY new object — over
@@ -324,11 +326,14 @@ scratch_limit_bytes = 67108864
 # spending your context window on material you may only want to keep. You read it at
 # kaibo://cas/<digest>, or you do not.
 #
-# Limits are fixed, not configurable: 1 MiB per artifact, 8 artifacts and 8 MiB per
-# call. A save past a limit is refused and stores nothing. Every saved artifact records
-# its author in the provenance sidecar (tool, cast, model, slot, the model's own label),
-# because the contents are model-written and something downstream may run them.
-# Env: KAIBO_ARTIFACTS_ENABLED. CLI: --allow-save-artifact (serve only).
+# Limits are fixed, not configurable: 1 MiB per artifact, 8 artifacts and 8 MiB per call,
+# and a one-line label of at most 200 bytes. A save past a limit is refused and stores
+# nothing. The answer names every artifact it saved; a call with a session_id also
+# persists those digests beside the conversation.
+#
+# Env: KAIBO_ARTIFACTS_ENABLED (accepts only 1/true/yes and 0/false/no — this flag will
+# not guess, so `off` or a typo is a startup error rather than a silent enable).
+# CLI: --allow-save-artifact (serve only).
 # ---------------------------------------------------------------------------
 
 # [artifacts]
--- a/docs/config.md
+++ b/docs/config.md
@@ -928,10 +928,13 @@ see the server's posture before asking.
 | artifacts per MCP call | 8 |
 | bytes per MCP call | 8 MiB |
 | label | one line, 200 bytes |
-| formats | `text`, `jsonl` |
+| formats | `text`, `jsonl`, `markdown` — a hint; anything else stores as text |
 
 A save past a limit is **refused and stores nothing**; the refusal names the limit, the
-actual size, and the way forward. Nothing is ever truncated — a digest handed back for
+actual size, and the way forward. The format is the one non-gate: the content arrives as
+a JSON string, so it is UTF-8 text by construction — source code, a config, a log all
+ship as `text` — and an unknown format name stores as `text/plain` with the coercion
+stated in the tool result, never refused. Binary artifacts come only from `generate`. Nothing is ever truncated — a digest handed back for
 content that is not what the model wrote would be silent corruption. The per-artifact
 limit is a backstop rather than a working ceiling: the content rides in tool-call
 arguments, so the model's own `max_tokens` binds first. The label is bounded and must be a
--- a/docs/config.md
+++ b/docs/config.md
@@ -864,6 +864,71 @@ ancestors must be a directory. Write errors that only a real write can reveal
 opening the store writes nothing, so it cannot probe for them. Memory mode is the one
 warned degrade, mirroring the persistence posture it follows.
 
+## Saving artifacts: `[artifacts]`
+
+**Off by default.** This is the only switch in kaibo whose default is off. Everything in
+`[tools]` is advertised unless you disable it; this one is disabled unless you enable it,
+because it is the only surface where a *model* decides that bytes become durable.
+
+`save_artifact` is a tool injected into the `consult` driver's own toolset. The model
+hands kaibo bulk content it wrote and gets back a digest; the answer's footer names each
+`kaibo://cas/<digest>` and you decide whether to read it. The point is your context
+window: a consult asked to produce a generated corpus, a long inventory, or a fixture no
+longer has to spend the answer on material you may only want to store.
+
+```toml
+[artifacts]
+enabled = false                             # default false; true grants standing permission
+```
+
+CLI/env: `--allow-save-artifact` (serve only) / `KAIBO_ARTIFACTS_ENABLED`. Unlike the
+`KAIBO_NO_*` flags, the env var sets the knob either way and the CLI flag *enables* — the
+built-in default is the conservative end of the range, so a layer that could only disable
+would have nothing to say.
+
+**Three conditions, all required.** The tool is absent from the model's toolset unless
+every one holds:
+
+| condition | surface | who decides |
+|---|---|---|
+| `[artifacts] enabled = true` | config / env / CLI | the operator, standing |
+| `save_artifacts: true` on the call | the `consult` tool argument | the calling agent, per call |
+| the media CAS is live | `[cas] enabled`, above | the operator |
+
+A call that passes `save_artifacts` on a server missing either of the other two is
+**refused**, with a message naming which one — never answered quietly without the
+artifacts it asked for. `kaibo://config` reports `[artifacts] enabled`, so a caller can
+see the server's posture before asking.
+
+**Limits are fixed and not configurable.**
+
+| limit | value |
+|---|---|
+| bytes per artifact | 1 MiB |
+| artifacts per MCP call | 8 |
+| bytes per MCP call | 8 MiB |
+| formats | `text`, `jsonl` |
+
+A save past a limit is **refused and stores nothing**; the refusal names the limit, the
+actual size, and the way forward. Nothing is ever truncated — a digest handed back for
+content that is not what the model wrote would be silent corruption. The per-artifact
+limit is a backstop rather than a working ceiling: the content rides in tool-call
+arguments, so the model's own `max_tokens` binds first.
+
+**What the model cannot do.** It can write and it can never read. There is no list verb,
+no read verb, and no `kaibo://cas` access from inside the loop. The result of a save is a
+digest and nothing else: it never reports whether the content was already in the store,
+because that answer would let one project's model team probe another's artifacts. Only
+the `consult` driver loop gets the tool; delegated explorer sweeps never do.
+
+**Authorship is recorded.** Each saved artifact's `<hex>.json` sidecar carries
+`tool = "save_artifact"` plus the cast, the model, the slot, the session, and the model's
+own one-line label, alongside the fields a `generate` sidecar has. The contents are
+model-written and something downstream may run them, so the sidecar is the trust record.
+
+Retention is yours, as with everything in the CAS: kaibo never prunes, and this tool
+makes minting cheap. See the `[cas]` section above.
+
 ## House rules: `[context]`
 
 kaibo's models work for other agents, so they benefit from inheriting the calling agent's
--- a/docs/config.md
+++ b/docs/config.md
@@ -820,9 +820,11 @@ explicit. Every other open failure stays fatal.
 
 ## Media CAS: `[cas]`
 
-**On by default, and it follows persistence.** The CAS is the content-addressed store
-where an artifact-producing tool (image generation) writes what it made. Its lifecycle
-has three states, reported as `mode` in `kaibo://config`'s `[cas]` section:
+**On by default, and it follows persistence.** The CAS is the content-addressed store an
+artifact-producing tool writes into. Two produce today: `generate` (images a provider
+rendered) and `save_artifact` (bulk text a consult's model wrote, see `[artifacts]`
+below). Its lifecycle has three states, reported as `mode` in `kaibo://config`'s `[cas]`
+section:
 
 | mode | when | what it means |
 |---|---|---|
@@ -842,9 +844,24 @@ CLI/env: `--cas-dir` / `KAIBO_CAS_DIR` move the store; `--cas-max-bytes` /
 
 **The address is the content.** Every object's filename is the SHA-256 of its bytes, so
 nothing (model or operator) can aim a write at a chosen path, and identical content is
-stored once. Objects are written once and never rewritten, unlinked, or evicted. Each
-object gets a `<hex>.json` provenance sidecar (prompt, model, cast, timestamp, mime,
-seed) so you can prune by hand with `find`/`jq`/`restic` — kaibo ships no GC.
+stored once. Objects are written once and never rewritten, unlinked, or evicted — which
+is why kaibo ships no GC: an object at a given digest stays that object forever, so no
+address anyone holds goes stale.
+
+Each object gets a `<hex>.json` provenance sidecar (prompt, model, cast, timestamp, mime,
+seed, and which tool produced it). It makes an object self-describing **to whoever holds
+its address**: one lookup by digest says what the bytes are and what format to serve them
+as. Read it that way. The store is not built to be swept — a survey means walking 65,536
+shards and parsing a file per object, and it only gets slower as the store fills. Operator
+inventory verbs are tracked in `docs/issues.md` and want an index kaibo maintains, not a
+scan.
+
+The sidecar is **first-writer-wins**: it records the first write of a given content and is
+never rewritten. Save bytes the store already holds and the record stays the original
+one — which may name a different call, a different cast, or `generate`. It is metadata for
+an operator, not an audit trail, and a content-addressed store that never rewrites cannot
+be one. For a durable per-call record, kaibo's answers are the tool-call telemetry each
+save emits and the session the answer was recorded into.
 
 **Retrieval is operator-surface only.** A generation result returns each artifact's
 digest, a `kaibo://cas/<digest>` resource URI, and (in disk mode) the real filesystem
@@ -854,7 +871,10 @@ projects and a browsable CAS would let one project's team enumerate another's ar
 
 **`max_bytes` refuses, never evicts.** A write that would pass the cap fails loudly and
 nothing is deleted to make room. The cap is opt-in because enforcing it costs an
-O(objects) walk of the store per new write.
+O(objects) walk of the store per new write. When a cap is set, *every* write is admitted
+the same way, including one whose content the store already holds: exempting those would
+let a full store answer "is this content already here?" by succeeding for held bytes and
+refusing for new ones, across every project this kaibo has served.
 
 **Failure is loud.** In disk mode, a structurally unusable CAS path fails startup with
 an error naming the escape hatches: a dir that resolves inside an allowed project tree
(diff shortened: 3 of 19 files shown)


tobert:
## Cross-family review 1/2 — DeepSeek (invariants pass)

Holistic review over the branch worktree (whole files, no diff), cast `deepseek` via kaibo's own CLI consult. Verdict: **ready to merge.**

---

## Security & correctness review of `save_artifact` (PR #125)

### 1. Read-only — the model must never gain a write path to the project

**Holds.** `SaveArtifactArgs` (`src/artifact.rs:370–379`) has exactly three fields — `label`, `content`, `format` — and no path, destination, or source parameter of any kind. This is mechanically enforced by the test `the_schema_has_no_path_parameter_of_any_kind` (`src/artifact.rs:548–558`), which asserts the sorted key set is `["content", "format", "label"]` and nothing else.

The write reaches disk only through `self.store.put(bytes, ext, &provenance)` at `src/artifact.rs:334–337`, which dispatches to the existing blessed `Cas::put` (`src/cas.rs:689`) or `MemoryCas::put` (`src/cas.rs:813`). `src/artifact.rs` itself contains zero `std::fs` mutations — the module never touches the filesystem directly.

### 2. `tests/no_write_path.rs` stays pinned at exactly 4 blessed lines

**Holds.** `src/artifact.rs` introduces no `std::fs` call sites. All writes flow through the existing `MediaStore::put` path whose three blessed lines live in `src/cas.rs` (`create_dir_all` at line 733, `OpenOptions::new()...create_new(true).open(path)` at line 938, `file.write_all(bytes)` at line 944). The pinned constant is `EXPECTED_BLESSED_LINES = 4` (`tests/no_write_path.rs:94`) — one in `store.rs` plus three in `cas.rs`. The `blessed_marker_count_matches_exactly` test (`tests/no_write_path.rs:199–212`) would fail if a new blessed line appeared.

### 3. The write is unaimable (no path/destination parameter)

**Holds.** The tool's whole input is inline `content` (`src/artifact.rs:375`). The address is the content's own SHA-256, computed inside `Cas::put` by `Digest::of_bytes(bytes)` (`src/cas.rs:690`). The `format` field is an allowlist of two enum-like strings (`"text"`, `"jsonl"`) mapped onto the closed `Extension` enum via the `FORMATS` table (`src/artifact.rs:104`), not a free extension string. The test `the_schema_has_no_path_parameter_of_any_kind` is the structural backstop — adding a fourth parameter to the schema fails that test first.

### 4. kaish must never see the CAS

**Holds.** `src/sandbox.rs` is completely untouched — grep returns zero lines mentioning `cas`, `CAS`, or `artifact`. The model's shell has no builtin, no mount point, and no path to the CAS. The `kaibo://cas/<digest>` retrieval is handled entirely in the MCP resource route `read_cas_resource` (`src/server/mod.rs:1254–1281`), which is operator surface only — it validates the digest with `Digest::from_hex` before any lookup, and returns a blob to the calling MCP client, never into the model loop.

### 5. Dedup must be unobservable (no existence oracle over other projects' artifacts)

**Holds.** The tool's response is a pure function of the content. `SaveArtifact::call` (`src/artifact.rs:446–463`) always returns the same string for the same bytes — `"Saved. The caller will receive this artifact with your answer, at kaibo://cas/<hex>"` — with no "already present", "new", or "deduplicated" marker anywhere in the success path.

This is tested at two levels:
- **Ledger level**: `saving_identical_content_twice_yields_byte_identical_results` (`tests/save_artifact.rs:242–258`) — two saves of the same content produce `SavedArtifact` entries that are `assert_eq!` identical, including digest, mime, bytes, and label.
- **Tool-result-text level**: `the_tool_result_text_is_identical_for_identical_content` (`tests/save_artifact.rs:263–288`) — the tool's actual `call()` output is byte-identical, and explicitly asserts no "already", "dedup", or "new" substring.

In `Cas::put` (`src/cas.rs:702–718`), when the object path already exists, the existing bytes are read back and verified against the digest. If verification passes, the write is a no-op — but the return value is still `Ok(digest)`, identical to a fresh write. If verification *fails* (corrupt bytes at the hash address), it returns `Err(CasError::Corrupt)`, which `ArtifactSink::save` surfaces as `SaveError::Store(...)` — but this is a genuine error (disk corruption), not an existence oracle. For `MemoryCas::put` (`src/cas.rs:816–817`), the `contains_key` check returns `Ok(digest)` immediately — identical output.

### 6. Operator surface vs. model team must not blur

**Holds.** The model sees only the `save_artifact` tool and its response. The `ArtifactSink` (`src/artifact.rs:218–224`) has `saved()` and `footer()` methods that are consumed by the **server** after the consult loop completes — at `src/server/mod.rs:1734` (`with_artifacts` wraps the answer before the provenance footer) and at `src/server/mod.rs:1864` (same pattern for `consult_submit`). The model never accesses these methods. No kaibo-state builtin (`jobs`, `ps`, etc.) is added to kaish — the design explicitly follows the AGENTS.md rule that "model-facing kaish never grows `jobs`/`ps` or any kaibo-state builtin."

The `kaibo://cas/<digest>` resource URI that appears in the answer is for the *caller* (the operator's proxy), not the model. The model cannot resolve it — kaish has no network and no CAS mount.

### 7. Caps refuse rather than truncate; nothing is stored on refusal

**Holds.** The admission order in `ArtifactSink::save` (`src/artifact.rs:279–347`) is:

1. Format validation (`UnknownFormat` — line 293–299)
2. Per-artifact byte cap (`TooLarge` — line 301–306)
3. Per-call artifact count (`TooMany` — line 309–313)
4. Per-call byte total (`TotalExceeded` — line 314–320)
5. **Then** `store.put(...)` (line 334)

Every check runs before any write. The ledger lock (`src/artifact.rs:308`) spans admission and write, so two concurrent tool calls in one turn serialize correctly. On refusal, nothing is added to the ledger (the lock guard is dropped on the early return, and the `saved` vec is untouched). Every refusal message names the cap, the actual value, and the recovery action (`src/artifact.rs:170–207`).

Tests cover all refusal paths:
- Oversize with cap + size + "split" recovery (`tests/save_artifact.rs:51–74`)
- Exact-cap boundary stores whole or nothing (`tests/save_artifact.rs:81–93`)
- Too-many with cap + "follow-up call" recovery (`tests/save_artifact.rs:98–121`)
- Total-exceeded with cap + spending + "smaller selection" recovery (`tests/save_artifact.rs:129–159`)
- Unknown format naming available alternatives (`tests/save_artifact.rs:207–221`)
- Missing label (`tests/save_artifact.rs:226–230`)

### 8. Sidecar authority change preserves retrieval of pre-existing image artifacts

**Holds.** The lookup in `Cas::entry_for` (`src/cas.rs:590–604`) is a two-step ladder:

1. **Sidecar first**: read the provenance JSON, parse `mime`, map to `Extension` via `from_mime`, then check if the corresponding file exists. This is one `stat`, not N probes.
2. **Probe fallback**: if no sidecar or unmappable mime, walk `Extension::ALL` in declaration order and return the first that exists.

`Extension::ALL` (`src/cas.rs:306–313`) is `[Png, Jpeg, Webp, Gif, Txt, Jsonl]` — images are first, so an orphaned image object (no sidecar) resolves correctly. The `from_mime` function (`src/cas.rs:364–378`) was extended to accept `text/plain` and the JSON Lines aliases in addition to the original image mimes — the image mappings are unchanged.

Tests confirm:
- Sidecar as authority: `lookup_reports_the_extension_the_sidecar_recorded_not_the_probe_order` (`tests/cas.rs:663–684`) — stores webp-first, png-second at the same content hash; sidecar says webp and webp is returned.
- No-sidecar fallback: `an_object_with_no_sidecar_still_resolves_by_probe` (`tests/cas.rs:691–705`) — a bare `.jpeg` file with no sidecar is found as JPEG.
- Unmappable-mime fallback: `an_unmappable_sidecar_mime_falls_back_to_the_probe` (`tests/cas.rs:712–729`) — `application/octet-stream` in the sidecar doesn't hide the `.png` object.
- Backward-compatible deserialization: `a_sidecar_written_before_the_authorship_fields_still_deserializes` (`tests/save_artifact.rs:329–339`) — an old `{"prompt":"...","model":"...","cast":"...","timestamp":1,"mime":"image/png"}` sidecar loads with `tool: None, slot: None, label: None, session: None`.
- The `absent_authorship_fields_are_omitted_from_the_stored_json` test (`tests/save_artifact.rs:346–367`) proves `#[serde(skip_serializing_if = "Option::is_none")]` works — null-like fields don't pollute the stored JSON.
- Text round-trip: `text_and_jsonl_artifacts_round_trip_by_digest` (`tests/cas.rs:736–755`) proves the new formats store and retrieve correctly through the same digest-addressed pipeline as images.

### 9. The `generate` lane still refuses non-image mimes

**Holds.** `store_generated_artifacts` (`src/server/mod.rs:3972–3994`) prevalidates every mime through `Extension::from_mime(&artifact.mime).filter(Extension::is_image)`. Since `Txt` and `Jsonl` return `false` for `is_image()` (`src/cas.rs:347–352`), an image provider returning `text/plain` is still refused loudly. Growing `Extension` with text variants did not widen what `generate` accepts.

### 10. Two-key gating is correctly enforced

**Holds.** The gate is resolved in exactly one place — `KaiboHandler::artifact_sink` (`src/server/mod.rs:1548–1592`) — used by both `consult` (`src/server/mod.rs:1681`) and `consult_submit` (`src/server/mod.rs:1826`). Three conditions: operator `[artifacts] enabled` (`config.rs:1049`), caller's `save_artifacts: true` (`src/server/mod.rs:351`), and a live media CAS (`self.media_cas.clone()`). Missing any key returns a loud `McpError` naming which key and how to fix it; the test `save_artifact_needs_the_operator_key_the_call_key_and_a_live_cas` (`src/server/mod.rs:4616–4685`) covers all four combinations.

The sink, once built, is placed in `ConsultConfig.artifacts` (`src/consult/config.rs:135`), which is on the `ConsultConfig` rung — not `ExploreConfig`. This placement mechanically prevents a delegated sweep from ever receiving the tool, since sweeps are built from the explore rung (`run_explore_phase` in `src/consult/engine.rs`). The test `a_delegated_sweep_never_carries_save_artifact` (`src/consult/engine.rs:6440–6464`) proves this offline. The `deliberate` dossier stage shares the explore rung and is likewise out of reach.

### 11. Toolset wiring is correct

The driver's toolset is built in `consult_tools` (`src/consult/engine.rs:1883–1967`). `save_artifact` is injected at lines 1963–1964 only when `cfg.artifacts` is `Some`. The toolset is rebuilt for the forced final turn (turn-cap recovery), and the rebuild reads the same `cfg.artifacts` Arc — so the per-call caps correctly span the full call, not individual turns. The tests `save_artifact_is_in_the_driver_toolset_only_with_a_sink` (`src/consult/engine.rs:6401–6433`) and the offline e2e `a_driver_that_saves_an_artifact_reports_it_in_the_callers_footer` (`src/consult/engine.rs:6494–6542`) confirm this wiring.

### 12. No fragility found in the `Provenance` schema extension

The new fields `tool`, `slot`, `label`, `session` (`src/cas.rs:439–455`) are all `Option<String>` with `#[serde(default, skip_serializing_if = "Option::is_none")]`. This follows the rule in the module doc (`src/cas.rs:394–407`): "every new field must be `Option`." Old sidecars deserialize with `None` for these fields. The `generate` lane now writes `tool: Some("generate")` (`src/server/mod.rs:4013`) — a new write for generate sidecars, but fully backward-compatible.

### 13. One observation worth noting (not a bug)

The shipped caps relationship `MAX_ARTIFACTS_PER_CALL * MAX_ARTIFACT_BYTES == MAX_TOTAL_BYTES_PER_CALL` (8 × 1 MiB = 8 MiB) means the per-call total cap can never fire with the current numbers — the count cap binds first. This is explicitly documented in the test `the_shipped_caps_leave_the_call_total_as_a_backstop` (`tests/save_artifact.rs:168–177`) and in the refusal error at `src/artifact.rs:183–188`, which says "save a smaller selection." The total cap becomes live the moment either of the other two constants rises. This is a deliberate design documented in both the test message and the code, not an oversight.

---

## Verdict: **Ready to merge.**

Every invariant from AGENTS.md holds, and each is backed by a test that would fail if the property regressed. The implementation is narrow, mechanically enforced (no path parameter, closed enum for format, placement on `ConsultConfig` not `ExploreConfig`), and thoroughly tested at every layer — sink admission, tool result opacity, two-key gating, toolset wiring, sidecar backward compatibility, and the offline e2e.

———
tokens · 675844 in · 9446 out · 4408 reasoning · 607488 cached

tobert:
## Cross-family review 2/2 — GPT via or-gpt (operational seams): NOT ready, nine findings

Holistic review over the branch worktree, cast `or-gpt`. The top findings (capacity/dedup existence oracle, store-error information leak, label footer injection, discarded dedup provenance) were hand-verified against the code and are real. A fix batch is going back to the implementation agent; the two-family split (DeepSeek invariants-clean / GPT seams-not-ready) matches the pattern from #121 and #122.

---

# Review finding: not ready to merge

The core security shape is sound, but there are several blocking operational and information-leak defects. In particular, the current implementation does not make dedup unobservable on failure paths, does not reliably preserve the digest of bytes already written, and cannot truthfully record per-save authorship when content is deduplicated.

## Blocking findings

### 1. CAS failures expose shared-store state and filesystem paths to the inner model

`SaveError::Store` embeds the complete `CasError` string, and `map_error` deliberately sends that text to the model. Capacity errors disclose exact shared-CAS usage, while I/O errors include the operator’s CAS path. That crosses the stated operator/model-team boundary.

> `SaveError::Store(e) => ... "kaibo's artifact store refused this write: {e}"`  
> `ToolExecutionError::other(error.to_string()).with_source(error)`  
> — `src/artifact.rs:201-205`, `src/artifact.rs:387-392`

> `.map_err(|e| SaveError::Store(e.to_string()))?;`  
> — `src/artifact.rs:334-337`

> `"current usage {current_bytes} bytes"`  
> — `src/cas.rs:178-185`

> `"creating cas shard dir {}"` / `"writing cas object {}"` / `"writing cas provenance sidecar {}"`  
> — `src/cas.rs:733-760`

This conversion also destroys the typed `CasError`: the attached source is only a `SaveError` holding a string, not the original capacity/corruption/I/O error (`src/artifact.rs:148-166`, `src/artifact.rs:334-337`).

**Required change:** retain `CasError` as the source, but give `SaveError::Display` a model-safe message. Capacity should not reveal `current_bytes`; I/O should not reveal the CAS root or shard path. Operator logs can render the full source chain separately.

---

### 2. Dedup is observable when the store is full, contrary to the feature’s security claim

Both stores bypass the capacity check when the requested content is already present. Consequently, at capacity, a save succeeds if the guessed bytes already exist and fails if they do not. The model sees that success-versus-error distinction, making the shared CAS an existence oracle.

> Disk checks the cap only in the `else if` after `obj_path.is_file()`; existing content bypasses it.  
> — `src/cas.rs:702-725`

> Memory returns immediately for an existing digest, before checking `max_bytes`.  
> — `src/cas.rs:813-825`

Tests explicitly pin this exemption:

> “the dedup write of the held object still succeeds with no slack”  
> — `tests/cas.rs:565-578`

That conflicts with the feature’s declared boundary:

> “It never says whether the content was new or already present, because that difference is an existence oracle”  
> — `src/artifact.rs:34-39`

> “it never reports whether the content was already in the store”  
> — `docs/config.md:918-922`

**Required change:** make admission independent of prior existence on the model-facing path. A straightforward safe rule is to apply the would-be capacity admission to every `save_artifact` call, including dedup, so full-store behavior is identical. Add tests proving that present and absent content produce indistinguishable outcomes when the CAS lacks headroom.

---

### 3. One sidecar per digest cannot fulfill the promised per-save authorship record

On disk, a deduplicated write never rewrites the existing sidecar; in memory, a duplicate returns before updating provenance. Thus a save can report success while its model, cast, session, label, or even producer is absent from the permanent trust record.

> The sidecar is written with `create_new`, and the returned `false` for an existing file is ignored.  
> — `src/cas.rs:757-764`, `src/cas.rs:916-945`

> Memory dedup returns before storing the supplied provenance.  
> — `src/cas.rs:813-818`

This contradicts the user-facing promise:

> “Each saved artifact's `<hex>.json` sidecar carries `tool = "save_artifact"` plus the cast, the model, the slot, the session…”  
> — `docs/config.md:924-927`

> “Every saved artifact's provenance sidecar records that a model wrote it…”  
> — `CHANGELOG.md:28-30`

The conflict is especially important across producers: bytes first written by `generate` retain the image-side provenance even if a later model-authored save claims them, and vice versa. The code itself defines absent `tool` on old sidecars as meaning the historical `generate` producer (`src/cas.rs:435-455`).

**Required change:** decide and implement provenance multiplicity. If every save must be attributable, use append-only, content-addressed provenance/attestation records per save rather than one immutable sidecar per content digest. If provenance is intentionally “first insertion only,” the trust claims in code and docs must be weakened—but that would not satisfy this feature’s stated authorship purpose.

Add tests for:

- save after an existing `generate` object;
- two saves of identical content from different models/sessions/labels;
- disk and memory modes.

---

### 4. Same bytes under different formats produce contradictory footer, resource MIME, path, and provenance

The object path includes the requested extension, but the shared sidecar path is only `<digest>.json`. Therefore identical content can be written under several extensions while the first sidecar remains authoritative.

> Object path: `<digest>.<ext>`; sidecar path: `<digest>.json`.  
> — `src/cas.rs:546-554`

> `entry_for` trusts the sidecar MIME first.  
> — `src/cas.rs:577-603`

The existing test deliberately stores WebP first and PNG second and establishes that the first sidecar wins:

> “the sidecar recorded image/webp, so that is what this object IS”  
> — `tests/cas.rs:653-683`

But `ArtifactSink` records the requested extension in its ledger, not the authoritative extension that retrieval will use:

> `mime: ext.mime()`  
> — `src/artifact.rs:334-345`

The footer then prints that ledger MIME while deriving the disk path independently through `store.path_for`:

> `a.mime` is rendered, while `store.path_for(&d)` may resolve a different extension.  
> — `src/artifact.rs:492-507`

Thus a JSONL save of bytes already stored as text can say `application/jsonl` while `kaibo://cas/<digest>` serves `text/plain`, and its `path:` can end in `.txt`.

**Required change:** make `put` return an authoritative receipt containing at least the digest and effective extension/provenance identity, or reject/avoid conflicting container writes without exposing prior existence. Footer MIME, resource MIME, path, and provenance must all derive from that one receipt. Add a producer-level test; the current CAS-only test codifies first-sidecar authority without checking what the second caller reports.

---

### 5. Already-written artifacts are silently orphaned on later consult failure, cancellation, or job eviction

The sink records successful saves during the loop, but artifact footers are appended only on the successful outer result. A provider failure, empty-answer failure, deadline, or tool-loop failure after a successful save returns no digest footer.

> Sync failure returns immediately before `with_artifacts` is called.  
> — `src/server/mod.rs:1708-1724`, `src/server/mod.rs:1733-1742`

> Deferred failure similarly returns only `consultation_failure_text`; the footer is added only in the `Ok(out)` arm.  
> — `src/server/mod.rs:1849-1883`

The deferred lane additionally aborts the future on cancellation and on eviction of a running job, leaving no tail that can surface the sink’s ledger:

> `job.abort.abort(); *state = JobState::Canceled`  
> — `src/jobs.rs:258-273`

> Dropping an evicted job aborts its task.  
> — `src/jobs.rs:100-109`

There is no end-to-end `consult_submit → save_artifact → job_get` test; the repository only tests the shared gate and sync engine footer (`src/server/mod.rs:4606-4685`, `src/consult/engine.rs:6489-6547`).

**Required change:** preserve and render saved-artifact receipts on every terminal path after the first successful save, including sync failure, deferred failure, cancellation, and eviction—or explicitly prevent cancellation/eviction from orphaning already-written receipts. Add end-to-end tests for successful deferred collection and failure-after-save.

---

### 6. `Cas::put` can commit the object, fail the sidecar, and then claim “Nothing was saved”

The disk write order is object first, sidecar second:

> Object write at `src/cas.rs:738-755`; sidecar write at `src/cas.rs:757-762`.

If the sidecar write fails because of disk exhaustion, permissions, or an I/O fault, the object remains permanently stored. `ArtifactSink` does not enter it into the ledger because `put` returned `Err`, and its model-visible error says:

> “Nothing was saved.”  
> — `src/artifact.rs:201-205`

The CAS acknowledges this object-without-sidecar crash shape and resolves it by extension probing later:

> “an object with no sidecar still resolves by probe”  
> — `tests/cas.rs:686-704`

So this is recoverable storage, but the digest is lost from the current call.

**Required change:** represent partial commitment explicitly, carrying the computed digest and whether the object landed. The failure result/footer must give the caller any retrievable digest rather than claim nothing was stored. A two-file, no-rewrite store cannot make object-plus-sidecar atomic, so the API has to model this state.

---

### 7. Memory-mode capacity omits provenance, unlike disk mode

Disk admission counts content plus serialized provenance:

> `incoming = bytes.len() + sidecar_json.len()`  
> — `src/cas.rs:697-725`

Memory admission counts only content even though the stored `MemObject` also clones provenance:

> `current` sums only `o.bytes.len()`, and admission adds only `bytes.len()`.  
> — `src/cas.rs:819-825`

> The accepted object then stores `provenance: provenance.clone()`.  
> — `src/cas.rs:828-835`

This makes `max_bytes` mode-dependent and allows large prompts/labels to consume unaccounted memory. It conflicts with the stated “same refuse-never-evict posture” (`src/cas.rs:787-789`) and “soft cap on total media-CAS size” (`src/cli.rs:222-226`).

**Required change:** include the in-memory provenance footprint in admission, preferably using the same serialized representation as disk so configured behavior is predictable across modes.

---

### 8. `label` is documented as one short line but is neither bounded nor line-validated

Only blank labels are rejected. Newlines and arbitrarily long labels are accepted:

> `let label = label.trim(); if label.is_empty() ...`  
> — `src/artifact.rs:279-288`

The raw label is then inserted into the structured footer:

> `a.label`  
> — `src/artifact.rs:492-501`

This permits fake artifact entries or `path:` lines inside kaibo’s own footer and bypasses the documented content caps through unbounded sidecar/footer metadata. The only test checks blank labels (`tests/save_artifact.rs:223-230`).

**Required change:** enforce a modest byte limit and reject `\r`/`\n` before touching the store. Include the limit in the generated tool description and test refusal-before-write.

---

### 9. The opt-in environment switch treats typos and `off` as enabled

`KAIBO_ARTIFACTS_ENABLED` is false only for empty, `0`, `false`, or `no`; every other string is true. Therefore `off`, `flase`, and arbitrary garbage enable the first inverted capability.

> `!v.is_empty() && v != "0" && v != "false" && v != "no"`  
> — `src/config.rs:3258-3268`

Tests cover only the recognized values and do not test invalid input (`tests/config.rs:2355-2372`).

For ordinary `KAIBO_NO_*` narrowing flags this permissive convention is less dangerous, but it is surprising for a default-off permission that allows model-directed durable writes.

**Required change:** use a strict boolean parser for this opt-in key and reject unknown values at startup. Document the accepted spellings.

## Correct and well-designed portions

### The sync and deferred lanes share the same up-front three-way gate

Both require operator permission, the per-call `save_artifacts` key, and a live CAS; an explicit unsupported request fails rather than silently running without the tool.

> Central gate and refusal messages: `src/server/mod.rs:1536-1591`.  
> Sync construction: `src/server/mod.rs:1681-1688`.  
> Deferred pre-spawn construction: `src/server/mod.rs:1822-1833`.

The deferred job captures the same `ConsultConfig` and appends artifacts on success (`src/server/mod.rs:1849-1878`). The missing piece is failure/cancellation preservation and an end-to-end deferred regression test.

### Concurrent tool calls in one turn are serialized correctly within one server/store instance

The per-call ledger lock spans admission and `MediaStore::put`, so concurrent calls cannot both consume the same remaining budget:

> `let mut ledger = self.ledger.lock()` through the store write and ledger update.  
> — `src/artifact.rs:270-275`, `src/artifact.rs:308-345`

The disk CAS also serializes dedup, capacity accounting, object write, and sidecar write across clones of that `Cas`:

> — `src/cas.rs:683-695`

A concurrent near-cap `ArtifactSink` test should still be added; the existing concurrency test covers `Cas::put`, not the per-call ledger (`tests/cas.rs:846-890`). Separate kaibo processes opening the same CAS do not share this mutex, so the documented soft cap remains racy across processes (`src/cas.rs:463-473`, `docs/config.md:801-804`).

### Per-tool caps refuse before calling the store and never truncate

Format, label, per-artifact size, count, and total checks precede `MediaStore::put`:

> — `src/artifact.rs:285-320`, with the write only at `src/artifact.rs:334-337`.

Tests pin whole-content-or-nothing at the per-artifact boundary and no ledger entry on refusal (`tests/save_artifact.rs:47-93`, `tests/save_artifact.rs:95-159`). For a new disk object, configured CAS capacity also counts object plus sidecar before `create_dir_all` or either file write (`src/cas.rs:719-733`, `tests/cas.rs:781-811`).

That guarantee applies to validation/cap refusals, not arbitrary I/O failure after the object write; finding 6 needs the wording and API corrected.

### Normal footer rendering distinguishes disk and memory modes correctly

An unused sink leaves the answer byte-identical, and each used receipt includes URI, MIME, size, and label:

> — `src/artifact.rs:473-501`.

`path:` is appended only when `MediaStore::path_for` returns one; that method returns `Some` only for disk mode:

> — `src/artifact.rs:502-507`, `src/cas.rs:888-896`.

Tests cover no-footer and disk-path behavior (`tests/save_artifact.rs:371-407`). Add an explicit used-sink memory-mode negative test. The extension/provenance collision in finding 4 must also be resolved before these fields can be considered authoritative.

### Older sidecars remain structurally readable

The new authorship fields are optional/defaulted:

> — `src/cas.rs:394-455`.

Tests deserialize sidecars predating both `seed` and the authorship fields (`tests/cas.rs:31-47`, `tests/save_artifact.rs:326-339`). Missing, unreadable, malformed, or unmappable sidecars fall back to probing known extensions (`src/cas.rs:563-604`, `tests/cas.rs:686-729`).

Object corruption is loud: reads verify the digest, and same-extension dedup verifies before returning success (`src/cas.rs:626-632`, `src/cas.rs:702-718`, `tests/cas.rs:265-317`). The cross-extension gap remains finding 4.

### The narrow write/security boundary is preserved

The inner schema has only `label`, `content`, and `format`; no path-like parameter exists:

> — `src/artifact.rs:370-379`, `src/artifact.rs:423-443`.

A dedicated test pins those exact three properties:

> — `src/artifact.rs:545-557`.

`ArtifactSink` writes only through `MediaStore::put`, which dispatches to the existing `Cas::put`/`MemoryCas::put` seam:

> — `src/artifact.rs:334-337`, `src/cas.rs:864-870`.

Production filesystem mutations remain the existing three blessed CAS lines—shard creation, `create_new`, and `write_all`—plus the store’s one blessed line. The audit pins the tree-wide count at four:

> — `src/cas.rs:733-760`, `src/cas.rs:935-945`, `tests/no_write_path.rs:78-94`, `tests/no_write_path.rs:169-211`.

The CAS remains operator-only: retrieval is an MCP resource, while the driver toolset contains only write-side `save_artifact`; delegated sweeps have no sink (`src/server/mod.rs:109-114`, `src/server/mod.rs:3234-3239`, `src/consult/engine.rs:1957-1965`).

## Configuration and documentation assessment

The intended precedence is consistent: file is loaded first, env overlays it, and serve CLI overlays both (`src/config.rs:1182-1217`, `src/main.rs:83-100`). `[artifacts]` defaults false (`src/config.rs:1974-1978`); the CLI flag is enable-only (`src/config.rs:2057-2063`); the per-call key is an additional gate, not a precedence override (`src/server/mod.rs:1538-1547`).

The CLI distinction is intentional but easy to miss: `--allow-save-artifact` exists only on `serve`, and direct `kaibo consult` always sets `artifacts: None` because it has no per-call client key (`src/cli.rs:236-275`, `src/cli.rs:527-555`, `src/cli.rs:780-785`). The guide and template say “serve only” (`docs/config.md:884-887`, `docs/config.example.toml:327-335`), but the changelog lists the flag without that qualifier (`CHANGELOG.md:20-30`). Add the qualifier there.

The guide’s three-way gate, fixed limits, driver-only scope, and memory/disk lifecycle otherwise match the intended code (`docs/config.md:867-930`, `docs/config.example.toml:314-335`). Before merge, revise the claims made false by the findings above—especially dedup opacity, “every saved artifact” authorship, and “nothing was saved”—and update the Media CAS introduction, which still describes its producer only as image generation (`docs/config.md:821-825`).

## Verdict

**Not ready to merge.** Required first:

1. Sanitize model-visible CAS failures while retaining typed operator diagnostics.
2. Close the capacity/dedup existence oracle in both disk and memory modes.
3. Redesign provenance so every successful save has an immutable authorship record.
4. Return and render the authoritative extension/MIME for cross-format dedup.
5. Preserve saved digests on consult failure, deferred failure, cancellation, and eviction.
6. Represent object-written/sidecar-failed partial commits truthfully.
7. Count provenance toward memory-mode capacity.
8. Bound and line-validate labels.
9. Strictly parse `KAIBO_ARTIFACTS_ENABLED`.
10. Add concurrent-ledger and end-to-end `consult_submit` artifact tests, then correct the guide, template, and changelog claims accordingly.

———
kaibo · cast `or-gpt` · explorer `~openai/gpt-mini-latest` · synth `~openai/gpt-latest`
tokens · 5572737 in · 46704 out · 5103070 cached · 300402 cache-write

tobert:
## Fix batch for the or-gpt review: all nine findings fixed, none disputed

Commits `43ff2de` (fixes) + `53cbf5a` (docs framing). Re-verified independently: **976 tests / 0 failed** (was 959), clippy silent, `aws-lc-rs`/`mimalloc` trees empty, `no_write_path` pinned count still exactly **4** — no fix needed a new blessed write line.

Highlights, each proven failing-first or by mutation:

1. **Existence oracle closed (both stores):** capacity admission now runs before the dedup branch, unconditionally — a full store refuses held and absent content identically. The old dedup-exemption test is deliberately inverted, with the disclosure argument in its doc comment.
2. **Model-visible store errors sanitized:** `SaveError::Store` keeps the typed `CasError` for operator logs (`tracing::warn!`) but renders no paths and no usage numbers to the model.
3. **Labels bounded:** 200 bytes, control characters refused before the store touches anything — closes the footer-injection hole. Live value in the tool description.
4. **`KAIBO_ARTIFACTS_ENABLED` parses strictly:** exactly `1/true/yes` / `0/false/no`; `off`, typos, and garbage are loud startup errors.
5. **Footers survive failure paths** in both lanes via shared render functions; a consult that saved then failed still hands the caller its digests. Job cancellation orphaning is documented honestly (a `session_id` preserves the digests — see below).
6. **Object-landed/sidecar-failed is truthful:** new `CasError::ProvenanceNotRecorded` carries the digest; the model and footer both learn the artifact WAS saved.
7. **Memory-mode capacity counts provenance**, same serialized representation as disk.
8. **Cross-format dedup renders the store's authoritative mime/path** in the footer (the conflicting write is not refused — a refusal would reopen the oracle).
9. **Session-persisted answers now include the artifact footer** — digests land in the `turns` table beside the conversation, zero schema change (was recording the bare answer; proven failing-first).

Design direction folded in from Amy: the provenance sidecar is housekeeping metadata read **by address** (it is how lookup types an object in one stat), not an audit or trust record and not a scan target — docs and the `cas.rs` module doc no longer pitch `find`/`jq` sweeps, and the future `kaibo cas ls`/`du` idea is reframed around an index kaibo maintains rather than a walk of the store. The durable record story is tool-call telemetry plus session persistence. Dedup provenance is first-writer-wins, documented truthfully and pinned by tests (including save-after-`generate`).

One reported gap, accepted: no handler-level `consult_submit` e2e (the handler has no offline client seam); both lanes compose through the same two tested render functions instead, plus an engine test that artifacts saved before a provider failure survive it.

Reviews: DeepSeek (invariants) ready-to-merge; GPT via or-gpt not-ready with these nine, all verified real and now fixed. Implementation and fixes by a Claude Opus subagent; verification by Claude Fable 5.

PR #126, opened 2026-08-05
feat(cas): read_cas — deliberate, metadata-first artifact retrieval (replaces the kaibo://cas resource)

tobert:
The retrieval half of the artifact contract becomes a tool, and the `kaibo://cas/<digest>` MCP resource is deleted. Follow-up to #125, on Amy's direction.

## Why replace the resource

- **Resources are ambient; a tool call is deliberate.** Hosts treat resources as attachable context (some prefetch/auto-attach) — the wrong posture for model-authored bytes. A tool call is explicit arguments + a permission prompt + a traced span: retrieval becomes something the operator's agent *does*, not something that happens to its context.
- **`resources/read` has no negotiation** — whole-blob by construction. Measured on a real 3.8 MB PNG: ~5 MB of base64, kept out of context only by Claude Code spilling it to a side file. Host mercy, not protocol.
- One audited surface instead of two. The URI **string** survives as the artifact's stable name; only the serving route is gone.

## The tool

`read_cas(digest, offset?, length?)` — client-facing, gated on CAS liveness alone (not `[artifacts] enabled` — `generate` artifacts need retrieval too), never in the inner model toolsets (pinned).

- **Metadata leads every response**: digest, mime, total bytes, `binary:`, label (when the sidecar has one), range served, and — disk mode — the real filesystem path, so a client holding a shell goes direct (Amy's call).
- `length: 0` → metadata only. Omitted `length` → a bounded 64 KiB window, never the whole object; the metadata names the total so paging is deliberate.
- Text arrives as text, with UTF-8-aware window trimming so a byte offset never splits a character and every reported range is resumable. Binary arrives as base64 **only when explicitly asked** — a base64 dump is never something that happens to you.
- **Images ≤ 2 MiB with no range** return an MCP image content block — one hop to a vision model's eye, the same mechanism as the inner `view_image`. Larger: metadata + path.
- An over-ceiling `length` (> 1 MiB, deliberately the same figure as `MAX_ARTIFACT_BYTES`) is refused with the recovery, never clamped — a silent clamp would corrupt the caller's next offset.
- Ranges are computed after the store's whole-object digest verification — corruption stays loud on every path, including metadata-only reads. "Cheap" here means cheap in tokens.

## Notable judgment calls (agent's, verified)

- **`FOLLOWER_TOOL_NAMES` + `has_substantive_tools()`**: `read_cas` rides a store that is on by default, so counting it toward the empty-surface guard would leave that startup refusal unable to fire on any stock install. Retrieval and the `job_*` collect verbs are now "followers"; a server offering only followers still refuses to start. A deliberate startup-behavior change, test-pinned.
- New `MediaStore::provenance` read-only accessor so the label reaches the metadata.
- Sweep: footer + `generate` result lines + tool description + `kaibo://tools` + guide + template + README now point at `read_cas`; instruction-budget tests unaffected.

## Verification

1005 tests / 0 failed (main was 980), 13/14 planner tests proven red-first against a stubbed planner; clippy `--all-targets` silent; `aws-lc-rs`/`mimalloc` trees empty; `no_write_path` pinned at exactly 4 — this PR adds no write site. Implementation by a Claude Opus subagent; verification by Claude Fable 5 with Amy. Cross-family reviews to follow as comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,15 +17,17 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
-- **`read_cas` — a tool for reading artifacts back, replacing the
-  `kaibo://cas/<digest>` resource.** Pass a digest and metadata comes first: mime,
-  total size, binary or not, the artifact's label, the range served, and the real
-  file path when the store is on disk (open that directly for anything large).
-  Reads are bounded — `length: 0` is a cheap look at what an object is, omitting
-  `length` returns the first 64 KiB and tells you the total, and `offset` pages the
-  rest. Text arrives as text; a small image arrives as an image you can see, and a
-  large one as metadata and a path instead of a wall of base64. Advertised whenever
-  the media CAS is on.
+- **`read_cas` — read a stored artifact back by digest.** Metadata comes first on
+  every response: mime, total size, binary or not, the artifact's label when its
+  record carries one, the range served, and the real file path when the store is on
+  disk (open that directly for anything large). Reads are bounded, and the default
+  fits the object: text gives up to 64 KiB from `offset` and tells you the total so
+  you can page; an image up to 5 MiB arrives whole and viewable, a larger one as
+  metadata alone; any other binary gives metadata until you ask for a range. `length`
+  is capped at 1 MiB and a larger ask is refused rather than trimmed, and paging always
+  advances — a window landing inside a multi-byte character returns those exact bytes
+  as base64 with a note, so a caller resuming at the range it was handed never stalls.
+  Advertised whenever the media CAS is on.
 
 - **A consult can hand you bulk output as an artifact instead of spending your
   context on it.** Ask with `save_artifacts: true` and the investigating model can
@@ -75,11 +77,10 @@ record. Each later release appends a new section at the top.
   gains a `[cas]` section reporting the knob and the live mode
   (`disk` / `memory` / `off`).
 
-- **Artifact retrieval is operator-surface only.** The new `kaibo://cas/{digest}` MCP
-  resource serves an artifact's bytes (base64, correct mime) to the calling client —
-  the inner model team never sees the CAS: it is not mounted into kaish and no
-  cast-facing tool reads it, so one project's team can never enumerate another
-  project's artifacts.
+- **Artifact retrieval is operator-surface only.** The `read_cas` tool serves an
+  artifact's content to the calling client — the inner model team never sees the CAS:
+  it is not mounted into kaish and no cast-facing tool reads it, so one project's team
+  can never enumerate another project's artifacts.
 
 - **Model listings show each model's output ceiling.** `kaibo models` and `list_models`
   render the provider's advertised max completion tokens beside the context window
@@ -129,14 +130,9 @@ record. Each later release appends a new section at the top.
 
 ### Removed
 
-- **The `kaibo://cas/<digest>` MCP resource, replaced by the `read_cas` tool above.**
-  Resources are ambient — hosts prefetch them, attach them to a turn, treat them as
-  context — which is a reasonable posture for a config file kaibo publishes and the
-  wrong one for bytes a model just wrote; a tool call is deliberate, with explicit
-  arguments and a permission prompt in most hosts. And `resources/read` had no way to
-  ask for less than everything: a 3.8 MB artifact came back as roughly 5 MB of base64
-  in a single read. The `kaibo://cas/<digest>` **string** is unchanged and still names
-  every artifact in footers and results — `read_cas` takes the digest out of it.
+- **The `kaibo://cas/<digest>` MCP resource** — briefly on `main`, never in a release,
+  replaced before shipping by the `read_cas` tool above. A stale request for that URI
+  answers with the migration rather than a bare unknown-resource error.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,16 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **`read_cas` — a tool for reading artifacts back, replacing the
+  `kaibo://cas/<digest>` resource.** Pass a digest and metadata comes first: mime,
+  total size, binary or not, the artifact's label, the range served, and the real
+  file path when the store is on disk (open that directly for anything large).
+  Reads are bounded — `length: 0` is a cheap look at what an object is, omitting
+  `length` returns the first 64 KiB and tells you the total, and `offset` pages the
+  rest. Text arrives as text; a small image arrives as an image you can see, and a
+  large one as metadata and a path instead of a wall of base64. Advertised whenever
+  the media CAS is on.
+
 - **A consult can hand you bulk output as an artifact instead of spending your
   context on it.** Ask with `save_artifacts: true` and the investigating model can
   write a generated corpus, a long inventory, or a fixture into kaibo's media store;
@@ -45,7 +55,7 @@ record. Each later release appends a new section at the top.
   `[casts.artist] image = "sd/core"`. The tool never inlines bytes: every artifact
   lands in kaibo's content-addressed media store under its own SHA-256 digest with a
   provenance sidecar (prompt, model, cast, timestamp, mime, seed), and the result
-  lists per-artifact `kaibo://cas/<digest>` resource URIs — plus the real file path
+  lists per-artifact `kaibo://cas/<digest>` addresses — plus the real file path
   when the store is on disk. Provider-native options (`aspect_ratio`,
   `output_format`, `seed`, `negative_prompt`, ...) pass through a `fields` object
   verbatim. Operations the provider declares deferred return a `job-N` handle on the
@@ -116,6 +126,17 @@ record. Each later release appends a new section at the top.
   the template you copy, `kaibo://config` is the resolved live state, and
   `kaibo://config/guide` explains what any of it means. The `configure` prompt points at
   all three.
+
+### Removed
+
+- **The `kaibo://cas/<digest>` MCP resource, replaced by the `read_cas` tool above.**
+  Resources are ambient — hosts prefetch them, attach them to a turn, treat them as
+  context — which is a reasonable posture for a config file kaibo publishes and the
+  wrong one for bytes a model just wrote; a tool call is deliberate, with explicit
+  arguments and a permission prompt in most hosts. And `resources/read` had no way to
+  ask for less than everything: a 3.8 MB artifact came back as roughly 5 MB of base64
+  in a single read. The `kaibo://cas/<digest>` **string** is unchanged and still names
+  every artifact in footers and results — `read_cas` takes the digest out of it.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -457,8 +457,9 @@ OpenAI-compatible image endpoints (hosted gpt-image, or a local stable-diffusion
 `sd-server` speaking the same `/v1/images/generations` shape). The bytes never inline into
 your context: each artifact lands in kaibo's content-addressed media store with a
 provenance sidecar (prompt, model, cast, seed), and the result lists per-artifact
-digests as `kaibo://cas/<digest>` resource URIs — plus the real file path when the
-store is on disk. Advertised only when a configured cast carries an `image` slot and
+digests as `kaibo://cas/<digest>` addresses — plus the real file path when the store is
+on disk. Fetch one with the `read_cas` tool: metadata first, bounded ranges on request,
+and a small image comes back viewable rather than as a wall of base64. Advertised only when a configured cast carries an `image` slot and
 the `[cas]` store is on; the project stays untouched — the store lives at a fixed XDG
 data path the model can't steer. See "Media CAS" in [`docs/config.md`](docs/config.md).
 
--- a/README.md
+++ b/README.md
@@ -458,8 +458,9 @@ OpenAI-compatible image endpoints (hosted gpt-image, or a local stable-diffusion
 your context: each artifact lands in kaibo's content-addressed media store with a
 provenance sidecar (prompt, model, cast, seed), and the result lists per-artifact
 digests as `kaibo://cas/<digest>` addresses — plus the real file path when the store is
-on disk. Fetch one with the `read_cas` tool: metadata first, bounded ranges on request,
-and a small image comes back viewable rather than as a wall of base64. Advertised only when a configured cast carries an `image` slot and
+on disk. Fetch one with the `read_cas` tool: metadata first, bounded reads (text pages by
+`offset`, an image up to 5 MiB comes back viewable), and never a wall of base64 you didn't
+ask for. Advertised only when a configured cast carries an `image` slot and
 the `[cas]` store is on; the project stays untouched — the store lives at a fixed XDG
 data path the model can't steer. See "Media CAS" in [`docs/config.md`](docs/config.md).
 
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -323,8 +323,8 @@ scratch_limit_bytes = 67108864
 #
 # What it buys: a consult producing bulk output (a generated corpus, a long inventory, a
 # fixture) writes it into the CAS and names the digest in its answer, instead of
-# spending your context window on material you may only want to keep. You read it at
-# kaibo://cas/<digest>, or you do not.
+# spending your context window on material you may only want to keep. You read it back
+# with the `read_cas` tool (metadata first, bounded ranges), or you do not.
 #
 # Limits are fixed, not configurable: 1 MiB per artifact, 8 artifacts and 8 MiB per call,
 # and a one-line label of at most 200 bytes. A save past a limit is refused and stores
--- a/docs/config.md
+++ b/docs/config.md
@@ -865,23 +865,34 @@ save emits and the session the answer was recorded into.
 
 **Retrieval is the `read_cas` tool, and it is operator-surface only.** Every producer
 names its artifacts by a `kaibo://cas/<digest>` address; `read_cas` takes the digest out
-of one and hands the content back. The MCP client and CLI call it; the inner model team
-never can — the CAS is not mounted into kaish and no cast-facing tool reads it, because
-kaibo state spans projects and a browsable CAS would let one project's team enumerate
-another's artifacts.
+of one and hands the content back. The MCP client calls it; the inner model team never can
+— the CAS is not mounted into kaish and no cast-facing tool reads it, because kaibo state
+spans projects and a browsable CAS would let one project's team enumerate another's
+artifacts. There is no `kaibo cas` CLI command; on disk, the path a read reports is how an
+operator reaches an object with their own tools.
 
-Reads are **metadata-first and bounded**:
+Reads are **metadata-first and bounded**, and the default depends on what the object is:
 
-| ask | you get |
-|---|---|
-| `length: 0` | metadata only — mime, total bytes, binary or not, the label, and the file path in disk mode |
-| no `length` | metadata plus the first 64 KiB from `offset`, never the whole object |
-| `offset` + `length` | metadata plus exactly that range (`length` up to 1 MiB) |
+| ask | text object | image | any other binary |
+|---|---|---|---|
+| `length: 0` | metadata only | metadata only | metadata only |
+| no `length` | metadata + up to 64 KiB from `offset` | the whole image, viewable, when it is ≤ 5 MiB; otherwise metadata only | metadata only |
+| `offset` + `length` | metadata + that range as text | metadata + that range as base64 | metadata + that range as base64 |
 
-Text comes back as text. A small image (up to 5 MiB) with no range asked for comes back as
-a viewable image; a larger one comes back as metadata and a path, because base64 in a
-context window helps nobody — open the file, or ask for a range explicitly. A binary
-object is never dumped as base64 unless you asked for a range.
+Metadata is always the first content block: digest, URI, mime, total bytes, binary or not,
+the label when the object's record carries one, a `provenance:` line when it has no record
+at all, the range served, and the real file path **in disk mode only** — memory mode has
+no file, so there is nothing to point at. `length` is capped at 1 MiB; a larger ask is
+refused rather than trimmed, so a caller's next `offset` is never wrong.
+
+**A binary object is never dumped as base64 unless you ask for a range.** An image past
+5 MiB comes back as metadata (plus the path, on disk) rather than ~7 MiB of base64 in your
+context — open the file, or page it deliberately.
+
+**Paging always advances.** A window that begins or ends inside a multi-byte character
+comes back as base64 of exactly the bytes you asked for, with a note saying why, so the
+served range still moves. Resuming at the range you were handed always terminates and
+always reassembles the object byte for byte.
 
 `kaibo://cas/<digest>` **was** an MCP resource until 2026-08-05 and no longer is. Two
 reasons. Hosts treat resources as ambient context — some prefetch, some auto-attach —
--- a/docs/config.md
+++ b/docs/config.md
@@ -878,7 +878,7 @@ Reads are **metadata-first and bounded**:
 | no `length` | metadata plus the first 64 KiB from `offset`, never the whole object |
 | `offset` + `length` | metadata plus exactly that range (`length` up to 1 MiB) |
 
-Text comes back as text. A small image (up to 2 MiB) with no range asked for comes back as
+Text comes back as text. A small image (up to 5 MiB) with no range asked for comes back as
 a viewable image; a larger one comes back as metadata and a path, because base64 in a
 context window helps nobody — open the file, or ask for a range explicitly. A binary
 object is never dumped as base64 unless you asked for a range.
--- a/docs/config.md
+++ b/docs/config.md
@@ -863,11 +863,32 @@ an operator, not an audit trail, and a content-addressed store that never rewrit
 be one. For a durable per-call record, kaibo's answers are the tool-call telemetry each
 save emits and the session the answer was recorded into.
 
-**Retrieval is operator-surface only.** A generation result returns each artifact's
-digest, a `kaibo://cas/<digest>` resource URI, and (in disk mode) the real filesystem
-path. The MCP client and CLI read those; the inner model team never sees the CAS — it is
-not mounted into kaish and no cast-facing tool reads it, because kaibo state spans
-projects and a browsable CAS would let one project's team enumerate another's artifacts.
+**Retrieval is the `read_cas` tool, and it is operator-surface only.** Every producer
+names its artifacts by a `kaibo://cas/<digest>` address; `read_cas` takes the digest out
+of one and hands the content back. The MCP client and CLI call it; the inner model team
+never can — the CAS is not mounted into kaish and no cast-facing tool reads it, because
+kaibo state spans projects and a browsable CAS would let one project's team enumerate
+another's artifacts.
+
+Reads are **metadata-first and bounded**:
+
+| ask | you get |
+|---|---|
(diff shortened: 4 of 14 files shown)


tobert:
## Cross-family review 1/2 — DeepSeek (invariants): ready to merge

Holistic review over the branch worktree, cast `deepseek` via kaibo's CLI consult. All AGENTS.md invariants confirmed with citations; findings are minor (five stale doc-comments, a nice-to-have follower-guard unit test, two cosmetic planner edge cases). Will be folded with anything the GPT pass finds.

---

## Verdict: Ready to merge, with minor doc-comment cleanups recommended

The change is sound across every dimension the question raises. Below, each claim is grounded in the concrete `file:line` evidence.

---

### 1. Inner model team never gains access to `read_cas` — CONFIRMED

The client-surface verbs are explicitly enumerated in a `CLIENT_ONLY_VERBS` constant:

**`src/consult/engine.rs:6453-6459`**
```rust
const CLIENT_ONLY_VERBS: [&str; 5] = [
    "read_cas",
    "generate",
    "job_get",
    "job_list",
    "consult_submit",
];
```

The inner team's toolset is built by `consult_tools` (`src/consult/engine.rs:1883-1967`). It constructs exactly `[RunKaish, explore, ?ViewImage, ?SaveArtifact]` — `read_cas` is never added. This is test-pinned by two tests:

- **`src/consult/engine.rs:6466-6501`** (`the_driver_toolset_never_carries_a_client_surface_verb`): asserts every verb in `CLIENT_ONLY_VERBS` is absent from the driver toolset under three configurations (default, with artifact sink, vision synth).
- **`src/consult/engine.rs:6508-6558`** (`a_delegated_sweep_never_carries_save_artifact`): a scripted end-to-end test that additionally asserts the explorer sweep's *running* toolset carries none of `CLIENT_ONLY_VERBS`.

The EXPLORER sweep builds its toolset in `run_explore_phase` (called from `explore_with` at `src/consult/engine.rs:1998`), which takes only the explorer arm and the preamble — no sink, no CAS access. The `explore` tool surface (`src/server/mod.rs` tool descriptions) names `run_kaish` + `explore′` — no `read_cas`.

---

### 2. Read-only invariant untouched — CONFIRMED

`src/server/cas_read.rs` is a **pure planner**: it takes an already-verified `CasObject` (bytes in memory) and computes a `CasView`. It contains **zero** filesystem calls — no `std::fs`, no `OpenOptions`, no `create_dir_all`. The handler at `src/server/mod.rs:2634-2702` reads from the CAS store via `store.get(&digest)` (which calls the existing `Cas::get` / `MemoryCas::get`), then feeds the verified bytes into the planner.

`tests/no_write_path.rs:38-41` pins the blessed write sites at exactly 4:
```
const EXPECTED_BLESSED_LINES: usize = 4;
```
(1 in `store.rs` + 3 in `cas.rs`). The planner adds zero new write sites.

---

### 3. Dedup/existence oracles do not reappear — CONFIRMED

The "not found" message at **`src/server/mod.rs:2660-2668`**:
```rust
"no artifact with digest {} — it was never produced on this \
 store, or (in memory mode) it did not survive a restart"
```
The digest is a content hash the caller already computed. No probing of file existence without knowing the content. The format is validated first at **`src/server/mod.rs:2651`** (`Digest::from_hex`), which enforces exactly 64 lowercase hex before any lookup touches the filesystem — path-traversal-shaped strings like `../../etc/passwd` are rejected by the charset/length check alone (`src/cas.rs:273-280`).

---

### 4. Digest validation before lookup — CONFIRMED

**`src/server/mod.rs:2648-2652`**:
```rust
// Validated before it can become part of a path — the same structural guard the
// resource this tool replaces applied, and the reason a traversal-shaped
// "digest" never reaches a lookup.
let digest = crate::cas::Digest::from_hex(&input.digest)
    .map_err(|e| McpError::invalid_params(format!("{e}"), None))?;
```
The `?` returns early on invalid format; `store.get(&digest)` is only reached after validation. Test-pinned at **`src/server/mod.rs:5725-5753`** (`read_cas_validates_the_digest_and_reports_a_miss_cleanly`), which tests four invalid inputs (`../../etc/passwd`, `ABCD`, empty string, 63-char hex) and asserts each is refused with "64 lowercase hex".

---

### 5. Corrupt objects loud, never folded into not-found — CONFIRMED

**`src/server/mod.rs:2658-2673`**:
```rust
let (bytes, ext) = match store.get(&digest) {
    Ok(Some(found)) => found,
    Ok(None) => {
        return Err(McpError::invalid_params(...))  // "no artifact"
    }
    // Corrupt or unreadable stays loud and distinct
    Err(e) => return Err(McpError::internal_error(format!("{e}"), None)),
};
```
`Ok(None)` → `invalid_params` (not found). `Err(_)` → `internal_error` (corrupt). The two paths are distinct error types: `invalid_params` vs `internal_error`. Test-pinned at **`src/server/mod.rs:5759-5785`** (`read_cas_surfaces_a_corrupt_object_loudly`), which tampers with an on-disk object and asserts the error message contains "corrupt".

The corrupt error message itself (`src/cas.rs:227-231`) names both the expected and actual hashes — the caller already holds the content (they computed the digest), so neither hash is a secret.

---

### 6. UTF-8 window trimming — SOUND, one cosmetic edge case

`utf8_window` at **`src/cas_read.rs:236-254`** handles three cases:

- **Start on continuation byte** (line 238-239): advances `from` past any `0x80..=0xBF` bytes until a start byte or `end`.
- **End mid-character** (line 246, `error_len().is_none()`): trims to `valid_up_to()`, returning the largest whole-character prefix.
- **Corrupt bytes** (line 252, `error_len().is_some()`): returns `None`, falling through to base64 in the planner.

Test coverage:
- `a_window_splitting_a_multibyte_character_trims_to_a_boundary` (`cas_read.rs:510-525`): 3-byte window into `"ééé"` (each é is 2 bytes) → returns `"é"` with range `0..2`.
- `undecodable_textual_bytes_fall_back_to_base64` (`cas_read.rs:531-536`): corrupt bytes `[0x66, 0x6f, 0x6f, 0xff, 0xfe, 0x62, 0x61, 0x72]` → base64.
- `offset_pages_through_a_large_object` (`cas_read.rs:354-377`): two pages reassemble exactly.

**One cosmetic edge case, not a bug:** When the window contains only continuation bytes or only a multi-byte start byte with no continuation (e.g., a 1-byte window containing `0xC3`), `utf8_window` returns `Some((from, from, ""))` — an empty `Text("")` body. The metadata reports the empty range honestly (`bytes N..N`). This is technically correct (no complete characters to serve) and occurs only in an extremely rare window shape. The empty text block is wasteful but harmless; the caller sees metadata + an empty content block. Not worth a code change given the rarity.

**Empty object + textual + no length** also produces a slightly misleading `served:` message (`"this object is binary"` for a textual 0-byte object, at `cas_read.rs:172-174`). This is because the message dispatch at lines 166-175 falls through to the `else` branch when `offset == 0 && total == 0`. Again, extremely rare (0-byte artifacts) and the metadata block alongside it correctly reports `binary: false` and `bytes: 0`. The caller is not misled.

---

### 7. FOLLOWER_TOOL_NAMES — SOUND, one test gap

**Definition** at `src/server/mod.rs:895-896`:
```rust
pub(crate) const FOLLOWER_TOOL_NAMES: &[&str] =
    &["read_cas", "job_get", "job_cancel", "job_list", "job_wait"];
```

**Usage** in `has_substantive_tools()` at `src/server/mod.rs:1338-1341`:
```rust
pub fn has_substantive_tools(&self) -> bool {
    self.advertised_tools()
        .iter()
        .any(|name| !FOLLOWER_TOOL_NAMES.contains(&name.as_str()))
}
```

Called from `main.rs:207` after the handler is built. This is the second startup guard (the first, `all_disabled()`, runs pre-handler on flags alone). The guard correctly refuses a server whose entire advertised surface is only followers.

**Test coverage** is almost complete:
- `tests/gating.rs:380-438` (`a_server_left_with_no_staffable_tools_refuses_to_start`): the staffing-path integration test — every backend unreachable, `--no-run-kaish --no-list-models`, asserts the guard fires with "no tools left to advertise" and "cast".
- `tests/gating.rs:444+` (`all_tools_disabled_refuses_to_start`): the flag-path integration test — every `--no-*` flag set.

**Gap:** There is no test that isolates the scenario "all flags OFF, but CAS is ON, so `read_cas` is the only advertised tool." In that scenario the flag guard (`all_disabled()`) fires first because all flags are off, so the FOLLOWER guard is never reached. But if a future change adds a flag for `read_cas`, the FOLLOWER guard would be the only defense, and no test would catch its absence. **Recommendation:** add a dedicated subprocess test or a unit test on `has_substantive_tools()` that asserts `false` when the handler advertises only `["read_cas"]` (e.g., by constructing a handler with CAS on but every other tool off). This is a **nice-to-have**, not a merge-blocker — the existing integration tests still exercise the guard indirectly.

`read_cas` gating is also test-pinned at `src/server/mod.rs:5791-5804` (`read_cas_is_advertised_only_while_the_media_cas_is_on`), confirming it rides CAS liveness alone.

---

### 8. Dangling resource references — MINOR DOC ISSUES

The old `kaibo://cas/<digest>` resource route is **properly removed**:
- `render_resource` (`src/server/mod.rs:3914-3937`): no CAS arm — returns `None` for `kaibo://cas/...` URIs.
- `kaibo_resource_templates()` (`src/server/mod.rs:3662-3683`): deliberately excludes the CAS template (comment at 3677-3681).
- `read_kaibo_resource_with_config()` (`src/server/mod.rs:4185-4243`): a `kaibo://cas/...` URI falls through to "unknown resource" at line 4238.
- Test-pinned at `src/server/mod.rs:5810-5844` (`the_cas_resource_route_is_gone`).

**Stale doc-comments found** (all reference the removed resource; none affect functionality):

| File:Line | Current text | Should say |
|---|---|---|
| `tests/cas.rs:662` | "the kaibo://cas resource is its only retrieval channel" | `read_cas` tool |
| `tests/cas.rs:794` | "mislabel the `kaibo://cas/<digest>` resource then stamps onto the bytes" | `read_cas` |
| `src/cas.rs:630` | "the extension carries the mime type a resource read needs to stamp on" | `read_cas` or "retrieval" |
| `src/cas.rs:993` | "a `text/plain` resource read" | "a `read_cas` retrieval" or "a text/plain read" |
| `src/artifact.rs:177` | "advertise a mime the resource will not serve and a path" | `read_cas` |

The `TOOLS_DOC` constant (`src/server/mod.rs:3692-3912`) is already updated — the "Reading artifacts back" section at lines 3868-3877 correctly describes `read_cas` as the retrieval tool. The instructions string (`kaibo_instructions_with_scope`) and tool descriptions (`server/mod.rs:2622-2632`) are all current. No dangling references in the MCP surface.

---

### 9. Error messages — no leaks

- **Invalid digest** (`server/mod.rs:2652`): surfaces the `Digest::from_hex` error (charset/length rule). No path or existence information.
- **Not found** (`server/mod.rs:2660-2668`): names the digest and the two possible causes. The caller computed the digest from content they hold, so the digest itself is not secret.
- **Corrupt** (`server/mod.rs:2673` via `CasError::Corrupt` at `cas.rs:227-231`): names expected vs actual hash. Content-addressed — reveals nothing the caller couldn't compute from the bytes.
- **Length refused** (`cas_read.rs:123-128`): names the ceiling and suggests paging via `offset`. No path or store state.

---

### Summary

**Ready to merge.** The five minor doc-comment stale references (`tests/cas.rs:662,794`, `src/cas.rs:630,993`, `src/artifact.rs:177`) should be cleaned up — they say "resource" where they should say `read_cas`. The recommended FOLLOWER_TOOL_NAMES unit test is a nice-to-have follow-up, not a merge-blocker. Every invariant in AGENTS.md holds: operator surface only, read-only intact at 4 blessed write sites, no existence oracles, digest-first validation, corrupt-is-loud, no dangling resource routes.

———
tokens · 1634763 in · 23456 out · 10829 reasoning · 1524992 cached

tobert:
## Cross-family review 2/2 — GPT via or-gpt (operational seams): NOT ready, four findings

Holistic review over the branch worktree, cast `or-gpt`. Finding 1 (UTF-8 window can return an empty, non-advancing page — an infinite paging loop) is verified real and is a genuine correctness bug. Findings 2 and 4 (description/schema/docs drift from planner behavior; migration path + stale references) are verified real. Finding 3 (always render a label line, distinguish provenance-missing) is **partially disputed**: per the project's design ruling the sidecar is best-effort housekeeping, not a promised record, so label stays omitted when absent — but the real half (a missing sidecar is a distinguishable, meaningful state) is being taken as a `provenance: absent` marker. Fix batch is with the implementation agent; DeepSeek's two minor notes ride along.

Meanwhile the live probe of the current build: a metadata-only HEAD costs ~60 tokens, and a 3.8 MB PNG under the (now 5 MiB) inline threshold arrived as a one-hop rendered image block with metadata leading — no host side-file detour, no base64 wall.

---

## Verdict: **not ready to merge**

The central design is sound, but the branch still has one correctness bug in UTF-8 paging and several client-facing contract/documentation mismatches. I would fix the following before merge.

### 1. UTF-8 windows can return a successful, empty, non-progressing page

For `"é"` with `offset: 0, length: 1`, `from_utf8([0xC3])` reports an incomplete trailing character with `valid_up_to() == 0`. `utf8_window` therefore returns `Some((0, 0, ""))`; `plan` turns that into an empty text block with `served: bytes 0..0`. A caller resuming at the reported endpoint repeats the same request indefinitely. A window trimmed at both edges can produce the same empty result.

> ```rust
> if let Some((from, to, text)) = utf8_window(...) {
>     ...
>     body: Body::Text(text.to_string()),
> }
> ```
> `src/server/cas_read.rs:182-190`
>
> ```rust
> Err(e) if e.error_len().is_none() => {
>     let to = from + e.valid_up_to();
>     ...
>     .map(|text| (from, to, text))
> }
> ```
> `src/server/cas_read.rs:236-251`

The existing multibyte test keeps one complete character and does not exercise an all-trimmed window.

> ```rust
> // Each `é` is two bytes, so a 3-byte window from 0 splits the second one.
> ...
> assert_eq!(view.body, Body::Text("é".to_string()));
> ```
> `src/server/cas_read.rs:505-525`

**Required fix:** successful pages must not contain an empty `Body::Text` with a non-advancing range. For an explicit range containing no complete scalar, either:

- refuse it with an actionable message telling the caller to increase `length` or align `offset`; or
- serve the exact bytes as explicitly-labelled base64 so the byte cursor advances.

Add tests for:

- a leading character cut before its first complete scalar;
- a window made entirely of continuation bytes;
- both edges trimmed to empty;
- a start trim that still leaves a later complete character;
- repeated paging that either reassembles the text or returns an actionable refusal.

There is also a smaller empty-object bug: an empty textual artifact at `offset: 0` with omitted `length` is described as binary because the zero-range explanation falls into the final branch.

> ```rust
> let why = if start >= total && offset > 0 {
>     ...
> } else if length.is_some() {
>     ...
> } else {
>     "nothing: this object is binary ..."
> };
> ```
> `src/server/cas_read.rs:166-175`

Empty text is legal because `ArtifactSink::save` checks only the maximum byte size, not a minimum.

> ```rust
> let bytes = content.as_bytes();
> if bytes.len() > self.caps.per_artifact_bytes { ... }
> ```
> `src/artifact.rs:400-410`

---

### 2. The shipped schema and prose do not accurately describe default and maximum reads

The tool description says omitting `length` returns 65,536 bytes, but the planner does that only for text. Binary objects get metadata only, except for the whole-image path.

> ```rust
> let want = match length {
>     Some(asked) => asked,
>     None if obj.ext.is_textual() => DEFAULT_READ_BYTES,
>     None => 0,
> };
> ```
> `src/server/cas_read.rs:155-164`
>
> ```rust
> omitting `length` returns the first 65536 bytes
> ```
> `src/server/mod.rs:2622-2632`

The `length` parameter also fails to state the explicit 1 MiB ceiling, even though the planner refuses anything larger.

> ```rust
> /// How many bytes to read. Omit for a bounded default window (65536 bytes) ...
> pub length: Option<usize>,
> ```
> `src/server/mod.rs:630-634`
>
> ```rust
> if asked > MAX_READ_BYTES {
>     return Err(...);
> }
> ```
> `src/server/cas_read.rs:120-129`

The configuration guide repeats the universal 64 KiB claim in its table, then contradicts it below by saying binary is never returned unasked.

> ```text
> | no `length` | metadata plus the first 64 KiB from `offset` ... |
> ...
> A binary object is never dumped as base64 unless you asked for a range.
> ```
> `docs/config.md:873-884`

Large-image wording also promises a path where memory mode cannot provide one. In memory mode an oversized image returns metadata only until the caller explicitly pages it.

> ```rust
> if let Some(path) = obj.path {
>     out.push_str(&format!("\npath: {}", path.display()));
> }
> ```
> `src/server/cas_read.rs:218-224`
>
> ```rust
> MediaStore::Memory(_) => None,
> ```
> `src/cas.rs:1034-1042`
>
> ```text
> a large one comes back as metadata and a path
> ```
> `src/server/mod.rs:2630-2632`

**Required fix:** make the tool description, parameter schema, guide, README, and changelog say:

- omitted `length`: up to 64 KiB of **text** from `offset`;
- omitted `length` on binary: metadata only, except a qualifying whole-image block;
- explicit `length`: base64 range, maximum 1 MiB, refused rather than clamped;
- oversized image: metadata only, plus a path only in disk mode; memory mode must page explicitly.

---

### 3. Metadata does not currently satisfy the stated label/provenance contract

`label` is omitted rather than always represented. That is the common generated-image case because generated provenance deliberately stores `label: None`.

> ```rust
> if let Some(label) = obj.label {
>     out.push_str(&format!("\nlabel: {label}"));
> }
> ```
> `src/server/cas_read.rs:208-224`
>
> ```rust
> tool: Some("generate".to_string()),
> slot: None,
> label: None,
> ```
> `src/server/mod.rs:4122-4135`

The planner test explicitly locks in omission.

> ```rust
> assert!(
>     !view.meta.contains("label:"),
>     "no label line without a label"
> );
> ```
> `src/server/cas_read.rs:315-325`

More importantly, an ordinary “no label” and a missing, unreadable, or malformed provenance sidecar are collapsed into the same `None`.

> ```rust
> /// `None` covers ... no sidecar ... an unreadable one, and one whose
> /// JSON doesn't deserialize
> pub fn provenance_for(...) -> Option<Provenance> {
>     let bytes = std::fs::read(...).ok()?;
>     serde_json::from_slice(&bytes).ok()
> }
> ```
> `src/cas.rs:619-627`
>
> ```rust
> let label = store.provenance(&digest).and_then(|p| p.label);
> ```
> `src/server/mod.rs:2675-2676`

This matters because a sidecar-less disk object is intentionally still retrievable through extension probing.

> ```rust
> /// The probe is the fallback ...
> pub fn entry_for(...) -> Option<(PathBuf, Extension)> {
>     ...
>     Extension::ALL.into_iter().find_map(...)
> }
> ```
> `src/cas.rs:641-660`

**Required fix:** carry provenance availability separately from `Option<label>`. Metadata should distinguish at least:

- provenance present, label present;
- provenance present, no label;
- provenance missing/unreadable.

Given the stated contract, render a label line consistently, such as `label: (none)`, rather than silently omitting the field. Add a read test for an object whose content exists but provenance does not.

---

### 4. Resource migration handling and dangling references are incomplete

The old resource is correctly absent from templates and dispatch, but a host using a cached template gets only `unknown resource`, with no recovery instruction.

> ```rust
> let body = render_resource(uri, schemas)
>     .ok_or_else(|| McpError::resource_not_found(
>         format!("unknown resource: {uri}"), None
>     ))?;
> ```
> `src/server/mod.rs:4238-4242`
>
> ```rust
> .expect_err("the CAS resource route no longer exists");
> assert!(err.message.to_lowercase().contains("unknown"));
> ```
> `src/server/mod.rs:5816-5831`

**Required fix:** keep the resource unadvertised and unserved, but recognize `kaibo://cas/<digest>` in the not-found path and return a migration error such as: “This resource was removed; call `read_cas` with digest …”. That directly helps hosts which cached the former resource template.

The unreleased changelog also simultaneously adds the removed resource and removes it later. The unreleased section should describe the state that will actually ship.

> ```text
> Artifact retrieval is operator-surface only. The new
> `kaibo://cas/{digest}` MCP resource serves ...
> ```
> `CHANGELOG.md:78-82`
>
> ```text
> The `kaibo://cas/<digest>` MCP resource, replaced by the `read_cas` tool
> ```
> `CHANGELOG.md:130-139`

Delete or rewrite the stale Added entry.

Several source/test comments still call the deleted resource the retrieval channel.

> ```text
> the kaibo://cas resource is its only retrieval channel
> ```
> `tests/cas.rs:659-663`
>
> ```text
> the `kaibo://cas/<digest>` resource is the ONLY retrieval channel
> ```
> `tests/save_artifact.rs:730-734`
>
> ```text
> the CAS resource
> ```
> `src/server/mod.rs:1311-1314`

There is also a concrete CLI documentation error: the guide says the CLI calls `read_cas`, but no such CLI command exists in the command dispatch.

> ```text
> The MCP client and CLI call it
> ```
> `docs/config.md:866-870`
>
> The available CLI command arms contain consult, oneshot, explore, kaish, batch, models, config, configure, and example-config, but no CAS read command. `src/main.rs:41-62`

The same incorrect CLI claim appears in code documentation.

> `src/cas.rs:174-179`  
> `src/server/mod.rs:111-121`

---

## Other reviewed seams

### Image dispatch is mechanically correct, but boundary and host behavior need stronger coverage

The exact 2 MiB boundary is included because the comparison is `<=`; 2 MiB + 1 is excluded.

> ```rust
> total > 0 && total <= INLINE_IMAGE_MAX_BYTES
> ```
> `src/server/cas_read.rs:134-151`

Current tests cover a 1 KiB image and threshold + 1, not exactly the threshold.

> `src/server/cas_read.rs:449-468`  
> `src/server/mod.rs:5664-5693`

Add an exact-boundary test. The MIME is consistently derived from the closed `Extension` mapping and placed on the MCP image block.

> `src/cas.rs:368-381`  
> `src/server/mod.rs:2697-2699`

However, the end-to-end image test uses `b"pretend-png-bytes"`, proving only block construction, not that a real host can decode/render it.

> `src/server/mod.rs:5632-5661`

The source cannot establish what every MCP host does with an image block. Use a valid tiny PNG/WebP/GIF/JPEG fixture and run at least one real-host probe; soften categorical prose such as “hosts render straight to a vision model” to “compatible hosts can render or deliver the image block.” The current generated-media path validates declared MIME but does not sniff image magic. `src/server/mod.rs:4087-4113`

### CAS liveness and follower gating are coherent

`read_cas` follows `[cas] enabled`, while job verbs follow effective producer liveness.

> ```rust
> let read_cas_live = config.cas.enabled;
> let jobs_live = consult_live || batch_live || deliberate_live || generate_live;
> ```
> `src/server/mod.rs:917-933`

`read_cas` and all job verbs are followers, and the startup guard requires at least one non-follower. Therefore the shipped binary cannot successfully start with only retrieval/collection tools.

> ```rust
> pub(crate) const FOLLOWER_TOOL_NAMES: &[&str] =
>     &["read_cas", "job_get", "job_cancel", "job_list", "job_wait"];
> ```
> `src/server/mod.rs:886-896`
>
> ```rust
> .any(|name| !FOLLOWER_TOOL_NAMES.contains(&name.as_str()))
> ```
> `src/server/mod.rs:1334-1342`
>
> ```rust
> anyhow::ensure!(handler.has_substantive_tools(), ...)
> ```
> `src/main.rs:193-213`

The startup error is nevertheless misleading when `read_cas` remains advertised: it says “no tools left to advertise” rather than “only follower tools remain.”

> `src/main.rs:202-212`

Update that wording and its integration assertion. The README’s blanket statement that each tool has a `--no-<tool>` flag is also now false because `read_cas` deliberately has none.

> `README.md:384-387`  
> `tests/gating.rs:11-19`

A pre-existing liveness inconsistency remains: advertised job verbs use effective staffing, but the `job-N` call-time guard uses raw `consult`/`deliberate` flags. A batch-only staffable server can therefore accept the shape of a `job-N` even when no live route can mint one.

> `src/server/mod.rs:917-933`  
> `src/server/mod.rs:3247-3266`

That does not make a follower-only server start and is not introduced by `read_cas`, but it is worth a focused follow-up.

### Range verification and ordinary bounds are otherwise sound

Disk objects are read and hash-verified before `plan` slices them, including metadata-only requests.

> `src/server/mod.rs:2654-2687`  
> `src/cas.rs:662-689`

Offset arithmetic is overflow-safe and clips a requested length to the remaining bytes.

> ```rust
> let start = offset.min(total);
> let end = start.saturating_add(want).min(total);
> ```
> `src/server/cas_read.rs:131-164`

Past-EOF requests are metadata-only, and explicit lengths over 1 MiB are refused rather than clamped.

> `src/server/cas_read.rs:166-179`  
> `src/server/cas_read.rs:120-129`

### Error exposure is mostly consistent with the intended operator surface

Malformed digests are validated before path construction, and traversal-shaped input is tested.

> `src/server/mod.rs:2648-2652`  
> `src/server/mod.rs:5739-5752`

Disk read errors can include the absolute CAS path.

> `src/cas.rs:682-688`  
> `src/server/mod.rs:2670-2674`

That is consistent with successful disk-mode metadata intentionally exposing the same path, so I do not consider it a new confidentiality leak for a caller already holding the digest. The invalid-digest error does echo the complete supplied string, though, so bounding or summarizing very long malformed values would avoid an unnecessarily large error response. `src/cas.rs:190-196`

Once the UTF-8 behavior, metadata/provenance state, client schema, migration error, and stale documentation are corrected, the implementation should be in good shape for merge.

———
kaibo · cast `or-gpt` · explorer `~openai/gpt-mini-latest` · synth `~openai/gpt-latest`
tokens · 4580636 in · 49687 out · 4008927 cached · 27043 cache-write

tobert:
## Fix batch: all findings addressed, one disputed half recorded, one extra bug found

Commits `7366d18` + `4700f20`. Independently re-verified: **1013 tests / 0 failed** (branch started at 1005), clippy silent, `aws-lc-rs`/`mimalloc` trees empty, `no_write_path` pinned at exactly 4.

- **Finding 1 (paging loop) — fixed, and it was worse than reported.** Fixing the non-advancing empty page surfaced a second bug in the same function: the window's leading-continuation skip meant some bytes appeared in *no* response, so a pager reassembling by reported ranges got an object with holes (caught by a byte-exact reassembly test). One rule now governs: **the served range starts exactly where the caller asked; only the tail is trimmed.** A window that can't hold one whole character ships as base64 of the exact bytes with a `note:` explaining how to realign — the cursor always advances. Property test walks every window size over a multi-byte corpus; both loop-shaped tests carry iteration bounds (the first drafts hung — the honest failing-first evidence).
- **Finding 2 (contract drift) — fixed** across all five surfaces with one per-kind rule: text → up to 64 KiB from `offset`; image → whole and viewable at ≤ 5 MiB (threshold raised from 2 MiB on operator direction, "we'll see who complains"), else metadata + path in disk mode only; other binary → metadata until `length` is asked; `length` capped at 1 MiB, refused not clamped.
- **Finding 3 — partial, as previously stated.** Taken: `provenance: absent (this object was stored without its sidecar)` — the state a crashed put mints, now visible; proven by deleting a real sidecar and reading the object. Declined: a mandatory `label: (none)` line — the sidecar is best-effort housekeeping by design ruling, and the meaningful distinction is carried by `provenance:`.
- **Finding 4 — fixed**: `kaibo://cas/` recognized in the unknown-resource path with a migration message naming `read_cas` and the digest (route stays unadvertised); unreleased CHANGELOG rewritten to describe what ships; all seven stale comment sites and the false CLI claim (three places) corrected.
- **DeepSeek's notes — fixed**: `has_substantive_tools` unit test drives a real handler advertising only `["read_cas"]` (teeth proven by mutating `FOLLOWER_TOOL_NAMES`); the zero-byte textual `served:` line now tells the truth.

Live-validated on the operator's running server pre-fix-batch: metadata HEAD ≈ 60 tokens; a 3.8 MB PNG under the threshold arrived as a one-hop rendered image block.

PR #127, opened 2026-08-05
feat(cas): warn severely when disk-mode CAS sits on an ephemeral filesystem

tobert:
The last unbuilt P1 media item (decided 2026-07-30). The scenario: a container with no volume mounted looks like disk mode, accepts a generate/save prompt, **spends the user's provider credits**, writes the artifact, then destroys everything on exit. Now: at startup, disk mode probes the CAS directory's backing filesystem; overlayfs/tmpfs/ramfs draws a SEVERE warning naming the fix (mount a volume) — then proceeds. Deliberately not a refusal and not gated on an ack flag, per the decided design.

Design points:
- **`cas::Backing`** (`Ephemeral { fs }` / `Durable` / `Unknown`) with a statfs probe that walks to the nearest *existing* ancestor — the CAS root is created lazily, so probing the leaf would return Unknown on exactly the fresh-container case the guard exists for (test-pinned).
- **`Unknown` is never folded** into either verdict: warn-on-unknown would cry wolf on every non-Linux install; durable-on-unknown would claim what was never established. Probe failure is debug-level silence.
- **Queryable after the fact**: `[cas] backing` renders in `kaibo://config` when the probe found something, and `kaibo config` (CLI) probes on the same prediction — an operator can ask before spending anything.
- **No new dependency**: `libc` was already declared (Linux-only) for `store.rs`'s network-fs guard; the probe mirrors that code's shape, low-32-bit compare and all. Non-Linux is a quiet `Unknown` with the seam left for platform-specific detection later.
- **Memory mode never probes** — it already owns a severe warning about precisely this loss; double-warning is test-pinned out.
- Read-only work: `no_write_path` stays pinned at exactly 4.

Verification: 1022 tests / 0 failed, clippy silent, `aws-lc-rs`/`mimalloc` trees empty. Real teeth: `/dev/shm` probes as tmpfs (runs where present, skips cleanly elsewhere); mutation-tested recording and rendering. One failing-first test corrected the design premise itself (a nonexistent absolute path resolves through the ancestor walk to a real answer, not Unknown). The tracked issues.md entry is deleted (ship = delete).

Implementation by a Claude Opus subagent; verification by Claude Fable 5 with Amy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,18 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **kaibo warns when the media CAS is on a disk that will not survive the
+  container.** A container with no volume mounted looks exactly like durable disk
+  mode: persistence comes up, the store opens, artifacts write and read back — and
+  the whole filesystem evaporates on exit, after you have paid provider credits for
+  what was in it. On Linux, startup now checks the CAS directory's backing
+  filesystem and warns severely on overlayfs, tmpfs, or ramfs, naming the filesystem
+  and the directory. It **proceeds** — a throwaway store is a legitimate thing to
+  run on purpose — and `kaibo://config` reports the finding under `[cas] backing`
+  (as does `kaibo config`), so it is checkable before you spend anything instead of
+  only in startup log. A durable filesystem, or a check that cannot answer, says
+  nothing at all.
+
 - **`read_cas` — read a stored artifact back by digest.** Metadata comes first on
   every response: mime, total size, binary or not, the artifact's label when its
   record carries one, the range served, and the real file path when the store is on
--- a/docs/config.md
+++ b/docs/config.md
@@ -908,6 +908,21 @@ the same way, including one whose content the store already holds: exempting tho
 let a full store answer "is this content already here?" by succeeding for held bytes and
 refusing for new ones, across every project this kaibo has served.
 
+**Disk mode warns when the disk is not really a disk.** On Linux, startup checks what
+filesystem the CAS directory sits on. If it is overlayfs, tmpfs, or ramfs — the shape of a
+container with no volume mounted — kaibo logs a severe warning and **proceeds**. It is not
+a refusal and there is no acknowledgement flag: running on a throwaway filesystem on
+purpose is legitimate, and what is not legitimate is finding out only after a generation
+has been paid for. The warning names the filesystem and the directory; `kaibo://config`
+reports the same finding under `[cas] backing`, and `kaibo config` prints it, so you can
+check before spending anything rather than hunting through startup log after.
+
+The fix is to mount a volume at the CAS directory, or point `[cas] dir` at one. A durable
+filesystem produces no warning and no `backing` line. So does a check that cannot answer
+(a non-Linux host, or a `statfs` that fails) — an unreadable filesystem type is not
+evidence of anything, and a guard that spoke every time it failed to look would be tuned
+out by the time it mattered.
+
 **Failure is loud.** In disk mode, a structurally unusable CAS path fails startup with
 an error naming the escape hatches: a dir that resolves inside an allowed project tree
 (refused the same way the state db is), or a file sitting where the store or one of its
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -46,15 +46,6 @@ range); the address string is unchanged.
 
 **Open work that remains:**
 
-- **A loud guard when the CAS is on ephemeral DISK storage.** The ghcr image is a
-  first-class distribution path, and a disk CAS at `$XDG_DATA_HOME/kaibo/cas` in a
-  container with no volume mounted will accept the prompt, **spend the user's provider
-  credits**, verify and write the artifact, then destroy it on exit. Ranked the top
-  remaining risk by the Gemini design pass. **Decided (Amy, 2026-07-30): detect
-  overlay/tmpfs backing and warn severely, then proceed** — not a refusal gated on an
-  acknowledgement flag. The memory-mode SEVERE warn that shipped covers the
-  no-persistence path; this remaining piece is the disk mode whose backing store is
-  secretly ephemeral (statfs the CAS dir, warn on overlay/tmpfs magic).
 - **Cross-agent CAS sharing is DEFERRED** (Amy, 2026-08-03): no tracking of which agent
   created which object, no allow-listing, until a concrete need shows up. The CAS's job
   in one line, hers: "cas should be a way for the kaibo agents to get stuff to the
--- a/src/cas.rs
+++ b/src/cas.rs
@@ -1109,6 +1109,118 @@ fn verify(digest: &Digest, bytes: Vec<u8>) -> Result<Vec<u8>> {
     Ok(bytes)
 }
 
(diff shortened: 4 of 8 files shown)


tobert:
## Cross-family review — DeepSeek: ready to merge

Single-family review, scaled to the change (small read-only guard). The unsafe statfs block, ancestor-walk semantics (symlinks, relative paths, root), Unknown-never-folded, no-double-warn, and the f_type masking were each verified with citations.

---

## Verdict: ready to merge

No correctness or safety issues found. The design is coherent, well-tested, and the tests prove the critical negative cases (Unknown doesn't warn, memory mode doesn't double-warn, Durable is silent). Here is the evidence item by item.

---

### 1. The `unsafe` statfs block is sound

Three distinct things to verify:

**Zeroed struct.** `libc::statfs` is a `#[repr(C)]` struct of integers. All-zero is a valid bit pattern for every field — no padding UB, no uninitialized reads. Same idiom as `store.rs:848`, same SAFETY comment shape at `src/cas.rs:1194-1195`.

```rust
// src/cas.rs:1194-1196
// SAFETY: `buf` is a zeroed, owned `statfs`; `c_path` is a valid NUL-terminated path.
// `statfs` only writes into `buf` and reads the path, and reports failure via `rc`.
let mut buf: libc::statfs = unsafe { std::mem::zeroed() };
```

**Pointer validity.** `c_path` is built by `CString::new(existing.as_os_str().as_bytes())` (`src/cas.rs:1190`). `CString::new` guarantees NUL termination and no interior NUL bytes. The pointer `c_path.as_ptr()` is valid for the duration of the `statfs` call (the `CString` outlives it). The function signature `libc::statfs(*const c_char, *mut statfs)` matches. ✅

**CString handling.** If the path contains an interior NUL (extremely rare on Unix but possible), `CString::new` returns `Err` and the probe returns `Backing::Unknown` — a safe no-answer rather than truncated-path confusion (`src/cas.rs:1190-1192`). ✅

---

### 2. Ancestor-walk semantics are correct

`first_existing_ancestor` (`src/cas.rs:1292-1300`):

```rust
fn first_existing_ancestor(path: &Path) -> Option<&Path> {
    let mut cursor = path;
    loop {
        if cursor.exists() {
            return Some(cursor);
        }
        cursor = cursor.parent()?;
    }
}
```

**Symlinks:** `Path::exists()` follows symlinks (calls `fs::metadata()` internally), so the probe targets the *actual* filesystem the data lands on — exactly what the operator cares about. `Path::parent()` is lexical, but the subsequent `statfs` syscall resolves symlinks at the OS level. The combination is correct: walk up lexically, check existence through symlinks, `statfs` the result. ✅

**Relative paths with nothing existing:** `.parent()` on `""` returns `None`, so the loop returns `None` → `probe_backing` returns `Backing::Unknown`. Tested at `tests/cas.rs:1119-1127`. ✅

**Root:** `/` always exists, so any absolute path always finds an ancestor. Tested at `tests/cas.rs:1135-1142`. ✅

**Production paths are always absolute:** the CAS dir goes through `expand_path` (`src/config.rs:3395-3396`) which resolves `$VAR` and `~`, producing a `PathBuf` from `XDG_DATA_HOME` or `HOME` (always absolute) or a `~`-expanded path. So `probe_backing` never sees a relative path in production. ✅

---

### 3. `Unknown` is never silently treated as ephemeral or durable

The predicate every caller uses is `ephemeral_fs()` (`src/cas.rs:1149-1154`):

```rust
pub fn ephemeral_fs(&self) -> Option<&'static str> {
    match self {
        Backing::Ephemeral { fs } => Some(fs),
        Backing::Durable | Backing::Unknown => None,
    }
}
```

Only `Ephemeral` produces `Some`. `Unknown` and `Durable` both produce `None`. The comment at `src/cas.rs:1138-1142` explicitly justifies this: treating `Unknown` as ephemeral would fire false warnings on every non-Linux install; treating it as durable would *claim* something the probe never established. Tested at `src/server/mod.rs:4978-4992` (both `Durable` and `Unknown` leave no `backing` line in config). ✅

---

### 4. Memory mode never double-warns

`finalize_media_store` (`src/server/mod.rs:1246-1319`) calls `self.backing_probe` **only** in the `CasMode::Disk` arm (line 1297). The `CasMode::Memory` arm (line 1255) issues its own durability warning on its own terms and never touches the backing probe. The `cas_ephemeral_fs` field starts at `None` (`src/server/mod.rs:1190`) and is only written in the Disk arm (`src/server/mod.rs:1300`). Tested at `src/server/mod.rs:4998-5012`. ✅

---

### 5. No write site added; the pin stays at 4

`probe_backing` calls only `statfs` (read-only syscall) and `.exists()` (read-only `stat`). Both are non-mutating. The `EXPECTED_BLESSED_LINES` at `tests/no_write_path.rs:94` remains `4` (1 in `store.rs` + 3 in `cas.rs`). ✅

---

### 6. No spurious fire on common durable filesystems

The match at `src/cas.rs:1202-1207`:

```rust
match (buf.f_type as i64) & 0xFFFF_FFFF {
    0x794C_7630 => Backing::Ephemeral { fs: "overlayfs" },
    0x0102_1994 => Backing::Ephemeral { fs: "tmpfs" },
    0x8584_58F6 => Backing::Ephemeral { fs: "ramfs" },
    _ => Backing::Durable,
}
```

Magic values for common durable filesystems:
- **ext4:** `0xEF53` — catch-all → `Durable` ✅
- **xfs:** `0x58465342` — catch-all → `Durable` ✅
- **btrfs:** `0x9123683E` — catch-all → `Durable` ✅
- **zfs:** `0x2FC12FC1` — catch-all → `Durable` ✅
- **nfs:** `0x6969` — catch-all → `Durable` ✅

Only the three in-memory/overlay magics trigger the `Ephemeral` arm. No false positives. ✅

---

### 7. The `f_type` low-32-bit masking is correct across architectures

The masking `(buf.f_type as i64) & 0xFFFF_FFFF` at `src/cas.rs:1202` is the same pattern used by `store.rs:856` for its network-filesystem guard. The analysis:

- **64-bit Linux:** `f_type` is `__fsword_t` = `long` = `i64`. The kernel fills it with a 32-bit unsigned value → always a non-negative `i64`. Mask is a no-op.
- **32-bit Linux:** `f_type` is `i32`. For magics with bit 31 set (ramfs `0x858458F6` = 2,240,043,254 > i32::MAX), the value is negative → `as i64` sign-extends to `0xFFFFFFFF858458F6` → `& 0xFFFF_FFFF` recovers `0x00000000858458F6` = 2,240,043,254. The match arm `0x8584_58F6` is also `i64` (inferred from the scrutinee type) with value 2,240,043,254. They match. ✅

All three magics recover correctly regardless of pointer width. The approach mirrors the precedent in `store.rs`. ✅

---

### 8. CLI probing is consistent with the handler

`cli.rs:1012-1014` uses the same `probe_backing` function and same `CasMode::Disk` guard, passing the result through the same `render_config_resource` (`src/server/config_resource.rs:643-653`). The only difference is that the handler probes at `finalize_media_store` time while the CLI probes at `kaibo config` time — both target the same resolved `cas.dir`. No divergence. ✅

---

No changes requested. The implementation is correct, safe, well-tested, and the tests prove every negative case that matters.

———
tokens · 795181 in · 16559 out · 12247 reasoning · 756224 cached

PR #128, opened 2026-08-05
chore: accuracy fixes — inert thinking_style flagged, template clarity, the no-CAS-index ruling

tobert:
Small accuracy/polish slate, plus encoding a design ruling.

- **`kaibo://config` now flags an inert `thinking_style`**: `ModelShape::resolve` reads the override only on the Anthropic wire, but the config render showed it as effective on any slot. It now joins the other tunables in `inert_tunables` when written on a non-Anthropic slot. Failing-first tested; CHANGELOG entry.
- **`config.example.toml`**: the five built-in backend stanzas were live TOML while the surrounding comment called them illustrative — now commented like the other optional examples. Added the clarifying clause that `gpt` legally names both a backend and a cast in separate namespaces. (Closes the tracked "config.example.toml clarity" issue — entry deleted.)
- **openai-images**: boundary test pinning `always_b64` as a case-insensitive trimmed *prefix* match (bare `gpt-image`, case, whitespace, and a contains-but-not-prefix non-match) — the #122 review leftover.
- **The no-CAS-index ruling (Amy, 2026-08-05) encoded where the old plan lived**: kaibo will not build a CAS index. The store stays opaque — access is direct by content address; what was created is tracked *with the conversations* (session turns carry the artifact footer, so digests already sit beside the conversation in the store); operator cleanup is plain file mtime. `docs/issues.md`'s retention essay shrinks to that posture plus the one real remaining gap (artifacts orphaned by a cancelled no-session `consult_submit`); `src/cas.rs`'s module doc and `docs/config.md` no longer point at a future index.

Not done, honestly: the "CAST_ENUM_RULES literal list" item from the old session notes does not exist in #122's review comments (checked all three comments plus the reviews API) — left untouched rather than inventing a finding.

Verification: full suite 0 failed, clippy silent, `aws-lc-rs`/`mimalloc` trees empty, `no_write_path` pinned at 4. Implementation by a Claude Sonnet subagent; verification by Claude Fable 5 with Amy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -305,6 +305,14 @@ record. Each later release appends a new section at the top.
   blind to `lane = "batch"` — rendering a batch slot's `effort` and `temperature` as
   effective when batch sends no sampling at all and floors the effort.
 
+- **`kaibo://config` now flags `thinking_style` as inert on a non-Anthropic slot.**
+  The override only moves anything on the Anthropic wire (it picks the adaptive vs.
+  budget tier there); every other wire classifies its thinking style from the model
+  id alone and never reads it, so a `thinking_style` forced on a DeepSeek/Gemini/
+  OpenRouter/generic-OpenAI slot — on the slot itself or inherited from
+  `[defaults]` — used to render as if it were shaping the request when it was a
+  pure no-op.
+
 - **The startup warning and `kaibo://config` can no longer disagree about an effort.**
   They answer from one shared rule now, so a value the batch lane lifts is reported by
   both (it used to be flagged in the resource and stay silent at startup), and the
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -358,30 +358,33 @@ scratch_limit_bytes = 67108864
 # base_url at load (rig fixes those endpoints).
 # ---------------------------------------------------------------------------
 
-# --- The five built-ins, shown with their current defaults. You only need to
-#     list a backend here to CHANGE something; these blocks are illustrative.
-#     Their alias names (claude, google, local, lemonade, gemma, gemma4) are
-#     reserved at both levels — naming a new backend or cast after one is a
-#     loud collision error. ---
+# --- The five built-ins, shown COMMENTED OUT with their current defaults. They
+#     ship in code already — kaibo runs with all five even when this whole file is
+#     absent — so uncomment one only to CHANGE something (a gateway base_url, a
+#     different key file, ...); these blocks are illustrative, not required
+#     configuration to copy verbatim. Their alias names (claude, google, local,
+#     lemonade, gemma, gemma4) are reserved at both levels regardless of whether
+#     you uncomment them — naming a new backend or cast after one is a loud
+#     collision error. ---
 
-[backends.anthropic]                         # alias: claude
-kind = "anthropic"
-api_key_env  = "ANTHROPIC_API_KEY"
-api_key_file = "~/.anthropic-key.txt"
+# [backends.anthropic]                         # alias: claude
+# kind = "anthropic"
+# api_key_env  = "ANTHROPIC_API_KEY"
+# api_key_file = "~/.anthropic-key.txt"
 # Uncomment to dial an Anthropic-Messages-API-compatible gateway/proxy (e.g. a
 # corporate LLM gateway) instead of api.anthropic.com directly. Optional —
 # unlike the openai kind, unset still resolves to rig's built-in endpoint.
 # base_url = "https://llm-gateway.example.internal"
 
-[backends.deepseek]
-kind = "deepseek"
-api_key_env  = "DEEPSEEK_API_KEY"
-api_key_file = "~/.deepseek-key"
+# [backends.deepseek]
+# kind = "deepseek"
+# api_key_env  = "DEEPSEEK_API_KEY"
+# api_key_file = "~/.deepseek-key"
 
-[backends.gemini]                            # alias: google
-kind = "gemini"
-api_key_env  = "GEMINI_API_KEY"
-api_key_file = "~/.gemini-api-key"
+# [backends.gemini]                            # alias: google
+# kind = "gemini"
+# api_key_env  = "GEMINI_API_KEY"
+# api_key_file = "~/.gemini-api-key"
 # Uncomment to dial a Gemini-API-compatible gateway/proxy (e.g. a corporate LLM
 # gateway) instead of generativelanguage.googleapis.com directly. Optional —
 # unset still resolves to rig's built-in endpoint. A HOST ROOT, same as the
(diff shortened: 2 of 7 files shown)


tobert:
## Cross-family review — DeepSeek (glance, scaled to the change): ready to merge

All four items verified with citations. One nit it caught: the PR description said "four" built-in backend stanzas — it's five (description miscount, now corrected in the body; the code and template were always consistent).

---

## Verdict: ready to merge

All four items check out. Here's the evidence for each:

---

### (1) `thinking_style` inert_tunables gating — correct

**The condition** at `src/server/config_resource.rs:476-479`:

```rust
if t.thinking_style != ThinkingStyleOverride::Auto
    && backend.kind.wire() != Some(WireKind::Anthropic)
{
    inert.push("thinking_style");
}
```

This flags thinking_style inert when it's a non-default override AND the wire is not Anthropic. The gating is sound for every case:

- **Anthropic wire** — `Some(WireKind::Anthropic) != Some(WireKind::Anthropic)` is `false`, so not flagged. Correct: `ModelShape::resolve` at `src/consult/shaping.rs:289-293` reads `ovr` in the Anthropic arm to pick adaptive vs budget tier.
- **Non-Anthropic WireKind** (DeepSeek, Gemini, OpenRouter, Openai) — flagged. Correct: `shaping.rs:305-308` — every other arm classifies thinking style from the model id alone and never consults `ovr`.
- **Media kinds** (Stability, OpenAiImages) — `kind.wire()` returns `None` at `src/credentials.rs:233-237`. `None != Some(WireKind::Anthropic)` is `true`, so flagged. Correct: `shaping.rs:280-285` returns early with `ThinkingStyle::None` before reaching any wire match — `ovr` is never consulted.

**Test coverage** is thorough, with three tests in `src/server/config_resource.rs`:

| Test (line) | What it covers |
|---|---|
| `config_render_flags_thinking_style_inert_on_a_non_anthropic_slot` (1088) | Per-slot override on DeepSeek → flagged; same override on Anthropic → NOT flagged |
| `config_render_resolves_inert_tunables_the_way_the_wire_does` (997) | `[defaults].thinking_style = "adaptive"` inherited onto an openai wire → flagged (line 1057 `vec!["effort", "thinking_style"]`) |
| `config_render_flags_written_reasoning_knobs_on_an_image_slot` (925) | thinking_style written on a media (image) slot → flagged (line 973 `vec!["thinking_budget", "effort", "temperature", "thinking_style"]`) |

The one case not explicitly tested — `Auto` on a non-Anthropic slot should NOT be flagged — is handled by the literal `!= Auto` check at line 476; `Auto` is the no-override default and behaves identically to absent on every wire. The three blind spots the 997 test closes (doc comment at 983-995) were the real historical gaps, now all covered.

---

### (2) Template guard tests — cover the fields meaningfully

**Three guard tests**, not one:

- `tests/config.rs:1449` — `the_shipped_example_config_parses`: parses the whole template through the real loader, spot-checks chimera cast. Survives a drifted template that loads but has broken structure.
- `src/config.rs:5393` — `the_shipped_template_documents_every_config_field`: the exhaustive field-name guard. Iterates all 17 raw structs' serde fields (derived via `serde_fields` at 5363-5383 — no hand-kept list), word-boundary matches each against the template text (5403-5409), and then checks every top-level `RawConfig` field appears as a `[section]`/`[section.` header (5449-5463). The word-boundary match at 5404-5408 prevents short-field false positives (`id`, `log`, `lane`, `kind` being substrings of prose).
- `src/server/mod.rs:7802` — `config_example_resource_is_listed_readable_and_valid`: the `kaibo://config/example` MCP resource must be listed, contain the `[backends.anthropic]` anchor, and parse as a valid `Config` (7837-7838).

**Five built-in backend stanzas, not four.** The file header itself says five at `docs/config.example.toml:7-8` ("five built-in BACKENDS"), the block header says five at `:361` ("The five built-ins"), the code registry at `src/config.rs:2228-2234` iterates exactly `[Anthropic, DeepSeek, Gemini, OpenRouter, Openai]`, and the five commented stanzas are at `:370`, `:379`, `:384`, `:408`, `:415`. The PR description said "four" — that's the only discrepancy, and it's a description miscount, not a code defect. The guard tests don't assert a count; they check that every field name appears somewhere in the template text, which a field that *only* appears in a built-in stanza's `kind` line would still satisfy.

**The gpt dual-name clause** at `docs/config.example.toml:483-488` is structurally sound. The enforcement behind it — `register_alias` at `src/config.rs:2164-2181` rejects same-level collisions but permits cross-level name sharing — is tested by `builtin_aliases_register_at_both_levels` at `src/config.rs:4382-4408`. The template's `mentions` closure would flag any missing field regardless of which stanza it appears in.

---

### (3) Prefix-match boundary test — pins the real behavior

The test `always_b64_is_a_case_insensitive_trimmed_prefix_match` at `src/openai_images.rs:618-641` exercises the helper at `:248-250`:

```rust
fn always_b64(model: &str) -> bool {
    model.trim().to_ascii_lowercase().starts_with("gpt-image")
}
```

Its must-match cases (619-624): `"gpt-image"` (bare family name), `"GPT-IMAGE"` (case), `"  gpt-image-1  "` (whitespace), `"Gpt-Image-1-Mini"` (mixed case).

Its must-NOT-match cases (630-635): `"dall-e-3"` (different family), `"sd-gpt-image"` (contains but doesn't lead — prefix vs substring), `"gptimage-1"` (missing hyphen — `starts_with("gpt-image")` fails), `""` (empty string).

This covers every edge: case insensitivity, whitespace trimming, prefix-not-substring, the hyphen boundary, the empty string, and the bare family name with no version suffix. The call site at `:277` uses the result to omit `response_format = "b64_json"` for gpt-image-family models that reject it — the behavior consequence the test guards.

---

### (4) CAS no-index consistency — no contradictory text

The `grep` sweeps across `src/` and `docs/` found zero claims that the CAS is indexable, listable, or has a `du` verb. Every mention of index/list/enumerate is an affirmation of the ruling or a statement that the capability does not exist:

- **The ruling**: `docs/issues.md:73` — "Ruled: no `kaibo cas ls`/`du`/`rm`, and no index to back them."
- **Module doc**: `src/cas.rs:112-113` — "kaibo builds no index and offers no scan verb"; `:139-142` — "no index, ever"; `:143-146` — digests "sit beside the conversation in the persistence store, not in a CAS-side index"; `:154-162` — "Never mounted into kaish" / "The address is the capability here."
- **Config docs**: `docs/config.md:854-857` — "kaibo builds no index and offers no scan verb"; `:868-874` — "the CAS is not mounted into kaish and no cast-facing tool reads it"; `:906-911` — the opt-in `max_bytes` walk is "an O(objects) walk" that "builds no index."
- **Template**: `docs/config.example.toml:291-296` — "NO GC, on purpose … Sweeping the store to survey it does not scale and is not what this is for"; `:298-304` — the `max_bytes` walk is an "opt-in" cost, not an index.
- **Devlog**: `docs/devlog.md:224-230` — "non-enumerable is the property to defend, not read-only."
- **API surface**: CAS has no list/index/scan method; `read_cas` takes only digest + offset/length; CLI has no `cas` subcommand; `CLIENT_ONLY_VERBS` at `src/consult/engine.rs:6442` keeps retrieval operator-surface only.

The "Someday" egress design note at `docs/issues.md:90-130` explicitly preserves the no-enumeration property (line 130: "the model team still must not enumerate"). No contradiction anywhere.

———
tokens · 1167138 in · 29466 out · 14721 reasoning · 1050496 cached

PR #130, opened 2026-08-07
feat(deliberate): keep the dossier, and let a second synth reuse it

tobert:
`deliberate` builds a dossier — minutes of live explorer sweep, and on one measured run a 652k-token sweep over three trees — then hands it to an offline synth and forgets it. Two things were impossible: **peek** (audit what the synth actually received when its answer looks thin) and **reuse** (ask a second cast the same question over the same evidence without paying for the sweep again).

Amy, from the kaish session: *"I want to have the dossiers tossed into cas so we can peek at them or reuse them sometimes."*

## What lands

- **Keep.** The finished dossier goes into the media CAS before stage 2 spends anything, so a deliberation that fails, times out, or comes back thin still leaves its evidence readable. The address rides the ack, the answer footer, and — on the batch lane — the persisted handle.
- **Reuse.** `deliberate` takes a `dossier` digest (bare or as the `kaibo://cas/<digest>` URI kaibo printed). No explorer runs; the call costs the synth alone. That is also the fair way to compare two synths: byte-identical evidence, so only the model differs.

## Decisions

- **`[cas] enabled`, not `[artifacts] enabled`.** The artifacts key exists for the one surface where a *model* decides bytes become durable. A dossier is kaibo's own byproduct of a call the operator already made — no model chose the content, the label, or the save.
- **Keeping never fails the call.** A refused write costs the record, never the deliberation — loud on the operator's log, silent to the model team.
- **Explorer arguments on a reuse call are refused, not ignored.** `attach` especially: a caller who attached a file, got a deliberation, and was never told the file went nowhere is the silent-wrong we don't ship. Checked before anything resolves. The *synth* overrides stay live — retargeting the model that reads the evidence is the point.
- **Any text artifact may be reused,** not only something `deliberate` wrote. The caller is the operator's proxy handing kaibo its own stored evidence. Binary is refused by name; so is a digest that resolves to nothing.
- **The batch lane now persists its handle at all** — it never did. That lane runs for hours, so a restart before collection is ordinary, and the label carries the dossier's address so `job_list` still leads back to the evidence.

## Invariants

No new write site: the bytes reach disk through `MediaStore::put`, the surface `generate` and `save_artifact` already use. `tests/no_write_path.rs` stays pinned at four blessed lines. The CAS is still never mounted into kaish, and the inner model team still cannot read, list, or probe it — `dossier` is an *operator* argument on an MCP tool.

Advertisement gating is unchanged: `deliberate` still appears only where a cast can staff the full job, even though a synth-only offline cast can now serve a reuse call. Recorded in `docs/issues.md` as a deliberate non-fix.

## Tests

Eight new unit tests in `src/server/dossier.rs` (verbatim keep + sidecar authorship, load round-trip in all three reference spellings, each refusal by name, the bounded single-line label, both render surfaces) plus a handler test that a batch deliberation's handle and its dossier address survive a simulated restart. Both new paths were sabotage-checked — breaking the keep and breaking the label each fail their test. Full suite green (1032 passing), clippy clean.

Docs: CHANGELOG, the config guide's CAS producer table (dossiers are usually the bulk of the store, and an operator sizing `max_bytes` needs to know), and two devlog entries — today's, plus the media arc's (#125–#129), which shipped without its story.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,23 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **`deliberate` keeps its dossier, and can reuse one.** The explorer sweep a
+  deliberation is built on is the expensive half — it can run hundreds of thousands
+  of tokens — and it used to be consumed invisibly: no way to see what the offline
+  synth actually reasoned over, and no way to ask a second model the same question
+  without paying for the sweep again. kaibo now stores the finished dossier in its
+  media CAS and names it in the reply (and in the answer when the deliberation
+  lands), so you can `read_cas` it. Pass that digest back as `dossier` and no
+  explorer runs at all: the same evidence goes to whichever cast you point at, for
+  the price of one synth — which is also the fair way to compare two synths, since
+  only the model differs. Any stored text artifact can serve as a dossier, not only
+  one `deliberate` wrote. It rides the CAS's own switch (`[cas] enabled`) with no
+  new key to turn on, a store that refuses the write costs you the record and never
+  the deliberation, and the explorer arguments (`attach`, the explorer overrides,
+  `explorer_max_turns`) are refused on a reuse call rather than silently ignored.
+  Batch deliberations now persist their handle too, with the dossier's address on
+  it, so `job_list` still leads back to the evidence after a restart.
+
 - **kaibo warns when the media CAS is on a disk that will not survive the
   container.** A container with no volume mounted looks exactly like durable disk
   mode: persistence comes up, the store opens, artifacts write and read back — and
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -283,6 +283,11 @@ scratch_limit_bytes = 67108864
 # entirely and un-advertises every tool that needs it (the startup log and
 # kaibo://config's [cas] section say which state you are in, and why).
 #
+# Three tools write here: `generate` (rendered images), `save_artifact` (bulk text a
+# model saved — see [artifacts]), and `deliberate`, which keeps every dossier it builds
+# so you can read it back or hand it to a second synth. Dossiers need no extra key and
+# are usually the bulk of the store; see kaibo://config/guide.
+#
 # It lives under the XDG *data* dir, deliberately NOT beside the state db: the session
 # store is reconstructible bookkeeping, while these are artifacts you spent real provider
 # credits on and may never be able to reproduce (the seed is recorded in a sidecar, but
--- a/docs/config.md
+++ b/docs/config.md
@@ -821,9 +821,25 @@ explicit. Every other open failure stays fatal.
 ## Media CAS: `[cas]`
 
 **On by default, and it follows persistence.** The CAS is the content-addressed store an
-artifact-producing tool writes into. Two produce today: `generate` (images a provider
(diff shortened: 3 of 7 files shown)


tobert:
## Cross-family review + live validation

Two reviews, both aimed at the branch as it stands (no diff supplied — holistic read of the current code).

### Gemini 3 Pro (`gemini-deliberate`, batch lane, 697k-token dossier)

> The `dossier-cas` PR introduces a structurally sound, safely isolated implementation of dossier reuse that completely respects kaibo's read-only invariants. … The code is ready as-is.

On the question I cared most about — whether reuse opens a cross-project hole — it held that placing retrieval on the operator surface and bypassing the sandboxed explorer means no read-scope escalation: a digest is an unguessable content hash supplied by the operator's proxy, and the offline synth receives text rather than a channel to the store. It also confirmed the keep path "fails open" honestly, the batch label/upsert mirror `batch_submit` (TTL pruning included), and that nothing downstream assumes an explorer ran.

One assumption it flagged: that `attach` is the only image-bearing argument on `deliberate`, so refusing it is enough. Checked — `DeliberateInput` has no other file or image parameter.

### DeepSeek (`deepseek`, agentic consult)

Invariants **clean**: no `std::fs` in the new module, every byte through `MediaStore::put`, blessed-line count still 4, no new surface for the inner model team, and `Digest::from_hex` keeps a traversal-shaped string from ever becoming a path. Error messages judged correct *for their audience* — `load_dossier`'s refusals reach the operator's proxy (which already reads `kaibo://config`), unlike `save_artifact`'s deliberately sanitized model-facing errors.

It found the real problem in the tests, naming three gaps. Two are now closed in `337d390`, both sabotage-checked:

- **The handler's reuse wiring was untested** — `deliberate` takes a live `Peer`, so the whole handler was unreachable offline. Split into `deliberate_call` taking the progress sink; the new test drives the reuse road end to end and proves the synth receives the *stored* bytes and the ack doesn't claim a sweep that never ran. (Faking an explorer on that path fails it.)
- **The stitch→keep order was unpinned** — stitching *appends* the sweep's routed text, so keeping first would store a dossier missing evidence the synth then reasons over. `stitch_and_keep` composes the two, and the existing sweep test pins it. (Inverting the order fails it.)
- **Left open, deliberately:** the direct lane's reused-dossier ack. That lane builds a real synth arm and spawns a job, so an offline test would only re-pin the two-branch sentence the render tests already cover.

### Live validation (branch binary driven over stdio; the running MCP server was not touched)

| step | result |
|---|---|
| build a dossier | kept at `kaibo://cas/dec3f6…895a`, 1594 bytes, ack names it |
| `read_cas` it | metadata-first, `text/markdown`, label `deliberate dossier — In src/server/dossier.rs, what does keep_dossier record…`, real disk path |
| reuse on **`gemini-batch`** (a synth-only cast, no explorer slot) | `Dossier reused (no explorer ran)` — submitted to a cast that could not have built it |
| reuse + `attach` | refused: "…`attach` would do nothing. Drop it to reuse the stored dossier, or drop `dossier` to sweep for a fresh one." |
| reuse with a bad digest | refused: "digest must be exactly 64 lowercase hex characters…" |

The same reuse leg on `fable` reached Anthropic's batch submit and bounced on an unfunded key — so the synth-only-cast path is proven up to the provider call there, and end to end on Gemini.

PR #131, opened 2026-08-07
feat(cli): kaibo deliberate + batch cancel (closes #82's open question)

### LINKED ISSUE #82: CLI: wire up consult_submit/job_wait and deliberate's direct lane

tobert:
Amy asked why `deliberate` isn't on the CLI: *"I think we built deliberate around the same time as cli mode and it just wasn't ready. so maybe the easier fix is to finish exposing kaibo features on cli."*

The history: `deliberate` shipped 2026-07-02 (`4f2f9ac`), the CLI front door 2026-07-17 (`1b41129`, consult+config first). It was already built and simply never got its turn.

**#82 records a reason, but only for the direct lane** — that lane's `job-N` is held in the running server's memory, so a one-shot CLI process would submit, exit, and take the job with it. The **batch lane never had a reason**: its handle is durable at the provider, the same property that gave `batch submit` a CLI story on day one.

## What lands

- **`kaibo deliberate`**, both lanes. Batch prints the durable handle on stdout and nothing else, so `HANDLE=$(kaibo deliberate …)` works and `kaibo batch get "$HANDLE"` collects it later — across reboots. Direct runs the long local completion in the **foreground** and prints the answer: the process *is* the job. That's Amy's call, and it answers #82's "candidate directions, not yet decided" with no daemon and no durable job store. It says on stderr that it's waiting, because a silent terminal for an hour is hostile.
- **`--dossier`** on the CLI, so a dossier built over MCP is reusable from the shell — both front doors share one store.
- **`kaibo batch cancel`** — the CLI could submit and list but never stop. The message says "requested", because a provider finishes what's already in flight and those items still cost what they cost.

## Shape

- `cas::open_media_store` now holds the three-state CAS decision (off / disk / memory-with-a-warning) that lived inside `finalize_media_store`. The CLI needs a store of its own to keep a dossier, and two copies of that decision is how two front doors start disagreeing about what "durable" means. MCP behavior is unchanged.
- `inert_explorer_args` reads its arguments through a borrowed view, so the MCP input struct and the CLI flags share **one** refusal instead of drifting. `--attach` with `--dossier` is refused by name on both roads.
- The `backend/provider-id` handle parser is one function now instead of the second copy `cancel` would have made.

## Tests + live validation

Five new CLI tests on the repo's existing templates (parse, both `deliberate` usage refusals, the shared handle split incl. Gemini's `batches/<id>` form, an unknown backend on cancel). Full suite green, clippy clean.

Live, against real providers:

```
$ HANDLE=$(kaibo deliberate "…" --cast gemini-batch --dossier kaibo://cas/dec3f6…)
kaibo: dossier reused (no explorer ran), handed to the batch lane on cast `gemini-batch`
       (synth `gemini-pro-latest`) — `kaibo batch get gemini/batches/mkxz9…`
$ kaibo batch cancel "$HANDLE"
kaibo: requested cancellation of `gemini/batches/mkxz9…`
$ kaibo batch list          # → BATCH_STATE_CANCELLED
```

That reused **this morning's dossier**, built over MCP by `gemini-deliberate`, on `gemini-batch` — a synth-only cast that could not have built it. Refusals verified through the binary with their exit codes (2 for each).

## Follow-up

`kaibo generate` and `kaibo cas read` are the remaining parity gaps, landing next on this bootstrap. `consult_submit`/`job_wait` stay MCP-only by design — their value is work continuing inside a session that outlives the call.

Closes #82.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,18 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **`kaibo deliberate` on the command line, and `kaibo batch cancel`.** Deliberate was
+  MCP-only for no recorded reason on its batch lane, and for a good one on its direct
+  lane (that lane's handle lives in the running server's memory, so a one-shot process
+  would submit, exit, and take the job with it — [#82](https://github.com/tobert/kaibo/issues/82)). Both are answered now. A batch
+  cast prints the durable handle and exits, so `HANDLE=$(kaibo deliberate …)` then
+  `kaibo batch get "$HANDLE"` works across reboots; a direct cast runs the long local
+  completion in the foreground and prints the answer, telling you on stderr that it is
+  waiting. `--dossier` reuses a kept dossier with no sweep — including one built over
+  MCP, since both front doors share the same store. And the CLI could submit and list
+  batches but never stop one; `kaibo batch cancel <handle>` closes that, wording the
+  result as the request it is, since a provider finishes what is already in flight.
+
 - **`deliberate` keeps its dossier, and can reuse one.** The explorer sweep a
   deliberation is built on is the expensive half — it can run hundreds of thousands
   of tokens — and it used to be consumed invisibly: no way to see what the offline
--- a/README.md
+++ b/README.md
@@ -259,8 +259,9 @@ human at a terminal can drive kaibo directly:
 | `consult` | `kaibo consult "question" [--cast … --attach … --session … --json]` |
 | `oneshot` | `kaibo oneshot "prompt" [--attach … --json]` (context also via stdin: `oneshot "review this" < diff.txt`) |
 | `explore` | `kaibo explore "question" [--cast … --json]` |
+| `deliberate` | `kaibo deliberate "question" [--cast … --attach … --dossier … --json]` |
 | `run_kaish` | `kaibo kaish -c 'script'` |
-| `batch_submit`, `job_get`/`job_list` (batch handles) | `kaibo batch submit \| get \| list` |
+| `batch_submit`, `job_get`/`job_list`/`job_cancel` (batch handles) | `kaibo batch submit \| get \| list \| cancel` |
 | `kaibo://config` resource | `kaibo config` |
 | `kaibo://config/example` resource | `kaibo example-config` |
 | `configure` prompt | `kaibo configure [goal]` |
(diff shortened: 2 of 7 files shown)


tobert:
## Cross-family review (DeepSeek, agentic consult over the branch — no diff)

Verdict on the two things I most wanted checked:

**The CLI and MCP roads are semantically equivalent**, verified point by point with citations: lane-capture-before-overrides, the inert-args refusal (one shared `ExplorerArgs` view), the offline-cast requirement, prompt resolution, dossier load, synth caps resolved key-free before the sweep, sink construction with the empty dedup seed, `stitch_and_keep`, batch item + handle persistence, and the provenance/usage accounting. One non-semantic divergence noted and accepted: the MCP road emits `PhaseStarted`/`PhaseFinished` around the dossier sweep and a tracing span; the CLI emits neither, so a CLI user misses two liveness lines. Exploration is identical.

**The `cas::open_media_store` extraction is behavior-preserving** — every mode, warning, and the fatal disk-open posture accounted for, and the in-memory placeholder from `KaiboHandler::new` is replaced (not doubled) before anything is stored.

Exit codes: all correct against 2/3/4. It flagged one deliberate *improvement* over the MCP road — a `submitter` construction failure is `setup` (3) here where the handler says `invalid_params`, since the backend resolved fine and the problem is environmental.

Direct lane: correctly bounded by `request_timeout + 60s`, cannot hang silently (it announces on stderr before blocking).

### Fixed in `631e02f`

- **Lane-capture order had no test.** The CLI inlines the order the MCP road extracted *because* it's fragile — reading `synth_lane()` after an override sees a laneless slot and refuses every deliberate-capable cast as "no offline synth". Now pinned with a fixture cast + `--synth-model`, and sabotage-checked: reordering produces exactly that refusal.
- **`batch cancel` on a batch-incapable backend** failed deep in poller construction as a setup error. Now a usage error naming the batch-capable kinds, matching `batch list --backend`.

### Left open, deliberately

- **No CLI test for the direct-lane deadline derivation.** The derivation is a pure function already tested on the MCP side, and the CLI's use is one line no offline test can distinguish from the alternative — such a test would assert its own mock.
- **No provider round-trip test** for the batch success path (the handle reaching stdout). The CLI calls `crate::batch::{submitter, poller}` as free functions and has no `BatchProviderFactory` seam like the server's, so every CLI behavior test stops at the pre-flight refusal. That's pre-existing structure, not new here — worth its own change if we want the CLI's provider paths covered.

PR #132, opened 2026-08-07
feat(cli): kaibo generate + kaibo cas read (completes CLI parity)

tobert:
**Stacked on #131** (base `cli-parity`) — merge that first; this diff is only the media half.

The last two parity gaps. With these, every MCP tool has a CLI form except `consult_submit`/`job_wait`, which stay MCP-only on purpose: their value is work continuing inside a session that outlives the call.

## `kaibo generate`

Renders through the cast's `image` slot into kaibo's own store, prints the digests. A *deferred* provider job polls in the **foreground** — the same answer #131's direct lane got, for the same reason: there's no server here to hold a `job-N`. `--field k=v` is repeatable, and the value's own JSON syntax carries its type, so `--field n=2` sends a number and `--field size=1024x1024` a string, with nothing extra to learn.

## `kaibo cas read`

This reverses one line of `docs/config.md` — *"There is no `kaibo cas` CLI command"* — on Amy's word: *"it'll be handy, and it doesn't weaken our story."* That line was about **reach**, not opacity. What doesn't change: `read` is the only verb, and no listing, usage report, or delete can exist without an index the store deliberately doesn't keep.

The shape came from Amy's push — *"we do know when we might output binary and the caller likely knows they're gonna get binary"*:

| | `read_cas` (MCP) | `kaibo cas read` (CLI) |
|---|---|---|
| metadata | first content block | **stderr** |
| text | bounded window | **stdout** |
| binary | base64 only on an explicit range | **raw bytes on stdout** |

The MCP rules protect a **context window**; a pipe has no such budget. So `kaibo cas read <digest> > arch.png` works with no flag — verified: a 3.96 MB PNG round-trips **byte-identical** against the stored object. A `--raw` flag was designed, built, and deleted once it was clear it was a worse spelling of the default. Both surfaces still share one planner, which gained a `Delivery` mode so its metadata stops claiming *"as a rendered image"* where nothing renders.

Disk mode only, and it says so: an in-memory store is empty in a new process, so reporting "not found" would send someone hunting for a typo in an address that was never the problem.

## The invariant, deliberately and visibly

Serving those bytes is a `write_all` to stdout, which the guard can't distinguish from writing a file. So it's the **fifth blessed line** in `tests/no_write_path.rs`, pinned count 4 → 5. Writing to stdout creates nothing — it's a descriptor the operator's shell already opened and aimed — so this store still has no destination parameter anywhere, and no model can steer it.

**The guard got stricter in the same change.** The teeth test written for the new marker *failed*: blessings were pinned per-`(file, marker)` but not per-**needle**, so pasting the new marker onto an `fs::write(out_path, …)` would have laundered a caller-named destination straight through — and the CLI is exactly where a `--out FILE` flag gets proposed. Needle pinning is back, and `store.rs`/`cas.rs` now declare which calls their markers excuse.

## Tests

Four new CLI tests (field typing incl. first-`=` splitting and the refusals, no-image-slot refusal, non-digest refusal, memory-mode refusal) plus two new teeth tests on the guard. Full suite green, clippy clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,16 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **`kaibo generate` and `kaibo cas read`.** The CLI can now render an image into kaibo's
+  artifact store and read anything back out of it, which completes the MCP-to-CLI parity
+  sweep. `kaibo cas read <digest> > arch.png` gives you the bytes with no flag to
+  remember: metadata goes to stderr and content to stdout, text as text and binary as raw
+  bytes, because the bounded base64 rules the MCP tool follows exist to protect a model's
+  context window and a pipe has no such budget. `read` is the only verb — the store stays
+  opaque and index-free, so there is still no listing, usage report, or delete. A
+  deferred `generate` polls in the foreground rather than handing back a handle nothing
+  could collect.
+
 - **`kaibo deliberate` on the command line, and `kaibo batch cancel`.** Deliberate was
   MCP-only for no recorded reason on its batch lane, and for a good one on its direct
   lane (that lane's handle lives in the running server's memory, so a one-shot process
--- a/README.md
+++ b/README.md
@@ -261,6 +261,8 @@ human at a terminal can drive kaibo directly:
 | `explore` | `kaibo explore "question" [--cast … --json]` |
 | `deliberate` | `kaibo deliberate "question" [--cast … --attach … --dossier … --json]` |
 | `run_kaish` | `kaibo kaish -c 'script'` |
+| `generate` | `kaibo generate "prompt" [--cast … --field k=v --json]` |
+| `read_cas` | `kaibo cas read <digest> [--offset … --length … --json]` |
 | `batch_submit`, `job_get`/`job_list`/`job_cancel` (batch handles) | `kaibo batch submit \| get \| list \| cancel` |
 | `kaibo://config` resource | `kaibo config` |
 | `kaibo://config/example` resource | `kaibo example-config` |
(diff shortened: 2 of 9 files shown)


tobert:
## Style-guide grooming + cross-family review fixes

**`docs/config.md` was too long and too argumentative** (Amy). The guide it should follow is `~/src/kaish/docs/style.md`, which states that *"kaibo and kaish-extras adopt this guide by reference as they evolve"* — kaibo has no copy and nothing in `CLAUDE.md` points at it, so I didn't know it existed. Worth fixing separately; see below.

Applied to the two sections this arc touched: the "context window vs pipe" argument moved out (the devlog carries it), the retrieval table's cells became complete clauses with rule-then-rationale rather than bare fragments, and the dossier paragraph lost its narration. Net −5 lines and it reads as reference now — state the rule, name the default, show the shape.

## DeepSeek review of this branch

**On the guard, which was the question I most wanted answered:** sound, and *strictly* stricter. It walked all five blessed lines against their needle lists (including the `cas.rs` line that legitimately trips two needles at once), confirmed the tree-wide count still pins what it's meant to, and identified the only remaining hole as a deliberate two-change coordination rather than a single oversight. On the substance: blessing a stdout `write_all` is **not** a weakening — the invariant protects the *project*, stdout creates nothing, the command is operator-driven with no model in the loop, and there's no destination parameter anywhere.

Output contract consistent (including `--json` and the case where the planner serves no body); it verified the CLI's independently-recomputed binary range agrees with what the metadata claims, in every branch. Exit codes clean against 2/3/4.

### Fixed in `facb19d`

- **The poll cadence was wrong and the comment was worse.** The CLI polled a deferred generation every 3s while the server polled every 10s, under a comment calling them "the CLI twin of the server's cadence". One constant now.
- **`Delivery::Bytes` had zero coverage** — all 20+ planner tests used `Rendered`. Now pinned that the mode changes the *claim* and never the body.
- **`cas read`'s byte-serving path had no test at all.** The memory-mode refusal fires before digest parsing, so the existing test never reached the code its name described. Now: a disk-mode store, a real object put through the store's own write path, both digest spellings read back, and an absent digest refused — sabotage-checked. The old case survives under an honest name.

### Left open, with reasons

- **`generate`'s success and deferred-poll paths have no test.** The scripted media double (`ScriptedMediaArms`) lives inside `server/mod.rs`'s test module, so it isn't reachable from `cli.rs` tests; covering this means moving it to `test_support.rs`. Worth doing, but it's a test-infrastructure change rather than part of this feature.
- **No `isatty` guard** on binary output. Deliberate: a caller who asked for a PNG by digest and didn't redirect will see garbage, but they won't be surprised about *what they asked for*, and metadata on stderr says what's coming.

## Follow-up worth its own PR

kaibo should point at the style guide it has adopted — a line in `CLAUDE.md` under "Writing for models" naming `kaish/docs/style.md`, so the next person (or model) writing operator docs here finds it before writing 25 lines of argument into a reference document.

PR #133, opened 2026-08-07
docs: adopt kaish's style guide by name, and put the changelog on a diet

tobert:
kaish's `docs/style.md` states that *"kaibo and kaish-extras adopt this guide by reference as they evolve"* — but nothing in kaibo points at it, so the adoption is invisible from inside this repo. That's how I wrote 25 lines of design argument into a config reference yesterday, and why Amy had to catch it by reading the diff.

Amy: *"I want to work style.md into kaibo's CLAUDE.md focusing on help, externally-facing prose, and comments."*

## What lands

Names the guide as the source for **how** our prose reads, then maps its weights onto kaibo's own files:

| Weight | kaibo files |
|---|---|
| Full | tool `description`s and schema param docs; every clap `about`, arg doc, and `--help` line; **every error, refusal, and diagnostic string** |
| Partial | preambles, the kaish cheatsheet, `///` on `pub` items |
| Terms only, one line per bullet | `CHANGELOG.md` |
| Terms only | `README.md`, `docs/*.md` |
| Exempt | `docs/devlog.md` — it tells the story from a point of view, and a story needs a voice |

Three things the guide implies that kaibo had never written down:

- **Error strings are the highest-value prose we own.** A model reads a refusal far more often than a description, and it arrives exactly when the reader is stuck.
- **Published and internal are different surfaces.** A tool description and a clap arg doc ship to a model; a `///` on a private item does not, and that's where mechanism belongs. The test isn't whether a sentence names an internal — it's whether the reader needs it to predict behavior.
- **Rationale in operator docs stops at the clause after the dash.** The argument goes in the devlog; the story goes in the PR.

The operator-docs and agent-facing bullets shed the general rules they now duplicate, keeping only what's specific to kaibo (the 2048-char host truncation, the non-English-first synth roster, the template-vs-guide split).

## The changelog ratchet

New guidance, and the part with a ratchet: **one line per bullet — the rule, and one clause of why.** The narrative already lives in the PR body that becomes the merge commit. Terser over time: nearby paragraph-shaped entries are fair game whenever you add one, and a release is a good moment to sweep the section you're about to retitle.

Applied to today's three entries, which were exactly the shape it rules out — **39 lines → 22**, three paragraphs → eight scannable bullets. A reader scanning for what changed shouldn't have to read an argument.

Full suite green (docs only; `config.md`/`config.example.toml` are unchanged here).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -209,12 +209,37 @@ x86_64-unknown-linux-musl`. Verify the boundary with `cargo tree -i aws-lc-rs`
 
 ## Writing for models
 
-Every prompt, preamble, tool description, and cheatsheet is text some model reads.
-When we touch one, we **re-read the whole block and judge it holistically** — not the
-diff in isolation. The ratchet is *compression* (改善): over time a block should carry
+Every prompt, preamble, tool description, help line, and error string is text some model
+reads. When we touch one, we **re-read the whole block and judge it holistically** — not
+the diff in isolation. The ratchet is *compression* (改善): over time a block should carry
 more impact per token and avoid accreting clauses.
 
-Two audiences are optimized differently:
+**kaibo adopts kaish's `docs/style.md`** — that guide names kaibo as an adopter, and it is
+the source for *how* our prose reads: a small vocabulary, one term per concept, exact
+numbers, the constraint before the hedge, and one clause of rationale after a rule. Read it
+before writing help text or operator docs. Its weights land on kaibo like this:
+
+| Weight | kaibo files |
+|---|---|
+| Full | every tool `description` and schema param doc in `server.rs`; every clap `about`, arg doc, and `--help` line in `cli.rs`; **every error, refusal, and diagnostic string a tool or command returns** |
+| Partial — keep the terms and the published/internal boundary, relax the rest | preambles and the kaish cheatsheet (`consult/prompts.rs`, `kaish_syntax.rs`), and `///` on `pub` items |
+| Terms only, one line per bullet | `CHANGELOG.md` |
+| Terms only | `README.md`, `docs/*.md` |
+| Exempt | `docs/devlog.md` — it tells the story from a point of view, and a story needs a voice |
+
+**Error strings are full weight, and they are the highest-value prose we own.** A model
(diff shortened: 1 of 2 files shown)


tobert:
## Dogfooded on the audience the rules exist for

Sent the new section to **Kimi K2.7** (`or-kimi`) with one instruction: read it as someone whose first language is not English, because that's the class of model the rules protect — our synth roster is mostly DeepSeek, GLM, Qwen, Kimi.

**It found ten idioms in a section whose entire subject is avoiding them**, and two contradictions. That's the review paying for itself; I'd have shipped all of it.

Quoting what it could not decide between:

| phrase | what it could not resolve |
|---|---|
| "the ratchet is compression" | a one-way mechanism, a trend, or a gear |
| "avoid accreting clauses" | creating, attracting, or gluing on |
| "its weights land on kaibo like this" | casual for "apply to" |
| "fair game to cut down" | idiom for "acceptable to target" |
| "sweep the section you are about to retitle" | metaphor for "clean up" |
| "read an argument" | a long justification, or a dispute |
| `weight` / `Full weight` / `surfaces` | terms of art I used but never defined |
| 改善 | "a non-English insertion with no explanation" |

**"surfaces" was the sharpest hit** — kaish's own guide flags that word as *"doing too much work across the corpus"* and says to name the thing. I put it in a heading.

### The two contradictions

1. **Compression vs. a required reason clause.** Both stated as rules, never reconciled. Now resolved explicitly: keep the clause when a reader who knows the reason can predict a case the rule doesn't name; cut it when it only restates the rule; when the source records no reason, state the rule alone rather than inventing one.
2. **"This is the one place where 'keep the why' does not win"** sitting beside a bullet that keeps one clause of why. Now "**at most** one clause of reason", which makes it optional so compression can win.

A second pass confirmed both resolved and the remaining prose readable.

### Also in `9ee6c3c`

The read-only invariant still said the blessed-line count was **four** and named only two files. It's **five** since #132. Corrected, and restated with the strictness that landed there: a marker is pinned to its file *and* to the specific call it excuses, so pasting the CLI's marker onto an `fs::write` to a caller-named path still fails.

PR #135, opened 2026-08-09
Re-draw a turn the provider could not parse, below the loop

tobert:
A single `MalformedFunctionCall` failed the whole phase and discarded the in-flight
investigation, and `consultation_failed` told the caller *"retrying is unlikely to
help"* — wrong for this class. Observed live: a Gemini Flash explorer emitted one
empty function call mid-`deliberate`. Closes the P2 tracker entry "Turn-level retry on
a malformed tool call".

## How the error actually surfaces through rig

This decided the shape, so it is worth recording.

- **Gemini → rig-core.** `finishReason: MALFORMED_FUNCTION_CALL` becomes
  `CompletionError::ResponseError("Gemini stopped with
  finish_reason=MalformedFunctionCall: …")`, raised while *parsing* the response
  (`providers/gemini/completion.rs::function_call_finish_reason_error`, rig-core 0.41).
  Its siblings are `MalformedResponse`, `UnexpectedToolCall`, `MissingThoughtSignature`,
  `TooManyToolCalls`.
- **rig-core → rig-agent.** `AgentRunner::run_model_turn` (`agent/runner.rs`) yields the
  failed `send()` straight out of the driver stream; `run()` maps it to
  `PromptError::CompletionError`.
- **The finding that matters: the transcript is gone.** `PromptError::MaxTurnsError` and
  `PromptError::PromptCancelled` both carry a `chat_history` — which is exactly why
  kaibo can already recover from a turn cap and a `view_image` break.
  `PromptError::CompletionError` carries **nothing**. So by the time `run_phase` sees
  it, every turn of the investigation is unrecoverable.

**Therefore: a phase-level retry was not an option.** It could only mean re-running from
the first prompt — paying for the whole investigation again to reach the state we just
lost, which is the cost the issue was complaining about.

## The retry shape chosen

A `CompletionModel` wrapper, `Retried` (`src/completion_retry.rs`), composed at the two
sites that already build `watched(...)`: `run_phase_logged` (the tool loop) and
`run_completion` (the single-shot lanes). Innermost, so the re-draw resolves before
`Watched` observes.

Why below the loop:

- The `CompletionRequest` at that level **already holds the entire transcript**, so a
  re-draw loses nothing.
- rig never learns the failed draw happened, so the re-draw **spends none of rig's turn
  budget** — the weakness that made rig's own `on_model_turn_finished` +
  `ModelTurnAction::Retry` unsuitable for the empty-answer recovery too.
- It mirrors `Watched`, the wrapper kaibo already uses on this seam.

`MALFORMED_REDRAWS = 2` (three draws worst case). **No delay between draws**, and that
absence is load-bearing: a fumbled tool call and a 429 want opposite treatments, and a
sleep here would quietly become the transport retry/backoff that stays a deliberate
non-goal (the adjacent "Upstream a retry/backoff for rig" entry, updated to say so).

Every re-draw emits a `tracing::warn` carrying `model`, `attempt`, `redraws`, and the
provider's error, so telemetry can count occurrences per model.

## Vocabulary — and what stayed out

Detection is a heuristic on the error text, by the same necessity as `classify_failure`
(the reason arrives inside a formatted message, not a typed variant). One list,
`is_malformed_generation`, now decides **both** the re-draw and the advice a caller
finally reads.

Deliberately excluded, named in the module doc, waiting on a live probe:

- **OpenAI-family "Response did not contain a valid message or tool call"** — plausibly
  the same class, but nothing yet says whether it is a fumble or a refusal, and a wrong
  guess spends three calls on every refusal.
- **Gemini `MissingThoughtSignature`** — if it fires because a replayed history dropped a
  signature, it fires on *every* draw, so retrying would burn the budget on every call
  rather than rescue anything.

## Classification split

`FailureKind::MalformedGeneration`, checked before the transient list (more specific
diagnosis; both point at another call, so precedence only sharpens the advice). Its text
states the exact number of re-draws already spent — a caller who doesn't know that keeps
retrying by hand, and one who does knows to switch families instead. The number is
interpolated from the const, so the sentence cannot become a lie.

Genuine rejections keep the "retrying is unlikely to help" advice, unchanged. The
transient string now reads "kaibo does not **back off and** retry", which stays true now
that one class *is* repeated; README and `docs/config.md` (embedded as
`kaibo://config/guide`) were corrected the same way, and the failure-class table gained
the malformed row plus the empty-answer row it had been missing since 2026-08-02.

## Tests

Failing-first: the engine and render tests were written against the unwired code and
both failed with the real error text before the fix. Re-checked by mutation — setting
`MALFORMED_REDRAWS = 0` fails the wrapper tests.

- `consult::engine::tests::a_malformed_tool_call_mid_loop_is_re_drawn_and_the_investigation_survives`
  — the real loop, real kaish tool, a scripted Gemini malformed error on the turn after
  the tool result. Asserts the third request still carries the gathered evidence.
- `server::render::tests::a_malformed_generation_steers_toward_a_retry_not_away_from_one`
  — three spellings across both lane markers.
- `completion_retry::tests::a_malformed_generation_is_drawn_again_and_the_second_draw_answers`
- `completion_retry::tests::re_drawing_stops_at_the_bound_and_the_provider_error_survives`
- `completion_retry::tests::a_rejected_request_is_never_drawn_again`
- `completion_retry::tests::a_re_draw_carries_the_same_request`
- `completion_retry::tests::the_vocabulary_covers_the_shipped_spellings_and_no_more`

One harness note: a re-draw sends a **byte-identical** request, so "which attempt is
this" is the one thing a content-driven responder cannot read off the request. Those
tests count draws behind a content predicate — the ordering *is* the behavior under
test, not a shortcut around the harness's design. `test_support` gained one builder,
`response_error`; no harness changes.

Gates: `cargo test` full (713 lib + every integration binary) green, `cargo clippy
--all-targets` clean, build clean. `tests/no_write_path.rs` untouched and green — the
new module has no filesystem surface. rustfmt only on the edited files.

## Open questions

- The two excluded spellings above want a live probe before they are added.
- Whether 2 re-draws is the right bound is a guess from one observed incident; the warn
  event is what will answer it.
- A cast whose sampling is near-deterministic will re-draw into the same fumble three
  times. Bounded, but real.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -263,8 +263,8 @@ record. Each later release appends a new section at the top.
 
 ### Fixed
 
-- **One malformed tool call no longer discards the investigation** — kaibo asks the model
-  for that turn again, twice, before failing the call.
+- **One malformed tool call no longer discards the investigation** — kaibo sends that
+  request twice more before failing the call.
 - **A consultation that produced no answer no longer comes back as a successful empty
   one.** A reasoning model's final turn can carry reasoning but no answer text, and
   kaibo would dress that empty string in a provenance footer and return it as success —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -263,6 +263,8 @@ record. Each later release appends a new section at the top.
 
 ### Fixed
 
+- **One malformed tool call no longer discards the investigation** — kaibo asks the model
+  for that turn again, twice, before failing the call.
 - **A consultation that produced no answer no longer comes back as a successful empty
   one.** A reasoning model's final turn can carry reasoning but no answer text, and
   kaibo would dress that empty string in a provenance footer and return it as success —
--- a/README.md
+++ b/README.md
@@ -684,11 +684,9 @@ named as such rather than blamed on the provider. If a provider is reliably slow
 that backend's `request_timeout_secs`. (Automatic retry/backoff belongs in the HTTP layer
 — tracked as an upstream `rig` contribution in [`docs/issues.md`](docs/issues.md).)
 
-One failure *is* repeated, because it isn't a transport failure at all: when the provider
-could not parse the tool call the model generated (Gemini's `MALFORMED_FUNCTION_CALL`),
-kaibo asks the model for that turn again, 2 more draws on the same request. A fumbled
-tool call then costs one extra call rather than the whole investigation. See
-[`docs/config.md`](docs/config.md), "Failure policy".
+One exception: when a provider cannot parse the tool call a model generated, kaibo asks
+that model for the turn twice more — otherwise a single fumbled tool call discards the
+whole investigation.
 
 **What's the cost?** `consult` spends tokens on the provider behind the chosen cast,
 so the cast is what decides. A DeepSeek-class team is inexpensive enough to reach for
--- a/README.md
+++ b/README.md
@@ -670,7 +670,7 @@ Telemetry is off by default and, when on, only opens an *outbound* OTLP connecti
 you configure.
 
 **What happens if a provider is overloaded or down (a 429/529, a reset, a wedged
-backend)?** kaibo does **not** retry — a single completion is bounded by the backend's
+backend)?** kaibo does **not** back off and retry — a single completion is bounded by the backend's
 `request_timeout` (default 15 min, the wall-clock for one model call) and a ≤10s
 `connect_timeout` that fails a dead endpoint fast. When a provider fails, the consult
 returns a **clean tool-result error** (`is_error`) naming the cast and the underlying
@@ -683,6 +683,12 @@ error (auth, bad request) doesn't, since a retry won't help; and a kaibo-side fa
 named as such rather than blamed on the provider. If a provider is reliably slow, raise
 that backend's `request_timeout_secs`. (Automatic retry/backoff belongs in the HTTP layer
 — tracked as an upstream `rig` contribution in [`docs/issues.md`](docs/issues.md).)
+
+One failure *is* repeated, because it isn't a transport failure at all: when the provider
+could not parse the tool call the model generated (Gemini's `MALFORMED_FUNCTION_CALL`),
+kaibo asks the model for that turn again, 2 more draws on the same request. A fumbled
+tool call then costs one extra call rather than the whole investigation. See
+[`docs/config.md`](docs/config.md), "Failure policy".
 
 **What's the cost?** `consult` spends tokens on the provider behind the chosen cast,
 so the cast is what decides. A DeepSeek-class team is inexpensive enough to reach for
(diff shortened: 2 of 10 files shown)


tobert:
## Cross-family review — kaibo `deepseek` cast (holistic, worktree checkout, no diff)

**Recommendation: MERGE.** No defects found. The change is correct, well-tested, and composes cleanly with existing infrastructure. Three minor observations (none blocking), one test gap proportionate to its risk.

### (1) Correctness of the Retried wrapper → PASS
- Re-draw carries the identical request: `src/completion_retry.rs:139` clones the `CompletionRequest`; `src/consult/engine.rs:330-363` proves byte-identical serialization between draws. The transcript lives inside `CompletionRequest`, so it survives every re-draw — the point of wrapping below the loop.
- Bound enforced: loop `1..=redraws` + final un-cloned call = `MALFORMED_REDRAWS + 1 = 3` total; pinned at `completion_retry.rs:264-279` with the provider error text surviving.
- Non-malformed errors are never retried (`:140` guard; `invalid_request_error` costs exactly 1 draw, `:285-294`). No loop risk — bounded `for` over a compile-time constant.
- No delay, documented at `:28-33` — correctly kept distinct from the transport-retry concern in `docs/issues.md`.

### (2) Vocabulary seam → PASS
Five markers (`:70-76`); `"malformed"` appears in no entry of the `TRANSIENT` list (`src/server/render.rs:113-129`), so no transient error slips into the retry class. Classification order (malformed before transient, `render.rs:102-110`) is the right priority. Deliberate exclusions (OpenAI "no valid message", Gemini `MissingThoughtSignature`) are named in the module doc and pinned absent at `:316-317`.

### (3) Composition with Watched → PASS
`watched(retried(...))` at both sites (`engine.rs:1057`, `:1449`) — Retried innermost. A failed draw produces no response, so the completion log holds one record per turn either way; a successful re-draw's finish reason is the one the caller cares about.

### (4) Invariants → PASS
Pure model-wrapping; nothing that would trip `tests/no_write_path.rs`; blessed-line count unchanged at 5.

### (5) Published prose → PASS
The advice string names what happened, why the re-draws are spent, and what to do next, with the count interpolated from `MALFORMED_REDRAWS` (`render.rs:222-238`) so the string cannot quietly become a lie. Docs table row and warn event match.

### (6) Tests → PASS, one gap
Five in-module tests + the `run_phase` integration test (transcript survives the re-draw) + the render steering test all pin load-bearing behavior, failing-first quality. **Gap (non-blocking):** no integration test drives the `oneshot`/`deliberate_direct` single-shot path through a malformed generation — proportionate, since a toolless completion cannot fumble a tool call; close it when a live probe shows the shape.

### Minor observations (non-blocking)
- **A.** `MALFORMED_REDRAWS` naming reads as "extra draws beyond the first" while the loop includes the first; behavior correct, docs clear, tests pin the count.
- **B.** `Retried::new` with `redraws: 0` is degenerate but unreachable today.
- **C.** `stream()` forwards without re-draw logic — honest (documented; kaibo never streams), but would silently lose the re-draw if streaming is ever adopted.

———
tokens · 647381 in · 20318 out · 10584 reasoning · 555648 cached

PR #136, opened 2026-08-09
fix: cross-check batch results against the submitted custom_id set

tobert:
## Summary

None of the three batch result parsers (Anthropic, Gemini, OpenAI) verified that
every submitted `custom_id` came back. A provider that silently dropped an item
rendered "Batch complete — 9 result(s)" after a submit of 10, with no error naming
the gap (`docs/issues.md` P3 entry, from the 2026-08-02 batch error-propagation
audit).

- **Count, not a set.** kaibo assigns every `custom_id` itself as the contiguous
  decimal indices `0..submitted` — `batch_submit`/`deliberate` never take an id
  from the caller — so a bare `i64` column on the batch handle fully determines
  the expected set. A new `submitted_count` column rides a `v1 -> v2` store
  migration (`ALTER TABLE ADD COLUMN`, nullable).
- **Legacy handles stay honest by construction, not by a special case.** The
  `ALTER TABLE` leaves every existing row `NULL` (SQLite does this for free), and
  `cross_check_absentees` treats `None` as "nothing to check" — a handle from
  before this shipped, or with persistence off, renders exactly as it always did.
  No guessed count, no false "complete."
- **The cross-check runs only where a provider's own state promises
  completeness**: Anthropic's `ended` (every terminal reason — errored, canceled,
  expired — still gets a results line, so it's unconditional there), Gemini's
  `BATCH_STATE_SUCCEEDED` specifically, and OpenAI's `completed` specifically. A
  cancelled/expired *partial* result set is expected to be short, and is never
  cross-checked even when a submitted count is on hand — that would misreport a
  known, expected gap as a silent drop. Each "don't check" branch has its own
  test pinning that behavior, not just the "do check" branches.
- **OpenAI's check lives in a new `finalize_openai_answers`**, called once after
  the output+error files are merged — a per-file parse alone can't tell a
  genuinely missing item from one sitting in the *other* file, so checking inside
  `parse_openai_output_jsonl` itself would have been wrong.
- **The synthetic absentee failure is loud and unambiguous**: prefixed `kaibo:`
  and never containing `provider error:`, so it can't be mistaken for a
  provider-reported failure. It names the missing id and the two real next steps
  — re-run `job_get` in case it's still landing, or check the id on the
  provider's own batch dashboard.

## Test plan

- [x] TDD, failing-first: each of the three cross-check sites
  (`parse_results_jsonl`, `parse_gemini_poll`/`parse_gemini_inlined`,
  `finalize_openai_answers`) has a test proving the fix — verified by
  temporarily neutralizing each cross-check call and confirming the new tests
  fail, then restoring.
- [x] Companion tests for Gemini and OpenAI proving a known-partial (cancelled/
  expired) result set is *not* cross-checked even with a submitted count present.
- [x] `store.rs` test proving a pre-v2 row (no `submitted_count`) reads back
  `None`, not a guessed number.
- [x] `cargo test` — full suite green (712 lib tests + all integration suites).
- [x] `cargo clippy --all-targets` — clean.
- [x] `tests/no_write_path.rs` — unaffected, pinned marker count still exact.
- [x] `rustfmt` on touched files only (repo isn't rustfmt-clean repo-wide).

`docs/issues.md`: deleted the shipped tracker entry; kept the two
accepted-as-designed observations from the same audit (Anthropic's no
batch-level `Failed`, Gemini's coarse error summarizer) as a short residual
note so they aren't re-flagged — those were explicitly out of scope for this
fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -265,6 +265,9 @@ record. Each later release appends a new section at the top.
 
 - **One malformed tool call no longer discards the investigation** — kaibo asks the model
   for that turn again, twice, before failing the call.
+- **A batch result no longer under-reports silently** — `job_get`/`job_wait` now flag any
+  submitted `custom_id` a provider never returned as a loud per-item failure.
+
 - **A consultation that produced no answer no longer comes back as a successful empty
   one.** A reasoning model's final turn can carry reasoning but no answer text, and
   kaibo would dress that empty string in a provenance footer and return it as success —
--- a/docs/devlog.md
+++ b/docs/devlog.md
@@ -54,6 +54,56 @@ that will keep retrying by hand, and a caller who does knows to switch families 
 The number comes from the const rather than the prose, so the sentence cannot quietly
 become a lie.
 
+## 2026-08-09 — a count kaibo never checked against itself
+
+The 2026-08-02 batch audit's one real finding, closed: none of the three result parsers
+(`parse_results_jsonl`, `parse_gemini_inlined`, `parse_openai_output_jsonl`) ever asked
+whether the provider sent back everything kaibo submitted. A provider that quietly dropped
+one item still rendered "Batch complete — 9 result(s)" after a submit of 10 — no error, no
+gap named, just a count one short of the truth. The per-item propagation the audit checked
+was sound; this was the one thing nothing checked *against*.
+
+**Count, not a set.** kaibo assigns every `custom_id` itself, as the contiguous decimal
+indices `0..submitted` — `batch_submit` and `deliberate` both build `items` that way, never
+taking an id from the caller. So the expected set is fully determined by one number. Storing
+the ids anyway would be redundant state that could itself drift from the count; a bare `i64`
+column on the batch handle is the whole fix on the persistence side.
+
+**Legacy handles stay honest by construction, not by a special case.** `ALTER TABLE
+batch_handles ADD COLUMN submitted_count INTEGER` leaves every existing row `NULL` — SQLite
+does this for free, no backfill migration to get wrong. `cross_check_absentees` treats
+`None` as "nothing to check", so a handle from before this shipped renders exactly as it
+always did. No guessed count, no false "complete" stamped on a gap kaibo can't actually see.
+
+**The interesting bug I avoided, not the one I was asked to fix.** Gemini's cancelled/expired
+poll branch and OpenAI's `expired`/`cancelled` states both return a *deliberately* partial
+result set — whatever finished before the clock or the cancel caught up. First draft of this
+threaded the submitted count into every terminal branch uniformly, which would have turned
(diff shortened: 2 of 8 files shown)


tobert:
## Cross-family review — kaibo `deepseek` cast (holistic, worktree checkout, no diff)

**Verdict: Merge.** The completeness-gating logic is correct across all three providers, the store migration is safe in both directions, the absentee message is clearly distinguishable from provider-reported errors, and every load-bearing branch has a failing-first test. One low-severity wording observation is the only note worth addressing.

### 1. Completeness gating → correct, no missed terminal state
- **Anthropic `ended` is genuinely unconditional-complete**: it is the only terminal state on that wire, and the results JSONL carries one line per submitted item regardless of item outcome (`succeeded`/`errored`/`canceled`/`expired` lines all present, `src/batch.rs:599-627`) — so the unconditional check at `:631` is safe; a gap is a real drop.
- **Gemini**: only `BATCH_STATE_SUCCEEDED` passes `submitted` through (`:1109`); `CANCELLED`/`FAILED`/`EXPIRED` all skip — and `src/batch.rs:3403-3420` proves a count on hand does not flip the check back on for a cancelled batch.
- **OpenAI**: only `completed` checks (`:1736-1743`); a validation-failed batch short-circuits before the check (`:2007-2016`).
- Unknown states on all three wires are a loud error, so a new provider state cannot silently degrade to "no check."

### 2. Store migration → safe both directions
Fresh DB runs v1+v2; existing v1 DB gets `ALTER TABLE ADD COLUMN` with pre-existing rows `NULL` — exactly the "unknown, don't guess" value (`src/store.rs:963-991`). An older binary opening a v2 database neither reads nor clobbers the column. Single `build_database` path, connect-per-operation, MP-WAL — invariants respected, no new open-site.

### 3. Synthetic absentee failure → distinguishable, style-compliant
Names what happened, that it is kaibo's own count check (not a provider-reported failure), and what to do; pinned against reading like a provider error (`src/batch.rs:2727-2732`). Count math holds when provider-reported failures and absentees coexist.

**Low-severity wording nit**: "Re-run `job_get` in case the result is still landing" is misleading — by the time the check fires the provider has declared the batch finished, so that window has closed. The dashboard half is the actionable path. Consider striking the first clause.

### 4. Invariants → untouched
No new write paths; `EXPECTED_BLESSED_LINES` stays 5.

### 5. Tests → nothing load-bearing unpinned
All six new tests pin their branches (do-check and don't-check per provider, legacy-NULL read-back); the `deliberate` batch lane persists `submitted_count = Some(1)` and the poll side inherits the shared `job_get` path.

### 6. server/mod.rs whitespace → clean
No semantic changes hide in the formatting normalization.

———
tokens · 5899892 in · 45761 out · 21340 reasoning · 5424896 cached

PR #139, opened 2026-08-09
docs: groom issues.md — delete two shipped entries, tighten a third, add one

tobert:
## Summary

Tracker grooming pass on `docs/issues.md` per the delete-when-shipped convention. This is the **bottom layer of a two-PR stack** — an rmcp test-harness PR will stack on top of this branch tip. Do not merge without checking in first.

### Deleted (verified fully shipped)

**1. P3 — `KaishWorker::read_file` is unbounded — stat-then-read growth race.**
The fix shape the entry asked for ("a capped read op on the worker (`read_file_capped(max)`) that refuses past the cap at the VFS layer; both resolvers pass their real ceiling") is fully in place:

- `src/sandbox.rs:544` — `KaishWorker::read_file_capped(&self, path, max_bytes)` bounds the read via `read_range(&path, Some(ReadRange::bytes(0, limit)))`, which `LocalFs` honors with `File::take(limit)` (sandbox.rs:487-490).
- `src/server/resolver.rs:583` stats `meta.len() > remaining` before calling `read_file_capped(canon.clone(), remaining + 1)` at `resolver.rs:609`.
- `src/server/containment.rs:172` stats `meta.len() > read_ceiling` before calling `read_file_capped(canon.clone(), read_cap)` at `containment.rs:213`.
- The old uncapped `read_file` method no longer exists anywhere in `src/` — `grep -rn "fn read_file\b\|\.read_file(" src/` returns nothing.
- Shipped in commit `47303ae` ("fix(attach): bound every raw read so a stat-then-read swap can't OOM"), confirmed via `git log -S"read_file_capped"`.

**2. P3 bullet — `thinking_style` missing from the `inert_tunables` render.**
`src/server/config_resource.rs:491-495`:
```rust
if t.thinking_style != ThinkingStyleOverride::Auto
    && backend.kind.wire() != Some(WireKind::Anthropic)
{
    inert.push("thinking_style");
}
```
Backed by a dedicated test, `config_render_flags_thinking_style_inert_on_a_non_anthropic_slot` (`config_resource.rs:1116`), asserting `vec!["thinking_style"]` on a DeepSeek slot. Shipped in commit `6e7c3e6` ("fix: kaibo://config now flags an inert non-Anthropic thinking_style"), matching the 2026-08-05 PR #128 claim.

### Tightened (skim pass turned up a third stale claim)

**P2 — "Attachment per-file streaming cap + FIFO-hang".** Its first bullet ("Per-file size cap is check-then-read, not streaming... needs a `KaishWorker`/kaish-vfs API that bounds the read") describes the *identical* gap closed by `read_file_capped` above — same fix, same commit. Removed that bullet, kept the still-open FIFO-hang bullet (unrelated: a swapped-in FIFO can still hang the single worker thread), and retitled the entry to reflect that the per-file cap shipped too.

### Added

**P2 — "Hybrid-provider casts as shipped defaults, and a 429-aware explorer story"** (per the coordinator, folded into this PR to avoid a same-file conflict with a second PR). Field report 2026-08-09: `gpt-deliberate`'s built-in explorer (`gpt-5.6-luna`, same family as the `gpt-5.6-sol` synth) self-starved on OpenAI's 200k TPM org ceiling — failed attempts still consume the window, so a same-cast retry can't recover. Live-verified fix: a per-call `explorer_backend`/`explorer_model` override to a Gemini explorer. Confirmed the built-in `gpt-deliberate` cast (`docs/config.example.toml:650-652`) does pin a same-family `gpt` explorer today, and that `explorer_backend`/`explorer_model` CLI overrides already exist (`src/cli.rs`) — so the cross-backend mechanism (`AGENTS.md`'s "freely cross-backend" cast design) is real, the gap is only in the shipped defaults. Names two open threads: (1) should the built-in `gpt` offline/deliberate casts default their explorer cross-provider, since family-match only matters for the synth's answer; (2) a possible 429-aware auto-retarget fallback, flagged against the house no-silent-fallbacks doctrine — any such fallback would need to be loud (named in the provenance footer) and probably opt-in.

## Test plan

- [x] `cargo build` — clean, confirms nothing references the removed doc text
- [x] Quoted the exact code lines justifying each deletion above
- [x] No `CHANGELOG.md` or `docs/devlog.md` entry — tracker grooming is internal, the git log is its record (per `AGENTS.md`)

Co-authored-by: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -253,31 +253,24 @@ otherwise sound. The attacker model stays narrow (a self-attack: a concurrent wr
 caller's own workspace, sub-millisecond window). Worth closing for symmetry when the model
 justifies it, but lower priority than the attachment read that one-shot-followed a symlink.
 
-### Attachment per-file streaming cap + FIFO-hang (DoS, self-attack) — count/total caps shipped
+### Attachment FIFO-hang (DoS, self-attack) — count/total/per-file caps shipped
 Surfaced by the Gemini Pro batch review of the VFS-read change (2026-06-23). These are
 **pre-existing** and all in the self-attack model (the caller owns the request and the
 workspace), so they're DoS, not confidentiality. The cheap batch-level bounds **shipped**
 (`attach::check_attachment_bounds`, wired into `resolve_attachments`): a hard cap on
 attachment *count* (`DEFAULT_MAX_ATTACHMENTS` = 64, checked before canonicalizing the list)
 and a *cumulative* byte budget (`DEFAULT_MAX_TOTAL_BYTES` = 32 MiB, the running total
-checked before each read so a batch of individually-legal files can't sum to an OOM). Two
-remain, both needing more than a pre-check:
-- **Per-file size cap is check-then-read, not streaming.** `resolve_attachments` checks
-  `std::fs::metadata` length against `read_ceiling`, then `worker.read_file` slurps the whole
-  file with **no byte cap** (`Job::Read` deliberately skips the script-output cap). A file
-  swapped for a huge one *after* the metadata check is read whole into a `Vec<u8>` → OOM,
-  bounded now only by the 32 MiB cumulative cap (so the blast radius shrank, but a single
-  swapped file up to that budget still slurps). The real fix is a *streaming* cap in the read
-  path (a `read_file` that takes a limit and aborts past `limit+1`), which needs a
-  `KaishWorker`/kaish-vfs API that bounds the read. (`classify` enforces the per-encoding cap
-  only *after* the full read, too late for memory.)
+checked before each read so a batch of individually-legal files can't sum to an OOM). The
+per-file streaming cap **shipped too**: `KaishWorker::read_file_capped(max)`
+(`sandbox.rs`) bounds the read itself via `read_range`/`File::take`, and both attachment
+resolvers (`server/resolver.rs`, `server/containment.rs`) pass their real per-file ceiling
+— a stat-then-read swap now stops at the cap instead of slurping to OOM. One remains:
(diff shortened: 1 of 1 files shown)


tobert:
## Cross-family verification — kaibo `deepseek` cast

Glance-scale review for a tracker groom: the two code claims not already hand-verified by the coordinator, checked against the branch. **PASS / PASS / PASS — nothing was deleted prematurely.**

1. **`thinking_style` inert flag — PASS.** `src/server/config_resource.rs:491-494` pushes `thinking_style` into `inert` when set on a non-Anthropic wire; `config_render_flags_thinking_style_inert_on_a_non_anthropic_slot` (`:1116-1168`) pins both directions (inert on DeepSeek, effective on Anthropic). The deleted bullet was stale.
2. **Per-file streaming cap — PASS.** `read_file_capped` (`src/sandbox.rs:544-560`) is the *only* read method on the worker; the handler applies a true byte-range read (`ReadRange::bytes(0, limit)`, `:472-491`). All five call sites pass a concrete ceiling (both resolvers, `sweep_attach.rs:417`, `view_image.rs:144`). No uncapped whole-file read survives.
3. **Rewritten FIFO-hang entry — PASS, still accurate.** `request_timeout` lives only on the HTTP client (`src/tls.rs:55-62`); all three attachment-resolution paths run before any HTTP client exists, and no timeout wraps any `read_file_capped` site. The entry's remaining ask is correct.

One crumb found: the doc-comment at `sandbox.rs:394` still references `KaishWorker::read_file` — the method is `read_file_capped` now. Internal comment, one word; riding the next PR that touches the area.

———
tokens · 1349729 in · 16134 out · 7691 reasoning · 1259136 cached

PR #140, opened 2026-08-09
test: drive kaibo's MCP surface from an in-tree rust client — and fix the gate it found broken

tobert:
kaibo's MCP surface — the handshake, `tools/list`, `tools/call`, `resources/read` — had no
in-tree coverage. Every test reaches `KaiboHandler` in process. This adds a Rust harness
that spawns the real binary and drives it as a client, and it immediately found a live bug
in the shipped wire.

## Why Rust, not Python

The plan on the tracker was an out-of-tree `mcp_drive.py`. Amy's ruling killed it: *"I
don't want to check in python. Let's consider how to drive a test with rust tests. rmcp can
be a server."* kaibo already depends on rmcp for the server half, so the client half is a
dev-dependency away.

## The bug this found

`#[tool_handler]`'s `router` attribute has a default, and the default changed under us:

| rmcp | default `router` |
|---|---|
| 0.16 | `self.tool_router` — the handler's own, runtime-gated router |
| 1.x, 2.x, 3.x | `Self::tool_router()` — a fresh router from the compile-time `#[tool]` set |

kaibo wrote the bare attribute. The 0.16 → 3.0.0-beta.5 upgrade (569d872, 2026-07-28,
"No tool behavior or wire shape changed") therefore swapped the gated router for an ungated
one, silently:

- A keyless server advertised **14** tools while its own startup log said, three times, that
  `batch_submit` / `deliberate` / `generate` had no cast to staff them. `kaibo://config`
  reported the correct 11 in the same session.
- `tools/call` reads the same router, so **`--no-run-kaish` did not block a `run_kaish`
  call**. Confirmed by hand before believing it: the tool vanished from `tools/list` and
  then ran the script. The `--no-<tool>` flags are the operator's capability surface, and
  for twelve days they were advisory on the wire.
- The injected `cast` enums and the `anthropic/alwaysLoad` pin on `consult` are stamped onto
  the instance's routes, so those were gone too.

Fix: `#[tool_handler(router = self.tool_router)]`, with a comment saying why it is not
decoration.

**No in-process test could have caught this.** `advertised_tools()` reads the gated router
and answers correctly whichever router the wire is using — the existing gating tests were
honest and useless at the same time. That is the argument for the new file.

## Harness shape, and why

- **One file, `tests/mcp_stdio.rs`** — harness plus tests. Extending it is adding a `#[tokio::test]`;
  promoting the `Server` struct to a shared `tests/…/mod.rs` is a one-move refactor the day a
  second file wants it, and until then a shared module only buys dead-code allowances.
- **Hermetic by clearing, not by overriding.** `env_clear()` first, then `HOME` and every XDG
  root kaibo reads (`XDG_CONFIG_HOME` → config.toml, `XDG_STATE_HOME` → the state db,
  `XDG_DATA_HOME` → the media CAS) pointed at per-test temp dirs. A developer's provider key
  or a stray `KAIBO_*` would otherwise decide the advertised set, which is the thing under
  test. kaibo runs no external command, so it needs nothing else from the environment. A
  missing `config.toml` is a non-error by design, so an empty config dir means "built-ins".
- **`--root <tempdir>`**, which also drops the invocation cwd from the allowed set — so the
  kaibo checkout the tests run from is not readable by the server under test, and a
  successful read can only have come from the temp project.
- **One process per test**, own temp dirs. No shared state, no ordering, parallel-safe. Six
  spawns, whole file runs in well under a second.
- **Every request bounded** by a 60s timeout with a named message, so a hung server fails its
  own test instead of wedging the suite. Generous on purpose: a hang detector, never a
  performance assertion.
- **Pin the whole advertised set**, not "contains `run_kaish`". The failure here was a set
  that got *bigger*; only an equality assertion notices that.

## What a keyless server actually does

It starts and serves — it does not hit the nothing-advertised refusal. The built-in
`openai-local` backend needs no key, so the casts it staffs keep `consult` /
`consult_submit` / `explore` / `oneshot` (plus the job verbs that collect handles)
advertised; kaibo checks that a cast can be *staffed*, not that a local server is listening.
`run_kaish`, `list_models`, `read_cas` need no cast. `batch_submit`, `deliberate`, and
`generate` need cast shapes no built-in provides, so their routes are dropped. That set is
pinned in `KEYLESS_TOOLS`.

## Tests

- `handshake_names_kaibo_and_ships_read_only_instructions` — server info is `kaibo` at the
  crate version, tools+resources capabilities present, instructions carry `read-only`,
  `consult`, `cast`.
- `a_keyless_server_advertises_exactly_the_castless_surface` — the pinned 11.
- `a_gated_tool_is_neither_advertised_nor_callable` — `--no-run-kaish`: absent from
  `tools/list` **and** refused at `tools/call`. This is the one that caught the bug.
- `run_kaish_reads_a_file_under_the_served_root` — `cat -n Cargo.toml` returns the temp
  project's contents.
- `run_kaish_refuses_every_write_and_the_project_is_untouched` — `echo >`, `mkdir`, `touch`,
  `rm` each refused as read-only, and the filesystem is checked afterwards. The read-only
  invariant, probed through the shipped MCP surface for the first time.
- `the_config_resources_round_trip` — `kaibo://config` and `kaibo://config/example`.
- `consult_carries_the_always_load_pin_and_no_other_tool_does` — the
  `anthropic/alwaysLoad` `_meta` stamp reaches the wire on `consult`, and only there.
  Added from the review: routing and meta injection are separate steps on the same router,
  so a change that keeps the routes and drops the stamp would have passed everything else.

## Teeth (proven, then restored)

- Revert to bare `#[tool_handler]` → `a_keyless_server_advertises_exactly_the_castless_surface`
  fails with 14 vs 11, and `a_gated_tool_is_neither_advertised_nor_callable` fails.
- Mount the project with `LocalFs::new` instead of `LocalFs::read_only` →
  `run_kaish_refuses_every_write_and_the_project_is_untouched` fails on `echo x > written.txt`
  with `exit: 0`.
- Point the `_meta` injection at a name no route has →
  `consult_carries_the_always_load_pin_and_no_other_tool_does` fails on the missing `_meta`.

## Review

DeepSeek hard look: MERGE, fix confirmed, harness judged hermetic and flake-free, and the
Fixed-bucket framing confirmed honest — no structural invariant was breached, the exposure
was the operator capability boundary only. It found the `_meta` gap above, and tripped on a
devlog sentence that compressed what the operator asked for with what the wire did (the tool
was listed *and* callable, not dropped-then-run); both are addressed in `0c01d52`.

## Dependency guard

`rmcp` dev-dependency with `client` + `transport-child-process` (pulls `tokio/process` and
`process-wrap`, both pure Rust; no reqwest, no TLS). Under resolver v2 dev-dep features do
not reach the shipped binary. Verified on the branch: `cargo tree -i aws-lc-rs`,
`-i mimalloc`, and `-i openssl-sys` all report no matching packages.

## Also here

- `src/sandbox.rs`: an internal doc-comment pointed at `KaishWorker::read_file`; the method is
  `read_file_capped`. Crumb from the DeepSeek review of #139.
- `docs/devlog.md`: dated entry for the arc.
- `CHANGELOG.md`: one Fixed bullet — the gates work again over MCP.

## Gates

Full `cargo test` green (719 unit + every integration suite, 0 failures), `cargo clippy
--all-targets` clean, `tests/no_write_path.rs` green with the blessed count unchanged,
rustfmt run on the new file only (the tree is not rustfmt-clean repo-wide).

## Left open

A gated tool is refused with rmcp's own `-32602: tool not found`, which does not say *why*.
Per the house rule on refusals, kaibo could answer "`run_kaish` is disabled on this server
(`--no-run-kaish`)" — that means overriding `call_tool` rather than letting the macro
generate it, so it is its own change, not a rider on this one.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -136,6 +136,12 @@ project and cannot run external commands.
 - **TDD.** Tests that can and will fail. The sandbox boundary gets failing-first
   tests — and we prove they actually work (e.g. mount the project writable with
   `LocalFs::new` instead of `read_only`, watch the write-denial tests fail).
+- **Tests start from a blank environment.** A test that spawns the kaibo binary
+  clears the child's env (`env_clear()`) and rebuilds only what it means — temp
+  `HOME` and XDG roots — never an enumerated `env_remove` list, which rots the day
+  a new variable ships. In-process tests inject env through `load_with`'s map, not
+  the process environment. So a developer's keys and `KAIBO_*` overrides can never
+  change what a test observes (`tests/mcp_stdio.rs` is the model).
 - **`docs/sandbox-probes.md` is the read-only/containment audit runbook.** The
   `cargo test` suites are the continuous guard; that doc is how we *live-test* the
   shipped boundary now and then (write/external/read-escape/env/path batteries via
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ record. Each later release appends a new section at the top.
 
 ### Fixed
 
+- **`--no-<tool>` gates work again over MCP.** Since the rmcp 3.0 upgrade a disabled or
+  unstaffable tool was still listed and still callable; it is now neither.
 - **`--include-report`'s help described the wrong stream.** It said the report is appended
   below the answer; it goes to stderr, so a script that piped stdout lost it silently.
 - **`--cas-max-bytes` was documented as a "soft cap".** A write past it is refused and
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2145,6 +2145,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "nix"
+version = "0.31.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf20d2fde8ff38632c426f1165ed7436270b44f199fc55284c38276f9db47c3d"
+dependencies = [
+ "bitflags 2.12.1",
+ "cfg-if",
+ "cfg_aliases",
+ "libc",
+]
+
+[[package]]
 name = "nom"
 version = "7.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2552,6 +2564,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "process-wrap"
+version = "9.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e842efad9119158434d193c6682e2ebee4b44d6ad801d7b349623b3f57cdf55"
+dependencies = [
+ "futures",
+ "indexmap",
+ "nix",
+ "tokio",
+ "tracing",
+ "windows",
+]
+
+[[package]]
 name = "prost"
 version = "0.14.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2891,12 +2917,14 @@ dependencies = [
  "futures",
  "pastey",
  "pin-project-lite",
+ "process-wrap",
  "rmcp-macros",
  "schemars",
  "serde",
  "serde_json",
  "thiserror",
  "tokio",
+ "tokio-stream",
  "tokio-util",
  "tracing",
  "uuid",
@@ -3310,6 +3338,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "signal-hook-registry"
+version = "1.4.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c4db69cba1110affc0e9f7bcd48bbf87b3f4fc7c61fc9155afd4c469eb3d6c1b"
+dependencies = [
+ "errno",
+ "libc",
+]
+
+[[package]]
 name = "simd_cesu8"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3584,6 +3622,7 @@ dependencies = [
  "libc",
  "mio",
  "pin-project-lite",
+ "signal-hook-registry",
  "socket2",
  "tokio-macros",
  "windows-sys 0.61.2",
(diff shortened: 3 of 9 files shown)


tobert:
## Cross-family hard-look review — kaibo `deepseek` cast (holistic, worktree checkout)

**Verdict: MERGE.** The router fix is correct, the harness is hermetic and flake-free, all invariants hold, and the changelog placement is honest.

### 1. Router fix — CONFIRMED
The instance router carries exactly four mutations (`src/server/mod.rs:1083-1235`): build, remove disabled/unstaffable routes (`:1153-1162`), inject cast enums (`:1181-1184`), stamp the `alwaysLoad` pin (`:1195-1202`). All four were lost under the bare attribute; all four are restored by `router = self.tool_router` (`:3505`). No other instance-stamped mutation exists to worry about.

### 2. Severity — Fixed, not Security
The wire served the wrong router, but every internal guard read the correctly-gated instance fields: `batch_submit` keyless fails at key resolution (`src/config.rs:722-728`), `deliberate` at cast staffing (`mod.rs:2134`), `generate` at the image-slot/CAS gates (`:2923`/`:2913`). The real exposure was `--no-run-kaish` not blocking the shell — an operator capability boundary, while read-only/containment/no-external-commands all held. The structural invariants were never breached, so `Fixed` is the honest bucket, and the bullet's wording meets the style guide.

### 3. Harness — hermetic, negligible flake risk
`env_clear()` covers every env var the config loader reads, including provider keys; the temp `HOME` kills the key-file fallbacks; explicit `--root` keeps the launch cwd out of the allowed set. 60s timeouts are hang detectors, not performance assertions; per-test processes, no shared state. The `KEYLESS_TOOLS` equality assertion produces a readable diff on the next intentional tool addition.

### 4–5. Invariants + teeth
Blessed count unchanged at 5; dev-dep features stay out of the shipped binary under resolver v2 (`Cargo.toml:190-196`); stdio-only untouched (client transport is test-side). Reverting the attribute fails both gating tests; a future default-change or attribute removal is caught. **One gap:** nothing asserts `_meta["anthropic/alwaysLoad"]` on the wire — a regression that keeps routing but drops the meta injection would pass. Minor (UX optimization, not correctness) — worth one assertion.

### 6. Prose
The load-bearing comment and changelog bullet both meet the guide. **One devlog sentence is ambiguous enough that this review tripped on it** (`docs/devlog.md:48-49`: "a server started with `--no-run-kaish` dropped the tool from its list and then ran the script anyway" — under the bug, `tools/list` and `tools/call` read the SAME ungated router, so the tool was listed too; the sentence compresses the operator's expectation and the observed call behavior into one clause). Worth a clarity pass.

———
tokens · 4632581 in · 28826 out · 14680 reasoning · 4488064 cached

tobert:
Post-review commit `27bdcee`, from Amy's feedback (tests should always start from a blank env): the two older binary-spawning tests in `tests/gating.rs` now use the same `env_clear()` + temp `HOME`/XDG pattern the harness established, replacing an enumerated `env_remove` denylist that would rot as new variables ship. AGENTS.md now states the rule under **Working here**. Gating + mcp_stdio suites green, clippy clean.

PR #141, opened 2026-08-10
docs: release-prep groom — changelog ratchet, two tracker entries, explorer field notes

tobert:
Release-prep grooming, done while #140 was in review (kaijutsu lead greenlit no-merge release work; release-readiness is today's plate).

## What and why

- **CHANGELOG unreleased: 390 lines → 128.** The AGENTS.md ratchet says the section shrinks before any release retitle: one line per bullet, the change plus at most one clause of reason, reasoning in the PR bodies (which are the merge commits). The two `### Fixed` buckets merged into one. No entry was deleted — every change that had a bullet still has one.
- **Tracker: hallucinated-tool corrective re-prompt (P2).** From the 2026-08-09 field repro (`UnknownToolCall: 'run_sha'`, 2× identical, dossier phase killed pre-synth; matches the 2026-08-05 `run_grail` case). Written to be distinct from #135's shipped retry: that resends the same request on a provider parse failure; this request parsed and an identical resend re-hallucinates, so the fix shape is feeding the error back as a turn.
- **Tracker: kaibo-authored refusal for gated tools (P3).** From #140's report — a gated tool refuses with rmcp's generic `-32602 tool not found`, which reads the same as a typo. Error strings carry Full weight in the style guide, so the refusal should distinguish "disabled by flag" from "no cast can staff it" and say what to do instead.
- **docs/casts.md: "Choosing an explorer" field notes.** The operator-facing halves of the two field reports: flash-tier explorers can hallucinate tool names on long restricted-toolset prompts (workaround: pro-tier, or gemini-flash-lite), and an explorer should be matched to its provider's TPM headroom rather than its synth's family.

Docs only; no code or behavior change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,393 +15,132 @@ record. Each later release appends a new section at the top.
 
 ## [Unreleased]
 
+### Added
+
+- **`deliberate` keeps its dossier in the media CAS, and `dossier` reuses one** — hand
+  the digest to a second cast to reason over the same evidence for one synth's price.
+- **A refused dossier write never fails the deliberation** — you lose the record, not
+  the answer.
+- **Batch deliberations persist their handle**, labelled with the dossier's address, so
+  `job_list` leads back to the evidence after a restart.
+- **`kaibo deliberate`** — a batch cast prints the durable handle and exits; a direct
+  cast runs in the foreground and prints the answer
+  ([#82](https://github.com/tobert/kaibo/issues/82)).
+- **`kaibo batch cancel <handle>`** — the wording says "requested", because a provider
+  finishes what is already in flight.
+- **`kaibo generate`** renders into the artifact store; a deferred render polls in the
+  foreground rather than returning a handle nothing could collect.
+- **`kaibo cas read <digest>`** — metadata to stderr, content to stdout, binary as raw
+  bytes; `read` is the only verb, and the store keeps no index.
+- **kaibo warns when the media CAS sits on a filesystem that will not survive the
+  container** (overlayfs/tmpfs/ramfs, Linux) — it proceeds, and `[cas] backing` in
+  `kaibo://config` reports the finding.
+- **`read_cas` — read a stored artifact back by digest.** Metadata always leads; reads
+  are bounded, ranges always advance, and an image up to 5 MiB arrives whole and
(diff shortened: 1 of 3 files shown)


tobert:
Cross-family review (DeepSeek, two passes via kaibo itself) — **MERGE, no defects.**

**Pass 1 — `consult` (deepseek-v4-flash explorer / v4-pro synth) over the worktree**, tracker entries and casts.md verified against the code:
- Hallucinated-tool entry: accurate. `src/completion_retry.rs` resends the *identical* request on a provider parse failure (`completion(request.clone())`, pinned by `a_retry_sends_the_identical_request`), which is exactly the mechanism the new entry distinguishes itself from.
- Gated-tool refusal entry: accurate. `remove_route` physically drops the route, no `call_tool` override exists, and `tests/mcp_stdio.rs::a_gated_tool_is_neither_advertised_nor_callable` pins today's rmcp-worded `-32602`.
- casts.md field notes: vocabulary consistent with the doc's backend/role/cast/slot definitions; `explorer_backend`/`explorer_model` confirmed real on `ConsultArgs`/`ExploreArgs`/`DeliberateArgs`.

**Pass 2 — `oneshot` (deepseek-v4-pro) with both changelog versions attached**, compression fidelity: **FAITHFUL.** Every original entry retains a bullet; no compressed bullet asserts anything the original did not, and no load-bearing qualifier dropped (Linux scope on the CAS backing warn, `[artifacts] enabled` opt-in, and the deliberate `thinking_style` fold all checked specifically).

The first consult pass couldn't reach main's pre-groom changelog from the worktree sandbox (containment working as designed); the oneshot pass closed that gap with both files attached.

tobert:
Post-review commit from Amy's feedback: the casts.md section is rewritten clinically — no named incidents, tiers not models, leading with "confabulation risk is why we usually want smarter models." Her question (was the failing explorer running with thinking + high effort?) is answered in the tracker entry with code evidence: yes, by construction — the DeepSeek wire emits `thinking: enabled` + `reasoning_effort: "high"` regardless of model id (`shaping.rs:306`, `:457`), so the corrective-feedback fix is the right shape, not deeper thinking. The incident specifics stay in `docs/issues.md`, which is the evidence-first tracker, not a model-facing doc.

PR #144, opened 2026-08-10
chore(deps): rmcp 3.1.2, and correct what #142 says about the SDK

tobert:
Amy asked whether a newer rmcp answers the SEP-2549 gap #142 patched, rather than kaibo carrying its own fill. It does, partly — and the part it misses is why kaibo keeps its own answer.

## What rmcp 3.1.1 actually fixed

Upstream #1120, "emit cache hints from handler macros":

- The fill lives in **rmcp-macros only** (`tool_handler.rs:64-82`, `prompt_handler.rs:64-72`): `tools/list` and `prompts/list`, when the session negotiated ≥ `2026-07-28`.
- `list_resources`, `list_resource_templates`, and `read_resource` have **no fill at all** — the default `ServerHandler` bodies still return `::default()` with both fields `None` (`src/handler/server.rs:380-403`). kaibo hand-writes every one of these methods, so the macro would never reach them regardless.
- The macro answers **`cacheScope: public`** (`tests/test_handler_cache_hints.rs:79-89`). kaibo's advertised surface is resolved per connection from this user's env, config, and staffing, so `private` is the true answer; `public` would invite an intermediary to serve one user's tool list to another.

So `sep_2549_cache_fields` stays, and so does the hand-written `list_tools` that suppresses the macro's version.

## Why bump anyway

For the rest of the 3.1 line, not for the fix: recognizing the 2026 MCP methods (3.0.0), stricter protocol-metadata validation and `honor supported_protocol_versions when negotiating initialize` (3.1.0), plus auth/SSE fixes. That is conformance work aimed at exactly the kind of strict client that made kaibo's tools vanish, and beta → released is the right place to sit before a tag.

**Zero call-site changes.** The only new crates are build-time proc-macro deps (darling 0.24, syn 3.0.3, both pure Rust). `cargo tree -i` is empty for `aws-lc-rs`, `aws-lc-sys`, `openssl-sys`, and `mimalloc` (checked with `--all-features`), so the static-musl story is intact. 1073 tests pass, clippy clean on `--all-targets`.

## The tightening

#142's comments describe rmcp 3.0.0-beta.5 and are now false in two places:

- "its serve loop overrides any handler attempt to answer lower" — no longer true (3.1.0, upstream #1093). Refusing the version is available now, so the comment says why kaibo still fulfils it instead: a client asking for the newest version wants what that version's other features buy it, and the two fields cost nothing to answer truthfully.
- "Delete this when a bumped rmcp fills the fields itself" was the wrong deletion signal. A bumped rmcp fills two of the five results, with the wrong scope. The test now says so and asserts kaibo's values, so it fails if the SDK's ever start winning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@ record. Each later release appends a new section at the top.
 
 ### Changed
 
+- **kaibo speaks MCP through the released rmcp 3.1.2**, off the 3.0 beta it shipped on,
+  for that line's protocol conformance fixes.
 - **The built-in deepseek and anthropic synths pin `max_tokens = 32768`** — a measured
   consult showed reasoning consuming half the completion budget before the answer
   started.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,7 +194,7 @@ checksum = "d64a66d21a80182b35b1741997a6d2456911f54b7eb1918aa4e2382fc205268d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -228,7 +228,7 @@ checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -239,7 +239,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -298,7 +298,7 @@ dependencies = [
  "regex",
  "rustc-hash 1.1.0",
  "shlex 1.3.0",
- "syn",
+ "syn 2.0.117",
  "which",
 ]
 
@@ -378,7 +378,7 @@ checksum = "f65693059b6b9c588b9f62fed1cedbf0a8b805631457ea162d68f0de186f3de5"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -520,7 +520,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


tobert:

Holistic, no diff supplied; the cast read the branch tree itself. **Verdict: no defects.** 657k in / 9k out / 621k cached.

Four questions asked, answered with `file:line`:

**1. Every result path SEP-2549 requires the fields on.** All five list/read operations apply the helper — `list_tools` (`src/server/mod.rs:3557`), `list_resources` (`:3621`), `list_resource_templates` (`:3635`), `list_prompts` (`:3681`), `read_resource` (`:3649` + the patch at `:3668-3669`, since the inner `read_kaibo_resource_with_config` at `:4502` has no `RequestContext`). Correctly *not* applied: `get_prompt` (`:3690` — `prompts/get` is neither a list nor a read), `tools/call` results, `set_level` (`:3607`), and the `McpError` paths at `:4562`/`:4573`, which pass through the `.map()` at `:3665` untouched because an error is not a result. **No path needs the fields and lacks them.**

**2. The lexical version compare** (`:3533-3543`) is correct for every version the SDK can hand it — MCP versions are ISO dates, which sort lexically as they sort chronologically, and `is_some_and` returns `false` when no version was negotiated, so the no-version case lands on fields-absent.

**3. The three rewritten comment blocks** state nothing the code does not do. On the deletion signal specifically: the hand-written `list_tools` at `:3552` suppresses the macro, so the only road to the SDK's `public` is deleting kaibo's method — and `tests/mcp_stdio.rs:615-619` asserts `"private"`, so **the test does genuinely fail if the SDK's values ever start winning.**

**4. Drift risk on the `list_tools` override.** Gating drift is not possible: the hand-written method serves `self.tool_router` (`:739`) — the same instance `new_with_env` mutated for dropped routes (`:1153-1162`), injected cast enums (`:1181-1184`), and the `alwaysLoad` pin (`:1195-1202`) — and `#[tool_handler(router = self.tool_router)]` at `:3545` keeps `tools/call` on it too. Losing that annotation is caught by `a_gated_tool_is_neither_advertised_nor_callable` (`tests/mcp_stdio.rs:401`); the roster itself is pinned by `a_keyless_server_advertises_exactly_the_castless_surface` (`:296`) and `consult_carries_the_always_load_pin_and_no_other_tool_does` (`:434`).

**Non-blocking note worth keeping** (the one asymmetry it surfaced): `list_tools`'s struct literal names every field with no `..Default::default()`, so a new *required* field on `ListToolsResult` is a compile error — the strongest guard we have. The other three list methods use `..Default::default()` (`:3626`, `:3640`, `:3686`), where a new field silently takes its default. The wire test catches that only if the spec makes the field REQUIRED at the negotiated version. Left as-is here; the spread is what makes those methods short, and the next rmcp bump is the moment to re-read them.

PR #145, opened 2026-08-10
chore(kaish): wire up kaish 0.14 ahead of its tag

tobert:
**DRAFT — DO NOT MERGE until kaish publishes `v0.14.0`.** This branch builds against kaish `main` through a temporary `[patch.crates-io]` git pin, and kaibo ships from crates.io.

Amy: *"we could kick off a worktree and start on it using a git ref, and switch to the release before merging."* This deliberately reverses the 08-09 "no git pin to kaish main, no prep branch" call, with the reversal scoped to this branch and undone before merge.

## Swap checklist (before this leaves draft)

1. Delete the `[patch.crates-io]` block at the bottom of `Cargo.toml` (it carries its own instructions).
2. Set `kaish-kernel` to `"0.14.0"`. The dep line still says `0.13.0` on purpose: kaish's workspace version is `0.13.0` on `main`, so a patch source claiming `0.14.0` would not satisfy the requirement.
3. `cargo update -p kaish-kernel`, re-run the full suite and clippy.
4. Re-run Battery F against the released kernel and confirm the observed behavior still holds.
5. Cross-family review, then Amy's word.

## What the bump actually moved

**One call site, and it was a deletion.** kaibo carried `strip_write_side_paragraphs` to strip the shared kaish contract's "Overlay mode" paragraph — a virtual write layer taught to kaish embedders in general, dead weight in a kaibo preamble and a mixed signal one paragraph before "writes are refused". kaish-help #297 made that guidance opt-in (`Concept::Overlay`, reached only via `Selector::with_overlay`), so the paragraph stopped arriving. The filter's own guard assertion is what failed and told us — which is what it was written for.

**The property outlived the filter.** The checklist we were handed said to delete the filter *and* its test. Deleting both would leave nothing failing if overlay guidance ever came back on by default. `core_carries_no_write_side_teaching` keeps asserting the property; only the mechanism moved, from our filter to upstream's default plus kaibo never opting in. Sabotage-checked: composing `.with_overlay()` turns it red.

**Everything else was verified inert against the tree, not filed:** lowercase `JobStatus` wire spellings and `jobs --json`'s new `path` field reach no parse site (model-facing kaish never grows `jobs`, so there was never a parse site to rot); `kill_on_parent_death` defaulting true has no children to signal with `subprocess` off; `Kernel::shutdown` and `ExecResult.approval` are surfaces a read-only embedder never touches. The #301 comma sweep was a no-op — kaibo never told a model to quote a comma.

## The boundary change worth reviewing

kaish 0.14 mounts an approvals ledger at `/v/approvals` in **every** kernel, unconditionally. That is a model-readable path kaibo did not have before, and its read-only-ness comes from kaish's `vfs/approvalsfs.rs` chokepoint — **not** from any of kaibo's four levers, because `/v/*` is kernel space our backend never sees. New **Battery F** in `docs/sandbox-probes.md` keeps checking the part of the boundary we don't own.

Written from a live run, and the live run corrected the expectation we were given: **`/v/approvals/log` is not empty** with approvals off — it is an append-only observation ledger of every command the script runs. So the assertion is *scope*, not emptiness: every entry must come from this call's own script under one `kernel_id`. kaibo spawns a fresh kernel per call (`src/server/mod.rs:2540`, `src/consult/engine.rs:1903`), so the ledger cannot span calls, sessions, or projects. The kaish session confirmed the ring buffer (4096 retained, observation entries evict freely) and is correcting its own doc line.

## Unrelated fix found while preparing the probe pass

**E3 was stale in the direction that matters.** It claimed kaibo makes "exactly one `std::fs` mutation in production" and would fail "if a second blessed site is added" — written before the media CAS and `kaibo cas read`. The real guard pins **5 lines across 3 files**. A pass criterion that has drifted below the truth reads as an all-clear, so E3 now points at `BLESSED`/`EXPECTED_BLESSED_LINES` as the source. The same drift had reached `no_write_path.rs`'s own module doc ("two blessed exceptions", "currently 4").

## Gates

1073 tests pass, clippy clean on `--all-targets`, against kaish `main` at `809e39f`. Re-run required after the swap.


tobert:
## Swap-checklist addition: the comma rule that #146 just shipped

Amy, 2026-08-11: *"kaish reserves a bare comma, so `sed -n 120,400p`" is fixed in next release of kaish and we should update such references to the `,` and kaish.*

#146 (merged as `a3559c6`) landed the quoted-range idiom **and stated the rule to the model in prose** — "kaish reserves a bare comma; quote a comma-bearing argument" — because on kaish 0.13 the unquoted form is a parse error. Once this branch is on the fixed kernel that rule is **false**, and a false rule in a preamble is worse than a missing one: a model reads it every turn.

So this branch owns the reversal. Added to the checklist:

6. **Re-audit the comma prose and its tests against the fixed kernel.**
   - Confirm the fix is actually in the release being pinned (0.14.0 or later) — run both forms live, the way #146 did, rather than trusting the changelog.
   - Drop the *rule* sentence from the preambles wherever it explains why the range is quoted. The quoted form itself still works on every version, so the **example** can stay; it is the justification that goes stale.
   - #146 added two tests that scan **every** `sed -n ` kaibo writes across the preambles and assert the quoting. Those assert a property that is about to stop being required — decide per test whether it becomes a style pin (we still write quoted, deliberately) or is deleted. Do not just delete both; a scanner with nothing left to assert is the failure mode #145's own `core_carries_no_write_side_teaching` note already warns about.
   - Grep wider than the preambles: the kaish cheatsheet (`kaish_syntax.rs`), any `docs/*.md` that repeats the reservation, and the #146 PR body's claim.

Same discipline as the `strip_write_side_paragraphs` deletion in this branch: upstream removed the need for our workaround, so the workaround goes — but only after we decide what property, if any, outlived it.

PR #146, opened 2026-08-10
feat(prompts): teach the span-around-a-grep-hit and `file` reading idioms

tobert:
Two reading moves kaibo's models were never taught, added to the block every preamble embeds.

## Precise ranges, the follow-up to a grep hit

The explorer already knew `cat -n FILE | sed -n '1200,2400p'`, but only as *truncation recovery*: something you reach for after a whole read comes back as a head+tail sample (exit 3). It never learned the same command as the **proactive** move after a `grep -rn` hit lands in a large file. The addendum now says so, and the explorer's HOW TO READ block branches off its whole-file default instead of contradicting it.

**Whole-file reading stays the default, deliberately.** It is what produces the accurate `file:line` citations kaibo rewards, and the preamble still says "Read files WHOLE" three times before it offers a span. The new sentence is the branch, not a replacement, and the test asserts the order.

The span is written as `cat -n FILE | sed -n '120,400p'` rather than bare `sed -n '120,400p' FILE`, matching every other span idiom in the tree: the pipe through `cat -n` keeps the *real* line numbers in the output, so a citation off a span read is as exact as one off a whole read. Only the range is emitted either way.

## The range is written quoted, and that is load-bearing

kaish reserves a bare comma (brace expansion, lists). On **kaish 0.13** — the version `main` and the released binary run — `sed -n 120,400p` is a parse error:

```
1:29 [parse]: an unquoted comma splits this into separate words — kaish reserves `,`
(brace expansion, lists); quote a comma-bearing argument to keep it one word
```

Quoted works on 0.13 and on the unreleased 0.14 alike. Both forms were run live against both versions before any prose was written. Teaching the unquoted form would cost a turn for every model running today's kaibo, so:

- the rule is **stated to the model in prose**, not only demonstrated — a model composing its own range needs the rule, not just the example;
- two tests scan **every** `sed -n ` kaibo writes across the preambles and attachment directives, and fail unless a quote follows it. Structural on purpose: it survives someone changing the example line numbers, and it catches a future "simplification" that drops the quotes.

## `file FILE` before an unfamiliar file

So a model identifies a binary from its content instead of discovering it by dumping bytes into its own context. Verified live against the running server:

```
Cargo.toml: text (text/plain)
target/debug/kaibo: binary (application/x-executable)
```

## Placement

Both idioms went into `KAISH_SANDBOX_ADDENDUM`, which reaches the explorer preamble, the consult driver, and the internal `run_kaish` tool description through one string. Only the explorer's HOW TO READ block needed a second edit, because its "open that file whole instead of reading only the lines around the match" sentence would otherwise argue with the new guidance. The consult driver was left alone: the addendum sits above its own reading paragraph and covers it, and duplicating a sentence into every preamble is a cost charged per call.

`kaibo_sandbox_doc()` (the verbose `kaibo://kaish/sandbox` resource, a superset of the addendum) picked up a `file` bullet and the quoting rule on its span bullet, so the two stay in step.

## Checked, deliberately not changed

kaibo's model-facing text never describes conditionals as `[[ ]]` at all, so kaish 0.11's `test` builtin left nothing stale to fix. The `[[ ]]` examples that do appear in the composed contract are upstream `kaish-help`'s, which kaibo composes rather than authors — not ours to rewrite. Per the brief, no new content was invented to fill a gap that isn't hurting anyone.

## Gates

- `cargo test` — **1076 passed, 0 failed** (baseline on `main` is 1073; +3 new tests).
- `cargo clippy --all-targets` — clean.
- `rustfmt` on the two edited files only; the repo is not rustfmt-clean, so no repo-wide format. One incidental reflow of an untouched test was reverted to keep the diff on-topic.
- `built_in_preambles_are_written_without_em_dash_clause_chains` green — no em-dash clause chains, no length estimates or size targets, positive framing throughout.

Each new test was proven able to fail: unquoting the ranges in the source fails both structural tests, and deleting the new explorer sentence or the `file FILE` sentence fails its own assertion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,8 @@ record. Each later release appends a new section at the top.
   at flash-tier explorers inventing `run_*` tool names under a restricted toolset.
 - **kaibo's prompts are written in plain, literal English** — most of the models kaibo
   drives are not English-first, and figurative phrasing costs them attention.
+- **kaibo's models read a wide span around a grep hit in a large file, and run `file`
+  on an unfamiliar one** — whole-file reading stays the default everywhere else.
 - **`oneshot` and `deliberate`'s direct lane are one literal request** instead of a
   toolless pass through the managed loop — proven request-for-request identical.
 - **Batch treats `effort` as a floor, not an override** — a slot asking deeper than the
--- a/src/consult/prompts.rs
+++ b/src/consult/prompts.rs
@@ -212,8 +212,10 @@ pub fn report_preamble() -> String {
          their impls, the call sites, and exact line numbers together, and most source \
          files are small enough to read in a single command. Use `grep -rn PATTERN .` \
          to find WHICH files matter (`-B4 -A8` shows a preview around each match). \
-         Once grep names a file, open that file whole instead of reading only the \
-         lines around the match. A whole-file read of a very large file comes back \
+         Once grep names a file, open that file whole. When the file is large, read a \
+         wide span around each match instead, with `cat -n FILE | sed -n '120,400p'`. \
+         That returns the range and keeps the real line numbers, so your citation \
+         stays exact. A whole-file read of a very large file comes back \
          truncated (exit 3), and the truncated result still contains the start and the \
          end of the file. Read the rest in targeted spans: run `grep -n SYMBOL FILE` \
          to get the line numbers the question needs, then read a wide span around each \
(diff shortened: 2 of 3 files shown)


tobert:

Holistic, no diff supplied; the cast read the branch and judged the blocks whole. **Verdict: sound.** Two actionable findings, one fixed here and one deliberately left.

**Fixed (`3a1cd6f`): the addendum's structural scan had nothing asserting it scanned anything.** The loop over `sed -n ` matches asserts a quote follows each one, but a loop over zero matches passes green — so removing the last span example would silence the scan rather than fail it. The sibling scan in `prompts.rs:886` already asserted its count; this one now does too. Honest about its weight: sabotaged, the content assertion trips first, so this is belt-and-braces against an edit that removes both.

**Left alone, and flagged instead: the "A span of about 1,200 lines fits in a single read" sentence** (`src/consult/prompts.rs:222`). The reviewer named it as the one tension with the AGENTS.md rule against size cues, which is a fair reading of the rule as written. It is **pre-existing text on `main`**, introduced by `f89cf89` — the commit that rewrote these prompts into plain literal English — so it has already survived one deliberate pass over exactly this prose. Two reasons it is arguably fine: it is *tool-capacity* guidance rather than a target for the model's own output, and the measured failure behind the rule is a driver stopping its turn early, which a "you can get 1200 lines per read" fact does not encourage. Changing it is a judgment call about older text with its own argument, so it does not ride in on a PR about two new idioms. Amy's call.

**What the review confirmed rather than assumed**, each grounded in the tree:

- The whole-file default survives the additions. "Whole" is stated three times before the branch, the branch carries an explicit condition, and "instead" ties the span to the specific action it replaces. It flagged "read a wide span around **each match**" as the sentence a weaker model could over-apply, with a one-word strengthening ("too large") as the option — worth keeping in mind if telemetry ever shows narrow reading.
- Every new test fails on deletion of the text it guards, and the em-dash and quoting scans are mechanically enforced across all four preamble sources.
- Nothing in the blocks is stale or self-contradictory after the change, including the pre-existing text.
- The apparent redundancy (the span idiom stated in both the addendum and the explorer's own HOW TO READ) is the licensed kind — AGENTS.md's "verbose to install behavior" applies hardest to the explorer, which is the one reading code.

1076 tests, clippy clean.

PR #147, opened 2026-08-11
feat(consult): answer an unknown tool call instead of failing the phase

tobert:
A model that named a tool which was never advertised killed the whole phase. Observed twice identically on a flash-tier explorer (`run_sha`, and `run_grail` on an earlier build), each time throwing away every turn of reading already done — before the synth had spent anything.

## The reframe is the change

`docs/issues.md` tracked this as *"a bounded corrective re-prompt"*: catch the error, feed it back, ask again. Amy reframed it before any code was written:

> can we catch incorrect tool calls and return an informational error? sorta hooking method_missing

That is a different fix. A re-prompt is kaibo noticing a failure and spending a model call to work around it. `method_missing` is the call **succeeding**, with an answer that says "that's not a thing, here's what is". The model was going to take another turn regardless. It doesn't need to be asked again; it needs to be told something useful.

And rig had already built it: `rig-agent 0.41` ships `AgentHook::on_invalid_tool_call` returning an `InvalidToolCallAction`, and kaibo has implemented `AgentHook` since the `view_image` break work. So this is a second hook on a stack we already own, not new plumbing.

## Decisions

**`Skip { reason }`, not `Repair`.** rig will let a hook rename the call and dispatch it — `run_sha` → `run_kaish` would have worked on *both* observed failures. It is also a silent fallback: kaibo guessing intent and running a command nobody asked for, where a wrong guess doesn't announce itself. Naming the real tools is strictly more information and leaves the choice with the model. Amy's call.

**It declines under `ToolChoice::None`.** That is kaibo's forced write-up turn (`forced_finish_turn`) asking for prose, so a tool call there is a model ignoring the ask rather than fumbling a name, and the write-up path's own diagnostics are the better error. rig rejects a `Skip` under that tool choice anyway — declining says so on purpose instead of leaning on the rejection.

**The message says the turn ran nothing.** Reading rig's source changed this: when any call is skipped, *none* of the turn's tool calls execute; peers come back as `TOOL_NOT_EXECUTED_DUE_TO_INVALID_PEER`. A model told only that one name was wrong would assume its other calls landed. So the refusal's "what to do instead" asks for that work again — a second half we'd have missed by not reading the code we were calling.

**Installed in `run_phase_loop`**, the seam every tool-driven phase runs through, so the consult driver, each delegated `explore′` sweep, and the `deliberate` dossier sweep all inherit it at once. The toolless lanes (`oneshot`, `deliberate`'s direct synth) bypass the agent via `Arm::complete` and can raise no tool call at all.

## Tests

The decision is a pure function (`unknown_tool_feedback`) so it tests without standing up a rig run — three unit tests cover the message, the available-but-barred case, and the `ToolChoice::None` decline.

The e2e is the original repro: a scripted model reads a file, calls `run_sha`, then answers. **Deleting the `add_hook` line fails it with the production error verbatim** — `UnknownToolCall: model attempted to call unknown or disallowed tool run_sha`. The sabotage doesn't approximate the bug, it *is* the bug.

Full suite green (727 lib + every integration suite, 0 failed), clippy clean.

## What this does not do

It is recovery, not prevention — the rate is unchanged. The `warn` per occurrence (model id + advertised toolset) is the first telemetry this class has had. `docs/issues.md` keeps the open half: incidence unmeasured, prevention untouched, and a repeat offender still burns a turn per recovery with nothing counting them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,8 @@ record. Each later release appends a new section at the top.
 
 ### Fixed
 
+- **A model calling a tool that does not exist no longer throws away the investigation**
+  — it gets a tool result naming the real toolset, and corrects itself on the next turn.
 - **`--no-<tool>` gates work again over MCP.** Since the rmcp 3.0 upgrade a disabled or
   unstaffable tool was still listed and still callable; it is now neither.
 - **A new Claude Code sees kaibo's tools again.** The client negotiates protocol
--- a/docs/devlog.md
+++ b/docs/devlog.md
@@ -14,6 +14,54 @@ per ship date; multiple ships on a date get sub-bullets.
 
 ---
 
+## 2026-08-11 — method_missing, for models
+
+We had this tracked as "a bounded corrective re-prompt": catch `UnknownToolCall`, feed the
+error back, ask again. That's how it reads in `issues.md`, written the day the second repro
+landed. Amy reframed it in one line before I'd written any code — *"can we catch incorrect
+tool calls and return an informational error? sorta hooking method_missing"* — and the
+reframe is the whole entry, because it changes what the fix *is*. A re-prompt is kaibo
+noticing a failure and spending a model call to work around it. `method_missing` is the
+call **succeeding**, with an answer that happens to say "that's not a thing, here's what
+is". The model was already going to take another turn. It doesn't need to be asked again;
+it needs to be told something useful.
+
+Then rig had already built it. `rig-agent 0.41` ships `AgentHook::on_invalid_tool_call`
+returning an `InvalidToolCallAction`, and kaibo has implemented `AgentHook` since the
+`view_image` break work — so the fix is a second hook on a stack we already own, not new
+plumbing. This keeps happening and it keeps being worth checking first: the same lesson as
+the retry policy we decided to upstream rather than hand-roll.
+
+Four resolutions were on offer, and the interesting one to reject was `Repair` — hand rig a
+replacement name and let the call through. It is *right there*, `run_sha` → `run_kaish`
+would have worked on both observed failures, and it is a silent fallback: kaibo guessing
+what the model meant and running a command nobody asked for. A wrong guess doesn't announce
+itself. `Skip` with a reason string is strictly more information and leaves the choice where
+it belongs. Amy picked it.
+
+Two things in rig's source changed the design. Under `ToolChoice::None` rig *rejects* a
+`Skip` outright — and that is exactly the tool choice kaibo's forced write-up turn uses. So
+the hook declines there, and it declines on purpose rather than leaning on the rejection: a
+tool call during a turn that asked for prose is not a fumbled name, it's a model ignoring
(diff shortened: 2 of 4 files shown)


tobert:

Dogfooded through kaibo itself, holistic and no diff supplied — the cast read the branch and judged the code whole. ~1.0M tokens in / 15.6k out (969k cached).

**Verdict: sound.** The `Skip`-over-`Repair` choice, the hook's statelessness across the view_image-break resume, the interaction with the empty-answer guard (a synthetic tool result satisfies `transcript_has_tool_results`, so recovery routes to the forced write-up rather than erroring), and the message's three-part refusal shape all came back clean. Two findings were real, and both are fixed in `1b6e51d`.

### 1. The `ToolChoice::None` branch is not the live guard I said it was

> This check is defense-in-depth: `forced_finish_turn` at line 1470 does **not** install this hook at all, so the check is unreachable in the current tree.

Verified independently before acting — `forced_finish_turn` (`engine.rs:1456-1471`) builds its own `AgentBuilder`, sets `.tool_choice(ToolChoice::None)`, and calls `.runner(prompt)` with no `add_hook`. So the hook never sees that tool choice.

The branch stays: it is the answer we'd want if the write-up path ever installed hooks, rig rejects a `Skip` there regardless, and it costs two lines. What changed is the prose. The doc comment now says plainly that nothing in this tree reaches it, and the test is renamed `a_prose_only_tool_choice_is_declined_rather_than_answered` — named for what it asserts rather than for a caller that doesn't exist.

This is the finding I'm most glad it caught. A comment that oversells its own reach is believed later, by someone deciding whether a path is covered.

### 2. Nothing tested the repeat offender

> The most impactful missing test is the repeated-invalid-calls → `MaxTurnsError` path.

Correct. Each recovery costs a turn and nothing counts recoveries, so the only thing between a pathological model and an unbounded loop is that a skipped turn is still a turn. The review reasoned that through (the `count_model_turns` history check as a second lock behind rig's own counter) and concluded it was structurally safe.

It was right — and `a_model_that_only_ever_names_phantom_tools_still_terminates` now proves it rather than arguing it. Budget of two, a model that names `run_sha` on every non-finalize turn, and it lands in the turn-cap finalize with an answer. Written under a `tokio::time::timeout` so that if the accounting ever changes it fails instead of hanging the suite forever.

### Findings deliberately not acted on

- **Co-tool-call case untested** (a valid `run_kaish` beside the phantom call). Real gap, but the scripted client can only observe what the *model* sees next, not which tools executed — so the test would assert the message rather than the behavior, which the unit test already does. The behavior itself is rig's, pinned by rig's own tests.
- **`ToolImageBreakHook` interaction untested.** The review worked out why it cannot fire (no tool call from an invalid turn executes, so no image ever reaches a tool result) and I agree. Not worth a test that would pass for a reason the code doesn't control.
- **"Slightly wordier than `SaveError`."** Accepted as-is — the extra clause is the whole-turn consequence, which is load-bearing information rather than decoration.

Gates after the fixes: 728 lib tests + every integration suite green, clippy `--all-targets` clean.

PR #148, opened 2026-08-12
feat(telemetry): export the logs signal, so kaibo's own events can be counted

tobert:
`[telemetry]` exported traces only. The span tree is rig's — prompts, completions, token usage — and it does not carry kaibo's own `tracing` events. Several things worth counting are **only ever events**: the `warn` when a model calls a tool that does not exist (`consult/engine.rs`), and the `warn` when a phase returns an empty answer and is forced into a write-up turn. Both were tracked as "incidence unmeasured", and the honest word was *unmeasurable* — the instrument existed and had no sink.

## Why now

Two parties wanted the same number, which is what moved this off P4. Genba (Amy's work fleet) asked for the unknown-tool recovery rate by name, to cite when they teach cross-model review; our own empty-answer entry wants the same sink.

The other road is closed. A provocation pass on 2026-08-11 found that a capable model asked to call a nonexistent tool simply declines — the local Gemma treated the name as a shell command, got kaish's `exit: 127`, found `checksum` in `help builtins`, and finished the task correctly; DeepSeek declined outright. **There is no benchmark for this class, only passive collection**, and passive collection needed a sink.

## Decisions

- **One opt-in covers both signals.** `logs` defaults on under `enabled`. The old deferral asked whether a logs signal was worth its content cost given traces already carry prompts — the answer is that kaibo's own diagnostics are strictly *less* sensitive than what an operator already accepted when they turned traces on. `logs = false` exists for a collector that takes spans but not records.
- **The logs filter is `kaibo=info`, narrower than the traces layer's `info`.** That asymmetry is the design, not an oversight: rig emits event-level chatter repeating prompt and completion text, and exporting it again as loose log lines would pay twice for the most sensitive bytes kaibo handles.
- **The logs endpoint is derived only from the standard shape, never guessed.** A collector at `/otlp/ingest` would otherwise get kaibo's diagnostics posted to a URL the operator never chose — discovered only by the records never arriving, which is exactly the silent failure this house refuses. Any other shape is a startup error naming `logs_endpoint`. Per Amy's ruling the same morning on `attach.rs`, the refusal states the action rather than pointing at a tracker.

## Proof, not assumption

- Broke the filter to `info` → the filter test fails with rig's chatter in the diff.
- Made the derivation fall back silently → the refusal test fails.
- A local HTTP sink caught the **real** exports: two POSTs, one to `/v1/traces` and one to the **derived** `/v1/logs`, both carrying the probe marker. The offline test proves the filter; only the wire proves the record leaves the process.
- Full suite green (733 lib + all integration), `cargo clippy --all-targets` clean.
- `cargo tree -i aws-lc-rs` and `-i mimalloc` both empty — the TLS invariant holds across the new deps (`opentelemetry-otlp`'s `logs` feature, `opentelemetry-appender-tracing`, and a dev-only `opentelemetry_sdk/testing` for the in-memory exporter).

## Tracker

`docs/issues.md` keeps the metrics signal and gains a new entry — *now go collect the two rates*. The plumbing is no longer the blocker, so the next honest status for those entries is a number or nothing.

## Review

Cross-family review by DeepSeek via kaibo's own `consult_submit`, posted as a PR comment. No diff supplied — it read the shipped source holistically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **`[telemetry]` exports the logs signal too** — kaibo's own `tracing` events, which the
+  span tree never carried, so a warning like a model calling a tool that does not exist
+  becomes countable.
+- **`[telemetry] logs_endpoint`** — omit it and kaibo derives the sibling of `endpoint`;
+  a non-standard endpoint is a startup error naming the key, never a guessed destination.
 - **`deliberate` keeps its dossier in the media CAS, and `dossier` reuses one** — hand
   the digest to a second cast to reason over the same evidence for one synth's price.
 - **A refused dossier write never fails the deliberation** — you lose the record, not
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1773,6 +1773,7 @@ dependencies = [
  "libc",
  "lru",
  "opentelemetry 0.32.0",
+ "opentelemetry-appender-tracing",
  "opentelemetry-otlp",
  "opentelemetry_sdk 0.32.1",
  "regex",
@@ -2271,6 +2272,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "opentelemetry-appender-tracing"
+version = "0.32.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c0080f0dc1d7c786f467cd85a4e395fcab11ee852004f39a29a18ab7c25d837"
+dependencies = [
+ "opentelemetry 0.32.0",
+ "tracing",
+ "tracing-core",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "opentelemetry-http"
 version = "0.32.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2341,6 +2354,7 @@ dependencies = [
  "portable-atomic",
  "rand 0.9.4",
  "thiserror",
+ "tokio",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,8 +153,15 @@ opentelemetry_sdk = "0.32"
 # builds a plain reqwest client and uses whatever TLS the (feature-unified) reqwest
 # was compiled with — i.e. our rustls-no-provider + ring. So we keep only the
 # transport features and let TLS come from the shared reqwest above.
-opentelemetry-otlp = { version = "0.32", default-features = false, features = ["trace", "http-proto", "reqwest-blocking-client"] }
+opentelemetry-otlp = { version = "0.32", default-features = false, features = ["trace", "logs", "http-proto", "reqwest-blocking-client"] }
 tracing-opentelemetry = "0.33"
+# The LOGS signal's half of the bridge: `tracing` events → OTel log records. Traces
+# carry rig's spans (prompts, completions, token counts); this carries kaibo's own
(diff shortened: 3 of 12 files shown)


tobert:
## Cross-family review — DeepSeek (`deepseek-v4-flash` explorer / `deepseek-v4-pro` synth), via kaibo's own `consult_submit`

No diff supplied; it read the shipped source holistically. **Verdict: no bugs found.** One suggestion applied (8513c42).

Per the house rule that kaibo claims get verified before they're accepted, I checked every checkable claim at source. Reporting the failures alongside the findings, because the pattern is more useful than the verdict.

### Verified true

- **Q3, the load-bearing one.** All three `tracing::warn!` sites in `src/consult/engine.rs` are at lines 776, 1359, 2345 — exact — and none sets an explicit `target:`, so each resolves to `kaibo::consult::engine` via `module_path!()`. `EnvFilter`'s `kaibo=info` prefix-matches on `::` boundaries, so the filter does capture what the feature claims to capture. This was the question that could have invalidated the whole feature; it holds.
- **Q1/Q2 lifecycle and feedback loop.** Both batch processors run on std threads with no tokio reactor, so the blocking reqwest client is right for both exporters; the `opentelemetry=off` directive sits on exactly the two layers that feed exporters, and the SDK's own failure path emits on that excluded target, so the loop does not close.
- **Q4 config plumbing.** The destructure guard in `config_resource.rs` is the forcing function for the render path; `apply_raw_env` has no equivalent, which is pre-existing and symmetric across every stanza rather than something this branch introduced.
- **The suggestion worth taking:** `opentelemetry_sdk` declared no features while the logs signal imports `SdkLoggerProvider`. `logs` is a default today, so a minor bump re-sorting the defaults could drop a shipped signal. Pinned in 8513c42.

### Claims that did not survive verification

All three are in the section the model volunteered *beyond* the four questions asked — which is itself the finding.

1. **`InvalidToolCallHook`** — the type is `UnknownToolFeedbackHook` (`src/consult/engine.rs:749`). Confabulated name.
2. **"the `#[instrument]` at :1140 creates a span with `gen_ai.operation.name = "invoke_agent"`"** — the span is `name = "run_phase"`, and its fields are `model`, `max_turns`, `gen_ai.request.thinking`, `gen_ai.response.finish_reason`. There is no `gen_ai.operation.name` field. `invoke_agent` appears in that file only inside a *comment*, describing rig's separate span that nests **under** `run_phase` — so the claim inverts the relationship.
3. **"the :2345 warn is in `record_qa_turn`, called from `run_phase_logged`"** — it is in `consult_session_turn` (`:2307`), which is a session-level entry point that calls the phase rather than being called by it.

### The interesting part: right conclusion, wrong evidence

That third section concluded log records correlate to their span via `trace_id`, which is also what my code comment asserts. Its supporting reasoning was wrong on two of three specifics — so I settled it on the wire instead of on either of our say-so.

I captured the real OTLP payloads and looked for a 16-byte id present in **both** the traces POST and the log record:

```
16-byte ids in traces: 5
ids ALSO present in the log record: 2
  53d7c9b49b3d807f1fd276ca9a98b734
  6b6169626f3a3a74656c656d65747279   <- ASCII "kaibo::telemetry", a false positive
```

One genuine shared `trace_id`. **Correlation works** — the comment stands, now on evidence rather than on a review that reached the right answer through wrong steps.

The shape to carry forward: the claims that failed were the unrequested ones, and they *had* line citations — they were specific and wrong, not vague. So "cites a `file:line`" is necessary and not sufficient; the check is whether the cited line says what the claim says.

PR #150, opened 2026-08-12
feat(orientation): publish file sizes in the repo map, and measure the prompt half honestly

tobert:
Started from a measurement, and ends with one null result and one change worth keeping.

## The study

16,105 real `run_kaish` reads over a month of traces, 18 models, attributed to the phase's model via the `run_phase` parent span:

- **median delivered read: 1,837 B** (~45 lines), against a preamble that already said "read files WHOLE"
- **whole-file reads truncated only 1.8% of the time** (286 / 16,105)
- **5× spread across models under identical prompt text** — `gpt-5.6-terra` 15% small reads, `deepseek-v4-flash` 54%, `z-ai/glm-4.7-flash` 85%
- the consult **driver** reads smaller than the explorer (59% vs 54% under 2 KiB)

Models were hedging against a risk that almost never fires, because from inside the sandbox a file's size is unknowable until it has been read.

## What's in this PR

**1. The map publishes sizes and the read-whole ceiling.** Each file lists its size; the block states the size at which one `cat -n` still returns a whole file, derived from the live `[sandbox] output_limit_bytes` rather than hardcoded (64 KiB cap → a 51 KB ceiling, since `cat -n`'s line numbering adds roughly an eighth on real source). Files above it are marked `read in spans` on their own line, so the exception rides the line it applies to. Sizes come from `stat --json` through the *same* kernel that enumerated the files, keeping orientation's one-source-of-truth property; a stat failure costs the sizes and never the map.

On this repo that renders 19 marked files — `src/server/mod.rs 402 KB`, `src/consult/engine.rs 316 KB`, `src/config.rs 246 KB`.

**2. The whole-file rule is stated in four registers**, with a test pinning each: imperative, data-anchored (the sizes), cost, and a re-anchor placed *after* the exceptions — the block previously ended on narrowing.

## The prompt half did not work, and that is the finding

Three questions per arm, top-level `explore` on deepseek, read sizes decoded off the OTLP wire:

| | control | treatment |
|---|---:|---:|
| median | 3,522 B | 3,385 B |
| mean | 5,553 B | 5,193 B |
| p90 | 11,889 B | 11,904 B |
| under 2 KiB | 26% | 34% |

Per-question medians swing 2,007–6,156 B *within* an arm, so the noise is an order of magnitude larger than the arm difference. **No detectable effect** — the second independent null.

The test was fair, not confounded: `server/mod.rs:2030` deliberately withholds `attach` from the top-level `explore` tool, so this measured the explorer where reading is its *only* evidence channel and a bigger read is worth the most. Where a sweep does have `attach` (the nested `explore′`, `deliberate`'s dossier), a small read followed by an attach is correct behavior and stride is the wrong metric.

Recorded so a third round of prompt tuning is not spent here. The next lever acts on the loop rather than the prose — information in the tool result itself.

## Gates

Full suite green (1089), `cargo clippy --all-targets` clean, `cargo tree -i aws-lc-rs` / `-i mimalloc` empty. The map's rendering, the ceiling tracking `output_limit_bytes`, the marked/unmarked split, and the sizes-absent degradation all have tests; each was proven failable by breaking the implementation first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **The repo map gives each file's size and the size that still reads whole**, so a model
+  compares two numbers instead of guessing how much of a file one read returns.
 - **`[telemetry]` exports the logs signal too** — kaibo's own `tracing` events, which the
   span tree never carried, so a warning like a model calling a tool that does not exist
   becomes countable.
--- a/docs/devlog.md
+++ b/docs/devlog.md
@@ -14,6 +14,69 @@ per ship date; multiple ships on a date get sub-bullets.
 
 ---
 
+## 2026-08-12 — three nudges, three nulls, and the lever that actually moves
+
+Amy noticed something we had all half-noticed: kaibo's agents read code in small,
+conservative slices. *"We've done rounds of tuning them to do bigger strides, but I
+wonder if we should do more thinking about how we prompt them."* The thinking turned
+into a measurement, and the measurement turned into a result none of us expected to
+like.
+
+The logs signal had just shipped, so for the first time there was a month of traces to
+ask. 16,105 real `run_kaish` reads, 18 models, each attributed to its phase's model
+through the `run_phase` parent span. The median delivered **1,837 bytes** — about 45
+lines — against a preamble whose first sentence already said "Read files WHOLE."
+Whole-file reads truncated **1.8%** of the time. So models were hedging against a risk
+that essentially never fired, and the obvious reason was that a file's size is
+unknowable from inside the sandbox until the file has been read.
+
+That gave a clean hypothesis and three ways at it. Publish the sizes in the repo map so
+the guess becomes a comparison. Say the rule several ways instead of once, because Amy's
+instinct was that repetition might carry where one clear sentence had not. Then put the
+budget in the tool result itself, so the number arrives at the moment the next read is
+chosen rather than in a preamble read once at the start.
+
+**All three moved nothing.** Not "a little" — nothing coherent. The third is the one
+worth recording, because it failed in an instructive shape: median up 17%, mean down,
+p90 down, small-read share up. A real effect moves those together. That is what noise
+looks like when you stare at it hopefully. Across every run, control and treatment
+medians sat in a ±10% band around 3,400 bytes.
+
+The honest conclusion was Amy's opening position, now with numbers behind it: *"the
+weights are the weights, we can only nudge them."*
+
+What the data does say, loudly, is in the column nobody was looking at. Under the
+**same** preamble and the same tools, `gpt-5.6-terra` reads at a median of 14,939 bytes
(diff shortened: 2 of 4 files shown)


tobert:
## Cross-family review — DeepSeek (`deepseek-v4-flash` explorer / `deepseek-v4-pro` synth), via kaibo's own `consult_submit`

**Verdict: blocked on the ceiling arithmetic.** It was right, and the fix is 4aa15cf.

Per the house rule that kaibo claims get verified before they're accepted, I checked its empirical claims at source before acting.

### The blocking finding — verified, and a real false promise

The review claimed the output cap applies to `cat -n`'s *numbered* output rather than the file's own bytes, that numbering costs exactly 7 bytes per line, and therefore that four fifths of the cap fails precisely for short-line files.

I measured it rather than take it:

```
src/tls.rs       bytes=3213 lines=62  numbered=3647  overhead=434  per_line=7
src/explorer.rs  bytes=4752 lines=133 numbered=5683  overhead=931  per_line=7
src/lib.rs       bytes=2067 lines=55  numbered=2452  overhead=385  per_line=7
```

Exactly 7, as claimed. And the falsifier it supplied holds: a 20 KB file of two-byte lines is well under the 51 KB stated ceiling, but numbering takes it to 92 KB against a 64 KiB cap — it truncates, while the map showed it **unmarked and promised whole**. My doc comment also had the reasoning backwards: it asserted four fifths covered short lines, which is exactly where 7 bytes/line dominates.

I scanned this repo: **no file trips it**, so the bug was latent rather than live. Short-line files (data, generated lists, CSV) are where it would bite.

### The fix: stop estimating

Rather than soften the prose to "typical source" — the review's first option — the mark is now **exact**. `wc -l --json` rides the same chunked call `stat` already made, so each file's real numbered size is computed and marked iff it exceeds the cap. An unmarked file is one that *will* come back whole. The review's other option, quoting a guaranteed `output_limit / 8`, is true but useless at ~8 KB.

The falsifier is pinned as `short_lines_are_counted_not_assumed`, and it fails on the old arithmetic — verified by regressing `numbered_output_bytes` to `bytes * 5 / 4` and watching it go red.

### Its other findings

- **Quoting is safe, and not a wrong-size bug** — confirmed. A path containing `'` is statted under its stripped name, so it misses on lookup and prints without numbers. A miss, never a wrong number on the right file.
- **Blast radius is wider than my comment said** — correct. `stat`/`wc` exit non-zero if *any* argument is missing, so one unreadable path costs its whole 64-file chunk. The comment now names the chunk.
- **Prose is reinforcement, not padding** — it judged all four registers as carrying distinct content and recommended no cut. I took that as-is.
- **The "About 1,200 lines" drift** it flagged in the preamble I left alone: that line sizes a `sed -n` *span*, not the whole-file rule, so it isn't a second answer to the same question.

Gates after the fix: 1090 tests green, `cargo clippy --all-targets` clean, `cargo tree -i aws-lc-rs` / `-i mimalloc` empty.

This is the review earning its keep — it found the one claim in the change that was false, in the direction that would have misled a model, and handed me the counterexample to pin.

PR #154, opened 2026-08-13
fix(wire): repair the vLLM response rig-core cannot decode

tobert:
A call to a vLLM-served model returned a complete answer — `finish_reason: "stop"`, the text right there in the body — and kaibo handed the caller a hard provider error. Reported by the **lfm2d lane** against Crusoe; reproduced here on `crusoe-thrift` (`Qwen/Qwen3-235B-A22B-Instruct-2507`).

Amy asked to **try 0.3.0 before tagging it**. This is what that found, and it is why the tag waited: gates were green, probes were stamped, reviews were applied, and one live `oneshot` found what none of that could.

## The bug

vLLM emits **both** spellings of the reasoning field on one assistant message:

```json
{"role":"assistant","content":"391","reasoning":null,"reasoning_content":null}
```

rig-core 0.41 declares one field for the two (`rename = "reasoning_content"`, `alias = "reasoning"`), so serde raises `duplicate field`. The failure does not *look* like a decode failure on the way out: rig's `ApiResponse` is untagged, so a body that misses the success shape falls through to the error arm, and `from_http_response` preserves a 2xx as `CompletionError::ProviderResponse`.

Proven rather than argued — the captured payload was deserialized against **rig's own type**: both spellings present = `duplicate field`, either one alone = OK.

**Not a regression.** The installed 0.2.0 binary fails identically on the same call. Latent, not introduced by reasoning-on-by-default — the failing model is non-reasoning and both fields are `null`.

**Not a Crusoe quirk.** The emitter is vLLM, so every vLLM-backed OpenAI-compatible endpoint carries it: Together, Fireworks, self-hosted.

## Decisions

**Repair the response, not the request.** Nothing kaibo can send stops a server from emitting both keys — the shape is the server's, not an answer to an option we set. The decode happens inside rig, so the last bytes kaibo owns are in the HTTP client it already injects via `.http_client(..)`. `Repaired` is that seam: the transport-side sibling of `Watched` and `Retried`.

**Narrow, and silent only where silence is correct.** The repair fires on the one shape that fails and returns the original bytes for everything else, so an unfamiliar body reaches rig byte-for-byte. Reasoning text carried only on the alias is **promoted, not dropped**. The one case that must discard text — two different non-null values — logs a warning rather than choosing quietly. Only the chat-completions client is wrapped; the Responses builder keeps the bare client, since its payload has no `choices` for the repair to recognize.

**No upstream report yet.** Amy: *"we will work around and revisit next time we bump rig."* The module doc carries that condition, and `repairs_nothing_when_rig_can_already_decode` is the trigger — it decodes the captured payload with rig's own type and **fails the day rig no longer needs help**, which is the day this module gets deleted.

## Verification

| gate | result |
|---|---|
| `cargo test` | **1106 passed, 0 failed** |
| `cargo clippy --all-targets` | clean |
| sabotage check | stubbing the repair to `None` turns **4 of the 7** new tests red, and leaves the deletion guard correctly green |

Live, against the call that actually failed:

- `oneshot --cast crusoe-thrift` — **was** `ProviderResponseError: status 200 OK`, **now** answers
- full multi-turn `consult --cast crusoe-thrift` — answers, having read the new module through the tool loop
- `crusoe`, `crusoe-ds4`, `crusoe-nemotron` — still answer as before

## Changelog

Lands in the `[0.3.0]` section rather than Unreleased, since that section is written but not yet tagged — same as #153.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,9 @@ record. Each later release appends a new section at the top.
 
 ### Fixed
 
+- **A complete answer from a vLLM-served model no longer arrives as a provider error** —
+  vLLM sends both `reasoning` and `reasoning_content`, which rig-core cannot decode, so
+  kaibo repairs the body before it is parsed.
 - **kaibo told models the wrong exit code for a refused write.** Six descriptions said
   `126` meant blocked; a refused write exits `1` saying `permission denied: filesystem is
   read-only`, and an external command exits `127`.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,10 @@ serde_json = "1"
 # base64 string (the only channel that lands image parts in model context). Already
 # in the tree transitively via rig-core; named here so the dep is ours, not implicit.
 base64 = "0.22"
+# rig's `HttpClientExt` is typed in `bytes::Bytes`, so `wire_repair`'s transport
+# wrapper needs the type by name. Already in the tree via reqwest/rig-core — zero new
+# crates; named here so the dep is ours, not implicit.
+bytes = "1"
 # XDG config.toml: named provider profiles + tunables. `deny_unknown_fields` on the
 # raw structs turns a typo'd key into a startup error rather than a silent no-op.
 toml = "0.8"
@@ -190,11 +194,6 @@ opentelemetry_sdk = { version = "0.32", features = ["testing"] }
 # `start_paused` tests (the deferred-generate poll loop) auto-advance tokio's clock
 # instead of really sleeping the poll interval.
 tokio = { version = "1", features = ["test-util"] }
-# Request-body capture in `tests/effort_wire.rs` and `test_support::CaptureHttp`:
-# rig's `HttpClientExt` trait is typed in `bytes::Bytes`, so implementing a fake
-# transport needs the type by name. Already in the tree via reqwest/rig-core — zero
-# new crates.
-bytes = "1"
 # rmcp's CLIENT half, for `tests/mcp_stdio.rs`: it spawns the real kaibo binary and
 # drives it over stdio the way a client does, so the MCP surface (handshake,
 # tools/list, tools/call, resources/read) is covered end to end in-tree. Same version
--- a/src/consult/engine.rs
+++ b/src/consult/engine.rs
@@ -471,10 +471,15 @@ impl Arm {
                         caps,
                     ))
                 } else {
+                    // The chat-completions wire is where vLLM sends both reasoning
+                    // spellings on one message, which rig-core 0.41 cannot decode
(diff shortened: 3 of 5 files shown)


tobert:
## Cross-family review — kaibo `consult`, deepseek cast

Dogfooded per house practice: the patch was written by Claude, so the review is a different lineage. `deepseek-v4-pro` synth / `deepseek-v4-flash` explorer, 497k in / 29k out.

**Verdict: repair logic and placement correct; the wrapper was the untested part.** Findings, and what happened to each:

| finding | action |
|---|---|
| **No test drives `Repaired::send` end-to-end** — every test called the function directly, so the module's one piece of novel code shipped with its property inferred, not asserted | **Fixed** in 84d7ec1 — two tests on a canned transport, sabotage-checked |
| Deletion canary over-couples to rig's error string | **Fixed** — comment now states the coupling is deliberate, so a *different* decode failure can't satisfy the trigger quietly |
| `memchr_contains` is a misnomer (never used the `memchr` crate) | **Fixed** — renamed `contains_bytes` |
| "Every response pays this scan" overstates cost — only OpenAI chat responses reach it | **Fixed** |
| Two theoretical false-negatives: a `\u`-escaped key the literal scan can't see, and a number `serde_json` can't represent without `arbitrary_precision` | **Documented** as known corners; each leaves the caller exactly where it was before this module existed |

**Confirmed sound, with reasoning worth keeping:**

- **The `Bytes`→`U::from` rebuild is sound at the type level, not just by inspection.** The trait bound is `U: From<Bytes>` with *no* `Deserialize` bound, so rig's own generic `send` can only ever construct `U` that same way — asking for `Bytes` and converting later is the identical conversion, one step down. Status and headers survive via `Response::from_parts`, and the body stays lazy, so error timing is unchanged.
- **No other kaibo path can hit this bug.** Batch's OpenAI lane is gated hosted-Platform-only and never decodes through rig's `CompletionResponse` at all (it parses its own JSONL); `discover` hits `GET /models`; media is an image wire. Leaving the Responses client bare is right — its payload has no `choices`.
- **Streaming can't reach the repair**, and if a raw SSE body ever did, `from_slice` fails and the body passes through un-mangled rather than corrupted.
- The both-non-null-different case is lossy **by necessity** — rig's field is singular, so a caller could never have consumed two values through it.

Per [the house rule on kaibo reviews](https://github.com/tobert/kaibo), I verified rather than accepted: spot-checked the batch gating citations (`batch.rs` `WireKind::Openai => backend.is_hosted_openai()`, `config.rs::is_hosted_openai`) against the real lines. They support the claim.

Gates after the follow-up commit: **1108 tests pass, clippy `--all-targets` clean.**

PR #156, opened 2026-08-14
feat(config): declared-only key sources + api_key_cmd

tobert:
## Summary

- **`api_key_cmd`** — a backend's key can come from a command whose trimmed stdout is the key (`["op", "read", "op://Vault/Item/Field"]`): raw argv, no shell, stdin closed (a child must never read the MCP stdio stream), 64 KiB refuse-not-trim, strict UTF-8, stderr shown only on failure, 30s timeout-kill, output never logged.
- **Key sources are declared, never seeded** (Amy's ruling this session) — the built-ins no longer read `ANTHROPIC_API_KEY` / `~/.anthropic-key.txt` out of the box; an operator names `api_key_env` / `api_key_file` / `api_key_cmd` once in config.toml, and file + cmd together is a load error naming both.
- **Invariant carve-out** — AGENTS.md's "cannot run external commands" sentence now names the one config-declared operator-side key fetch; the model-facing sandbox levers are untouched and the exec path is unreachable from it.
- Setup guidance, `config.example.toml`, `config.md`, README (incl. the Docker stanza), and CHANGELOG rewritten for declared-only; the resident instructions stay under the 2048-char budget.

## Test plan

- ~22 new tests, sabotage-checked: exec contract (trim, empty/non-UTF-8/64KiB/nonzero/missing/timeout-kill), stdin-null, sentinel oracle (env wins without running the command), post-merge XOR + empty-argv load errors, `$VAR` non-expansion, `key_status` Present-without-run, Arm-path, binary-level stream isolation (a marker key never reaches kaibo's own stdout), and a codified concurrency test for sync resolution on the multi-thread runtime.
- Full `cargo test` green (an earlier draft of this body excused two `tests/attach.rs` failures; on the rebased tree all 10 attach tests pass, so the excuse is retired). Clippy `--all-targets` clean.

## Reviews

- **GLM 5.2 (zai/GLM-5.2 via Crusoe)** did both cross-family reviews: the rev-1 design review (its findings shaped the plan — seed-collision root fix, timeout test seam, stdio-isolation, post-overlay XOR, sentinel oracle) and the rev-2 implementation review over the full diff (found one garbled doc sentence and a dead `setup_section` arm — both fixed — plus three test-strength upgrades, all in).
- **Trial run**: this PR is being driven end-to-end by **ds4 flash (DeepSeek V4 Flash)** as the working agent — first full PR under that model; flag anything that smells different from the usual arc.

💘 Generated with Crush

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -58,7 +58,10 @@ with a hosted synth; a call picks its team with the `cast` arg. Backends and cas
 `config.toml`, `KAIBO_*` env, then CLI flags (precedence: per-call > CLI > env > file >
 built-in); a missing config file is a non-error. See `docs/config.md`, and
 `docs/casts.md` for the backends/casts design rationale. kaibo never modifies the
-project and cannot run external commands.
+project and runs no external command except the one its own config declares —
+`api_key_cmd`, the operator-side key fetch: argv-only, stdin closed, time-boxed,
+output never logged, unreachable from the model-facing shell (see the operator
+surface rule in the invariants).
 
 ## Invariants — do not weaken without a failing-first test
 
@@ -106,6 +109,13 @@ project and cannot run external commands.
   projects, so surfacing it to a cast is a cross-project leak. Concretely — **model-facing
   kaish never grows `jobs`/`ps` or any kaibo-state builtin**; operator job/state visibility
   lives only in the tools (MCP, CLI, later the REPL), never in the shell we hand a model.
+  The one operator-side *resolve* surface — the config-declared `api_key_cmd` (a backend's
+  key from a command like `op read`; note this is a READ surface, unlike the store/CAS
+  record/emit surfaces) — follows the same line: config-gated, argv-only (no shell),
+  stdin closed, `KEY_CMD_TIMEOUT`-bounded, output never logged, and reachable only
+  through `Backend::resolve_key` at client construction, which the model-facing kaish
+  (sandbox lever 3, unchanged) can never reach. The model can *trigger* it by calling a
+  tool on a cmd-backed backend, but cannot control its argv or environment.
 - **Persistence engine (turso) — do not weaken.** Pure-Rust `turso`, **exact-pinned**
   (`=0.7.x` — `.db-tshm` format and API both drift between releases) with
   **`default-features = false`** (defaults pull mimalloc's global-allocator hijack + fts)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,14 @@ record. Each later release appends a new section at the top.
 - **`[telemetry] capture_content`** — off by default, so exported spans carry no prompts,
   completions, or tool payloads.
 - **`[telemetry] capture`** — export one named attribute without exporting all of them.
+- **Backend keys can come from a command** — `[backends.<name>] api_key_cmd = ["op",
+  "read", "op://Vault/Item/Field"]` runs the declared command (no shell, stdin closed,
+  30s ceiling, output never logged) and uses its trimmed stdout as the API key: a
+  1Password / `pass` / `gh auth token` credential without a secret in a file.
+- **Key sources are declared, never seeded** — the built-in backends no longer read
+  `ANTHROPIC_API_KEY` / `~/.anthropic-key.txt` / etc. out of the box; an operator names
+  the source (`api_key_env`, `api_key_file`, or `api_key_cmd`) in config.toml once, and
+  `api_key_file` + `api_key_cmd` together is a load error naming both.
 
 ### Changed
 
@@ -84,6 +92,12 @@ record. Each later release appends a new section at the top.
   rendered poll and as `submitted` in `batch get --json`.
 - **A failed batch now names the cast that ran it**, the provenance footer a completed
   batch already carried.
+- **Setup guidance steers toward declaring a key source in config.toml** (with
+  `kaibo example-config` for the shape) instead of exporting an env var.
+- **Breaking: an upgrading operator must add a key source to every backend they use** —
+  the built-ins no longer seed one, so a cast built on an undeclared backend silently
+  drops out of the roster; add `api_key_env`/`api_key_file`/`api_key_cmd` per backend to
+  bring it back (`kaibo example-config` shows the shapes).
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -188,6 +188,7 @@ and keep stdin open:
 claude mcp add kaibo -- docker run --rm -i \
   -u "$(id -u):$(id -g)" \
   -v "$PWD:/work:ro" \
+  -v "$HOME/.config/kaibo:/config:ro" -e XDG_CONFIG_HOME=/config \
   -e DEEPSEEK_API_KEY \
   ghcr.io/tobert/kaibo:latest
 ```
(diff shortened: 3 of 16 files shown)


tobert:
## Once-over review (Claude Opus 5, independent of the two GLM passes)

Verified claims against the code and ran the branch on a real box rather than reading the diff alone. **The design is sound and the invariant carve-out is honest** — including that it names its own residual ("the model can *trigger* it… but cannot control its argv or environment"), which is the right way to write one. Confirmed independently:

- The only non-test caller of `resolve_key_from_cmd` is `Backend::resolve_key` (`config.rs:732`).
- `kaish-kernel` is still `default-features = false, features = ["localfs"]` — `subprocess` is not compiled in, so the model-facing shell cannot reach exec.
- `src/sandbox.rs`, `tests/sandbox.rs`, and `tests/no_write_path.rs` are untouched.
- `read_capped` buffers `cap + 1` so exactly-cap passes and cap+1 refuses, and drains the rest so a verbose child can't wedge the pipe. Correct.
- Stdin-null is the right call and the reason is the important one: a child must never read the MCP stdio stream.
- clippy `--all-targets` clean; **1126 tests pass**.

Five findings, ordered by what I'd act on first.

### 1. Needs a rebase — the branch is behind three merges to `main` (blocker, mechanical)

`main` has moved: #157 (kaish-kernel → 0.14.1), #158 (dashscope), #159 (user-agent). This branch still pins `kaish-kernel = "0.14.0"`. A trial merge conflicts in **`CHANGELOG.md`** and **`tests/config.rs`** (both additive — resolve as the union in stack order). `src/credentials.rs`, `src/server/mod.rs`, and the docs auto-merge, but note #158 added a `dashscope` provider kind and a third media kind to the `CAST_ENUM_RULES` string, so re-check those hunks after resolving.

### 2. `api_key_cmd` runs per arm construction, not once — the doc says otherwise

`credentials.rs` says the command is "run once", and `config.rs:731` repeats "run once… resolved by the same lazy rule as the file (first client build)". There is no memoization: `Backend::resolve_key()` is called at every client build (`consult/engine.rs:399,411,426,442,457`, `media.rs:226,241`, `discover.rs:171`, `batch.rs:756,770,1308,1321`), and `Arm::from_slot` builds one arm per role. So a single `consult` spawns the command **at least twice** — once for the synth arm, once for the explorer.

For a key file that's free. For `op read` it means two subprocess spawns and ~600ms on the critical path of every call, plus **a 1Password audit-log entry per arm per call**, which is the part an operator would find surprising. Either memoize per backend for the process lifetime, or correct the doc to say "per client build" and note the cost. I'd lean memoize — the file source is already effectively cached by being cheap.

### 3. The stderr tail can carry the secret into an error string

On a nonzero exit the last 512 chars of the child's stderr go into the error (`credentials.rs:462-484`), which reaches logs and the MCP client. The module deliberately excludes *stdout* for exactly this reason — but a wrapper script run with `set -x`, or a tool that echoes the value while failing, puts the secret on stderr. Narrow and operator-controlled, and I would not block on it, but "output never logged" is the module's stated posture and this is the one path where output *is* surfaced. Worth a sentence in `config.md` telling operators not to trace a key script.

### 4. `resolve_key_from_cmd` panics on empty argv

`Command::new(&args[0])` (`credentials.rs:408`) indexes without checking. Unreachable through config — `config.rs:1800` refuses `api_key_cmd = []` at load, and that's tested — but this is a `pub fn` where an error return is already the idiom. One line.

### 5. The `tests/attach.rs` failures the PR body excuses did not reproduce

Full `cargo test` is green here, all 1126. The tracker does carry that flake as a known P1, so this isn't a contradiction — but the "green except two known failures" note shouldn't be carried into the merge commit if they now pass. Worth one re-run before merging.

### Operational note (not a defect)

Declared-only is working exactly as designed, and the failure is loud and actionable — a call to an undeclared backend gets *"no key source declared — set api_key_env, api_key_file, or api_key_cmd in config.toml"*, and the handshake roster correctly **drops** the unusable casts rather than advertising them. That is the right shape.

Worth knowing before merge: on a box relying on the old built-in seeding, `anthropic`, `deepseek`, and `gemini` lose their key sources, which takes every cast built on them out of the roster until three stanzas are added. Ran it against a real config to confirm — ten casts went away. Cheap to fix, but it is a migration, not a no-op upgrade.

tobert:
## Cross-family review of the fix commits (post-rebase, `0e18e00..a653097`)

Second kaibo review pass, this time over the rebase + the fixes for the 2026-08-15 findings. Cast `crusoe` (explorer `deepseek-ai/Deepseek-V4-Flash`, synth `zai/GLM-5.2`); citations spot-verified against the branch before posting.

**Verdict: no issues found — the four fixes are correct and secure.**

### 1. `resolve_key_from_cmd` — no secret leakage on any error path

Every error return in `src/credentials.rs:459-567` formats only `program` (`args.first()` — the command *name*, a config-declared pointer, never the secret) and numeric values. Stdout and stderr *content* never appear in any error string:

| Error path | What it formats |
|---|---|
| Empty argv | static string only |
| Spawn failure | `program` |
| Timeout | `program`, timeout duration |
| Nonzero exit | `program`, `{status}`, **`err.len()` — byte count only, stderr content suppressed** |
| Oversize stdout | `program`, `KEY_CMD_MAX_OUTPUT` |
| Non-UTF-8 stdout | `program` |
| Empty stdout | `program` |

Anyhow context chains are clean: `resolve_key_where` (`src/config.rs:739-740`) adds only the backend name. `read_capped`'s captured vectors are used only via `.len()`. `key_status` never runs the command at all, so no offline path can leak.

### 2. Empty-argv guard — consistent with the module's refusals

The `pub fn` guard (`src/credentials.rs:464-469`) names what was refused, why, and what to do (`kaibo example-config`), matching the module's refusal style. It is defense-in-depth beside the load-time guard (`src/config.rs:1867-1874`), which deliberately leaves `[""]` to fail at resolve time like any missing binary. Neither panics.

### 3. Doc claims match the traced call frequency

Traced `resolve_key`'s callers: one call per arm build (`consult/engine.rs`, `media.rs`, `server/resolver.rs`, `batch.rs`, `discover.rs`), arms built per cast role, no memoization anywhere. All three doc sites (`src/credentials.rs`, `src/config.rs`, `docs/config.md`) now agree: **once per role, not once per call** — e.g. a two-role `consult` on a command-backed key is two `op read` spawns and two audit-log entries.

### 4. Declared-only is uniform across every `ProviderKind`

`backend_template` (`src/config.rs:2351`) is the single construction point and seeds all three key-source fields `None` for every kind. The conventional names (`ANTHROPIC_API_KEY`, `.gemini-api-key`, DashScope's own) still exist in `credentials.rs` but are never referenced from backend construction — they survive only in the opt-in live-probe helper. Tests pin no-seeding for `openai-images`, `dashscope`, and `gemini-images` specifically (four assertions each), and only `ProviderKind::Openai` is `key_optional`.

### 5. Breaking bullet

The CHANGELOG bullet accurately describes the migration: no seeded sources → a cast on an undeclared backend drops from the roster → fix is one `api_key_env`/`api_key_file`/`api_key_cmd` per backend, `kaibo example-config` for the shape.

---

Also confirmed on the rebased tree: the PR body's `tests/attach.rs` excuse is stale — all 10 attach tests pass.

PR #158, opened 2026-08-15
feat(dashscope): generate images on Alibaba's wan family

tobert:
Adds `kind = "dashscope"`, kaibo's third media kind, so `generate` can run on an Alibaba DashScope subscription. Closes the gap that made image generation reachable only through Stability, hosted OpenAI, or a local sd-server.

## Why this shape

**Probed before written.** Every claim below was measured against a live endpoint, each with a negative control, because an all-clear that cannot fail means nothing:

- The OpenAI-compatible `/images/generations` path **does not exist** on a DashScope host — it answers with the same `No static resource` error as a route I invented. So `kind = "openai-images"` pointed at DashScope was never an option.
- The documented **async** `text2image/image-synthesis` route is an account entitlement this subscription lacks (`403 current user api does not support asynchronous calls`, against a bogus route's `400 Model not exist` — the route is real, the permission isn't). It is also async-only: without the header it 400s.
- The **synchronous** `multimodal-generation/generation` route reaches `wan2.6-t2i` and is what kaibo speaks. No polling lane needed.

**Media-only on purpose.** The same DashScope host serves text on an OpenAI-compatible endpoint kaibo already speaks, so a qwen model there belongs on a `kind = "openai"` backend with that base URL. A reasoning slot pointed at `dashscope` is a load error that says exactly this.

## The decision this PR rests on

DashScope delivers artifacts as **presigned links, never inline bytes**. kaibo had a standing rule against fetching artifact URLs, stated in `openai_images.rs` and pinned by a test asserting the phrase "never fetches".

Amy's call when it surfaced:

> we fetch URLs all the time in inference, this is similar, so we should tune the language and add the fetch into cas. otherwise the user has to dink with curl and raw tokens.

That rule was right for the case it was written for and wrong as a blanket. Where base64 is available, a URL means the request lost `response_format` and refusing is free. Where links are the only delivery, refusing just moves the same network hop into the operator's shell with the key on a command line. And a content-addressed store cannot address what it has not read — the digest *is* the address, so bytes are the artifact's existence condition, not a convenience.

So: the refusal is now scoped to the kind that means it, its message points at the parameter that fixes it, and its test asserts the blanket claim is **absent**.

## The fetch

`cas::fetch_artifact_bytes` — beside the store rather than in the provider, so the next URL-delivering provider inherits the bound and the refusals instead of restating them. `MediaArtifact` is unchanged and still carries bytes; pushing the fetch past the provider seam would have churned Stability and openai-images for nothing.

- **https only, on every hop.** The second commit is a fix found while reviewing my own first: reqwest follows up to ten redirects and `https_only` defaults to **false**, so an https link redirecting to http was followed and the up-front scheme check only ever described the first hop. `tls::https_only_client` closes it, added in the one place kaibo builds reqwest clients so ring is still installed at every site. Kept separate from `https_client` — everywhere else the endpoint is an operator's own `base_url`.
- **Bounded twice**: the declared `Content-Length` before any body is read, and the running total while it is. Empty bodies refused.
- **No new write surface** — bytes route through the existing `store_generated_artifacts` → `MediaStore::put`, so `no_write_path.rs` stays at exactly 5 blessed lines.

**Mime comes from the fetch response's `Content-Type`**, and one the media store cannot name is refused rather than defaulted to png — a wrong extension under a real digest is a permanent mislabel, and the URL's suffix is a hint no operator can correct.

## Measured facts that shaped the request

`enable_interleave` must be seeded or a text-only prompt is refused; `n` is 1–4; `size` is `"W*H"` bounded by total pixels (589824–2073600); and **unknown parameter names are silently accepted**, so kaibo sends only what a caller wrote rather than leaning on the provider to catch a typo.

## Tests

Pure build/parse units, an offline transport test over a real socket, and an `#[ignore]`d live probe walking generation → link → bytes → artifact that checks the magic bytes actually match the claimed mime.

Two tests were verified by breaking the thing they guard:

- Seeding and the fetch guard: broke both, watched the right three fail.
- The TLS-on-every-hop test **initially passed for the wrong reason** — a listener that accepts and hangs up makes both clients error, so it stayed green with `https_only` removed. Rewritten as a differential: the fixture serves a real 200 and asserts the ordinary client gets it while the fetch client does not. Confirmed failing without the flag.

## Gates

`cargo clippy --all-targets` clean · 1131 tests pass · rustfmt clean on every touched file · `cargo tree -i` empty for `aws-lc-rs` and `mimalloc` (with a positive control proving the command works) · `no_write_path` unchanged at 5 blessed lines · live probe green against a real subscription.

## Also

Corrects a doc comment in `openai_images.rs` that had `key_optional` backwards from shipped behavior.

## Review

Cross-family review by Gemini via kaibo is running; results will be added as a comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,17 @@ record. Each later release appends a new section at the top.
 
 ## [Unreleased]
 
+### Added
+
+- **`kind = "dashscope"`** — a media backend for Alibaba's wan image family, so
+  `generate` can run on a DashScope subscription.
+
+### Changed
+
+- **kaibo now fetches a generated artifact's URL when that is how a provider delivers
+  it**, over TLS and size-bounded, instead of refusing it — otherwise the operator
+  fetches the link by hand with their key.
+
 ### Fixed
 
 - **Bump `kaish-kernel` to 0.14.1** — the explorer's shell no longer drops piped or
--- a/README.md
+++ b/README.md
@@ -517,7 +517,8 @@ config has three concepts for configuring models:
 
 - **backend** — a *connection*: which wire protocol (the completion kinds `anthropic`
   | `deepseek` | `gemini` | `openrouter` | `openai`, or the media kinds `stability` |
-  `openai-images` behind the `image` role), base URL, and where its key comes from.
+  `openai-images` | `dashscope` behind the `image` role), base URL, and where its key
+  comes from.
   Secrets never live in the TOML — only the *name* of an env var or the path to a key
   file. `openrouter` is a keyed gateway with a fixed endpoint — one key reaching every
   major model family, reasoning on by default via its unified `effort` param.
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -358,7 +358,7 @@ scratch_limit_bytes = 67108864
 # ---------------------------------------------------------------------------
 # [backends.<name>] — CONNECTIONS only: `kind` picks the wire and its client. The
 # completion wires (anthropic | deepseek | gemini | openrouter | openai) serve the
-# reasoning slots; the media kinds (stability | openai-images) serve `image` slots
+# reasoning slots; the media kinds (stability | openai-images | dashscope) serve `image` slots
 # only. The rest describes the wire (endpoint, key source, per-request deadline).
 # Models live on cast slots, never here.
 #
@@ -368,7 +368,8 @@ scratch_limit_bytes = 67108864
 # (required for a new backend), the anthropic and gemini kinds (optional —
 # unset still dials rig's built-in api.anthropic.com / generativelanguage.googleapis.com),
 # and the media kinds (optional — unset dials api.stability.ai for stability,
-# api.openai.com/v1 for openai-images); set it to point the same wire at a
+# api.openai.com/v1 for openai-images, dashscope-intl.aliyuncs.com for dashscope);
+# set it to point the same wire at a
 # compatible gateway/proxy (or a local server) instead — a ROOT in every case,
 # since the client appends its own path. Every other keyed kind rejects
 # base_url at load (rig fixes those endpoints).
@@ -483,7 +484,7 @@ wire = "responses"
 # `deep-dive`) so it can route by name without reading this file.
 #
 # Roles: explorer, synth (the reasoning phases), and image (the media member — it
-# points at a media backend, kind stability or openai-images, and staffs `generate`;
+# points at a media backend, kind stability, openai-images, or dashscope, and staffs `generate`;
 # see the image-slot examples below). A typo'd role is a loud load
 # error. A cast may omit roles — the tool that needs a missing role fails at call time,
 # naming the gap ("cast `notes` has no synth slot"); absent = capability absent.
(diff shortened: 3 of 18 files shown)


tobert:
## Cross-family review — Gemini 3.5 Flash via kaibo (`cast=gemini`)

Ran against the branch with `src/dashscope.rs` and the transport test attached. Verified every finding against the cited code before acting on any of it — three held, one was already fixed, two I declined.

### Confirmed and fixed

**Mime should not come from the `Content-Type` header.** The strongest finding, and I went further than suggested. Two real failure modes, both confirmed: `cas::Extension::from_mime` has no `image/jpg` arm (`src/cas.rs`), and object stores commonly serve JPEG under exactly that spelling; and a bucket sending `application/octet-stream` or no type would have had a valid PNG refused. Both would be kaibo's bug, not the provider's.

The review proposed header-first with a sniff fallback. I made the **bytes the sole authority** — `view_image::sniff_mime` recognizes exactly png/jpeg/gif/webp, the same four the CAS stores as an image, and bytes cannot be wrong about themselves the way a header can. The header is retained only for the error message. New test covers the case that motivates it: an error page served under `Content-Type: image/png`, which the header-driven version would have stored as a png.

**Redirects should not be followed at all.** I'd already closed the TLS-downgrade half independently (`ece7376`, `https_only`), but the review's SSRF point stands — following a chain lets a provider response steer kaibo at any reachable host. Now `Policy::none()`. Verified before adopting rather than assumed: DashScope's presigned OSS links answer `200` directly with a correct `Content-Type`. A redirect now gets its own `FetchError::Redirected`, reported ahead of the generic status check, because "a presigned link is short-lived, generate again" is the wrong diagnosis for a 302.

**A 2xx carrying an error payload.** Speculative — I never observed it, every probed error came with a proper 4xx — but cheap and the failure mode is real. Placed in the envelope-missing branch rather than as an up-front check so it cannot misfire on a valid response; pinned by a test with both an envelope and a top-level `code`.

**Error-string actionability.** `NoImages` and `InvalidBody` now name what to do next.

### Declined, with reasons

**`Vec::with_capacity(declared)`** would let a server force a 64 MiB allocation by declaring one and sending a single byte — it makes the memory posture *worse* in the name of improving it. Amortized growth is already bounded by the running check.

**`checked_add` on the running total** is unreachable. `bytes.len()` is capped at `MAX_FETCH_BYTES` on every iteration, so the sum cannot approach `usize::MAX`. The review acknowledged as much ("network buffer limits prevent a single chunk from being large enough").

### Independently confirmed

Its extension-point sweep matched mine — no missed enumeration. Worth noting the class-based design is why: `config.rs` validates on `ProviderClass::Media(_)` rather than a kind list, so a new media kind inherits base-URL legality, role/slot agreement, and diagnostics for free. The only hand-maintained strings were `CAST_ENUM_RULES`, the `generate` description, the orientation prose, README, and the two docs files — all updated.

It also endorsed all-or-nothing on partial fetch failure, which I'd argued for on the grounds that a short list is a failure the caller cannot notice.

### Note on one test

An assertion I wrote to cover the redirect policy could not fail — it merely constructed both clients. Deleted rather than kept; the offline coverage gap is now stated in `tests/tls.rs` instead. `https_only` means the fetch client cannot be pointed at a local plaintext fixture, so the redirect path is exercised only by the `#[ignore]`d live probe.

Gates after the fixes: clippy clean · 1135 tests · rustfmt clean on touched files · live probe green against a real subscription.

PR #160, opened 2026-08-15
feat(telemetry): honour the OTEL_* environment, and redact content by default

tobert:
First of two. This one makes traces **safe**; the follow-up adds the metrics signal, which is what makes traces *optional*.

## The split

kaibo now answers two questions separately instead of conflating them:

| | knob | default |
|---|---|---|
| Do spans leave at all? | `enabled` | on when `OTEL_*` names an endpoint |
| Do they carry content? | `capture_content` | **false** |

Before this, telemetry was all-or-nothing and therefore off — kaibo reads private source and rig's spans carry prompts, completions, and source snippets, so anyone who wanted latency and token numbers had to ship their code to a collector to get them. With content redacted a trace carries model ids, token counts, durations, finish reasons, and exit codes. That's safe enough that telemetry can turn itself on when the environment already names a collector.

## Following the spec rather than inventing a scheme

Pulled the GenAI semantic conventions (they moved to their own repo) and implemented their rule: *"instrumentations SHOULD NOT capture them by default, but SHOULD provide an option for users to opt in."* Their env spelling `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` is honoured verbatim, so an operator who already sets it elsewhere gets the same behaviour here.

| Variable | Effect |
|---|---|
| `OTEL_EXPORTER_OTLP_ENDPOINT` | A **root** — signal path appended. Setting it enables telemetry. |
| `OTEL_EXPORTER_OTLP_{TRACES,LOGS}_ENDPOINT` | **Full URLs**, used verbatim. |
| `OTEL_SERVICE_NAME` | `service.name` |
| `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` | content opt-in |
| `OTEL_SDK_DISABLED` | off, beating everything |

Root-vs-full-URL is the spec's own rule and would 404 against a real collector if inverted, so both directions are tested.

**Precedence: `KAIBO_*` / CLI > `config.toml` > `OTEL_*` > off.** This inverts kaibo's usual env-beats-file rule for one family, deliberately: `OTEL_*` is *ambient*, set by a platform for its own collector rather than aimed at kaibo, so it may supply what you left blank and never override what you wrote. An explicit `enabled = false` is absolute. `OTEL_SDK_DISABLED` is the one exception that beats everything — a kill switch something else can override is not a kill switch.

## Why a filter, and why an allowlist

The redaction is applied on export, not at the call sites, and that's **forced rather than chosen**: `gen_ai.prompt`, `gen_ai.input.messages`, `gen_ai.tool.call.arguments`, `gen_ai.tool.call.result` and their siblings are emitted inside **rig**. kaibo cannot decline to set them, only decline to export them. So `otel_filter` wraps the exporter — the same transparent-wrapper shape `completion_watch::Watched` and `wire_repair::Repaired` take.

**Allowlist, not denylist**, because the failure directions aren't symmetric. A denylist fails **open** the day a dependency emits a new content attribute — `cargo update` starts shipping source with nothing in the diff to notice. An allowlist fails **closed**: an unblessed name is never exported and being wrong costs a missing dashboard field. Same reasoning `no_write_path.rs` uses in pinning an exact blessed count rather than enumerating what's forbidden.

**Three surfaces, not one** — an allowlist covering only the first would leak through the others:

1. Span attributes.
2. **Event attributes** — kaibo's own log lines name paths (`running kaish: cat -n src/auth.rs`).
3. **The error status description** — where a provider's response body lands. The conventions draw this line the same way: `error.type` safe and kept, `exception.message` sensitive and dropped.

Two spec gotchas are encoded rather than left to be rediscovered: `gen_ai.tool.description` carries a sensitivity warning but is `Recommended`, not `Opt-In` (it's on the safe list here for a kaibo-specific reason — our tool descriptions are published product text — and the comment says so); and `gen_ai.retrieval.documents` is `Opt-In` with *no* warning, gated for volume rather than privacy, so the two axes aren't conflated.

## Verification

Not just unit tests. A **real consult** exported to a capturing collector through `OTEL_EXPORTER_OTLP_ENDPOINT` alone. The raw OTLP protobuf:

```
MUST NOT APPEAR                SHOULD APPEAR
✓ absent  MAX_FETCH_BYTES      ✓ present 12  gen_ai.usage
✓ absent  pub const            ✓ present  2  gen_ai.request.model
✓ absent  cat -n               ✓ present  5  qwen3.8-max
✓ absent  In one sentence      ✓ present  2  gen_ai.tool.name
✓ absent  You are the          ✓ present  2  service.name
```

The same run with content opted in **does** put the answer text on the wire — the differential proving the search was looking somewhere real.

`tests/otel_filter.rs` is the canary for the surfaces rig doesn't populate on that path: a marker planted in all three, asserted absent redacted and present opted-in. Each of the three scrubs was sabotage-checked independently (disable one → the right test fails). The env precedence tests were checked the same way.

`kaibo://config` reports `capture_content` and a one-line `content_policy`; the startup log prints the same sentence beside the endpoint, so someone who enabled this through an ambient variable can see what's leaving.

Gates: clippy clean · 1155 tests · rustfmt introduces no new debt (the pre-existing count in `config.rs`/`tests/config.rs` is unchanged).

## Next

Metrics — `gen_ai.client.token.usage`, `gen_ai.invoke_agent.{duration,inference_calls,tool_calls}`, `gen_ai.execute_tool.duration`. Content-free by construction (confirmed: no metric in the spec references any content attribute), so opting out of tracing entirely still leaves cost, latency, and delegation counts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ record. Each later release appends a new section at the top.
 
 - **`kind = "dashscope"`** — a media backend for Alibaba's wan image family, so
   `generate` can run on a DashScope subscription.
+- **kaibo honours the standard `OTEL_*` environment**, and exports to
+  `OTEL_EXPORTER_OTLP_ENDPOINT` without further configuration.
+- **`[telemetry] capture_content`** — prompts, completions, and tool payloads are
+  redacted from exported spans unless you opt in.
+- **`[telemetry] capture`** — admit individual attributes by semantic-convention name.
 
 ### Changed
 
@@ -27,6 +32,9 @@ record. Each later release appends a new section at the top.
   fetches the link by hand with their key.
 - **Every outbound request now identifies itself as `kaibo/<version>`** — kaibo's
   traffic previously carried no `User-Agent` at all.
+- **Telemetry enables itself when an OTLP endpoint is in the environment**, which is
+  safe because content is redacted by default. An explicit `enabled = false` still wins,
+  and `OTEL_SDK_DISABLED` beats everything.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,11 +19,10 @@ record. Each later release appends a new section at the top.
 
 - **`kind = "dashscope"`** — a media backend for Alibaba's wan image family, so
   `generate` can run on a DashScope subscription.
-- **kaibo honours the standard `OTEL_*` environment**, and exports to
-  `OTEL_EXPORTER_OTLP_ENDPOINT` without further configuration.
-- **`[telemetry] capture_content`** — prompts, completions, and tool payloads are
-  redacted from exported spans unless you opt in.
-- **`[telemetry] capture`** — admit individual attributes by semantic-convention name.
+- **kaibo reads the standard `OTEL_*` environment**, including `OTEL_SDK_DISABLED`.
+- **`[telemetry] capture_content`** — off by default, so exported spans carry no prompts,
+  completions, or tool payloads.
+- **`[telemetry] capture`** — export one named attribute without exporting all of them.
 
 ### Changed
 
@@ -32,9 +31,10 @@ record. Each later release appends a new section at the top.
   fetches the link by hand with their key.
 - **Every outbound request now identifies itself as `kaibo/<version>`** — kaibo's
   traffic previously carried no `User-Agent` at all.
-- **Telemetry enables itself when an OTLP endpoint is in the environment**, which is
-  safe because content is redacted by default. An explicit `enabled = false` still wins,
-  and `OTEL_SDK_DISABLED` beats everything.
+- **Telemetry turns on when `OTEL_EXPORTER_OTLP_ENDPOINT` is set** — safe now that
+  content is redacted by default.
+- **An explicit `[telemetry] enabled = false` beats the environment**, and
+  `OTEL_SDK_DISABLED` beats every source.
 
 ### Fixed
 
--- a/docs/config.example.toml
+++ b/docs/config.example.toml
@@ -259,20 +259,44 @@ scratch_limit_bytes = 67108864
                                 # POSIX-unrestricted. A typo here is a load error.
 
 # ---------------------------------------------------------------------------
-# [telemetry] — OpenTelemetry export: traces and logs. OFF BY DEFAULT, and that
-# default is load-bearing: kaibo reads private source, and the spans rig emits carry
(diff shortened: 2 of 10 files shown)


tobert:
## STE pass — `f06d86c`

Applied kaish's `docs/style.md` (which kaibo adopts by reference) to the prose this branch added, grooming at the point of touch.

**American spelling.** `honours` / `honoured` / `behaviour` → `honors` / `honored` / `behavior`. Both the guide's rule and what the corpus already does — `config_resource.rs` and `config.rs` say "honor" in five places, so I'd introduced an inconsistency as well as a violation.

**Two metaphors cut**, both in the class the guide singles out — naming a mental act as a physical one, which a reader working from partial context or a second language cannot recover:

- *"the skeleton of what happened survives without the flesh"* → an event keeps its name and timestamp, because the record that something happened is not content.
- *"the last place kaibo owns"* → "the export step, which kaibo does own".

I checked three words before cutting them rather than after: `ambient`, `owns`, and `on the wire` are already in the corpus with one meaning each (`config.rs:718`, `docs/config.md:29`/`:1258`, `config.rs:5334`), so they stay as terms. The guide's own instruction is to count the uses before arguing.

**Table cells are complete clauses.** The `OTEL_*` table had fragments — *"A root; kaibo appends `/v1/traces`"* — which make the reader reconstruct the verb, and a model reading one row out of context can't. Every cell now names kaibo as the subject and says what it does with the value.

**Template shorter, guide longer** — the split `CLAUDE.md` asks for. The template says what to type; the allowlist rationale moved to `config.md`, with the template pointing at `kaibo://config/guide`. Also dropped *"an attribute nobody blessed"*: **blessed** already means something specific here (an exemption from the write-path guard), so reusing it for allowlist membership reads as a second concept.

**One mechanism leak removed** — the guide named `src/otel_filter.rs`, which an operator can't act on and doesn't need in order to predict behavior.

**CHANGELOG**: one line, one reason per bullet. Two bullets carrying three sentences became four.

**The published diagnostic got the most work**, since it's full weight — it's a string a tool returns, and it shows up in `kaibo://config`:

```
before:  prompts and responses INCLUDED (52 attributes allowed)
after:   content exported — 62 attributes, including prompts and responses
         content redacted — 52 attributes, no prompts or responses
```

Both spellings are parallel, both name what is *absent* rather than only that a policy is in force, and both state the count — the number is what someone checks against a change they just made. Its test now asserts those properties instead of matching a substring.

Gates unchanged: clippy clean · 1155 tests · rustfmt debt still at the pre-existing 6.

PR #166, opened 2026-08-22
feat(media): twelve Stability operations behind an `op` parameter

tobert:
## What

Three wired Stability routes become **fifteen**. `generate` now reaches every synchronous `stable-image` route outside the generate family: six `edit`, four `control`, and the two sync `upscale` routes.

## Decisions

- **The operation is a property of the call, not the model slot.** A cast's `image` slot names one model and Stability has 25 operations behind it, so the route cannot come from the model id. `op` selects; the model id keeps its old job of choosing within the generate family.
- **A table, not twelve enum variants.** These routes differ only in a path segment, so `STABLE_IMAGE_OPS` is one source for the schema list, the parser, the refusal, and the costs — the same argument `artifact::FORMATS` and `upload::SIGNATURES` record.
- **Costs are published to the model.** One credit ≈ one US cent, and `upscale/conservative` is **20×** `generate/core`. A model that reads the price before choosing spends differently. A test pins the spread so a table edit that flattened it fails rather than quietly making the numbers not worth their space.
- **Two structural guards at the single dispatch point.** `accepts_ops` defaults false like `accepts_inputs`; an unwired `op` refuses inside `media_request` instead of falling through to the model's default route (which made `media_request` fallible, as it should have been).
- **`prompt` is not sent to routes that document none.** `edit/remove-background` and `upscale/fast` declare no `prompt` property. This resolves a note the code had deferred — "noted rather than solved speculatively" — which wiring these families made due.

Deliberately absent, each for a reason stated in the table's doc: `upscale/creative` and `edit/replace-background-and-relight` are **deferred**, so wiring them as sync would mis-declare `Shape` — the one thing this module refuses to sniff from a response. Audio and 3D need CAS extensions for `mp3`/`wav`/`glb` that don't exist yet.

## Two bugs I caught in my own work, both by mechanical checking

**`mask` was listed as required** on `edit/erase` and `edit/inpaint`. The spec requires `image` alone; a mask is optional and an alpha channel on the image is the documented alternative. Under a field named for what a route *demands*, that told a caller to go find something it never needed — a wrong instruction published to a model, producing no error to discover it by.

Caught by verifying all twelve rows against the live OpenAPI spec rather than by eye: path exists, a 2xx carries `image/*` (so `Shape::Sync` is a fact), `inputs` equals the spec's required-and-binary fields, `credits` matches the route's own section. **That check's first run parsed zero rows** (rustfmt had split the table) and reported "0 mismatches" — a clean result from a check that examined nothing. It now asserts it saw all twelve before concluding anything.

**A genuine spec contradiction, recorded rather than smoothed over:** `edit/erase` lists `prompt` as *required* while declaring no such property. `takes_prompt` stays `true` there — the required list is the half that produces a 400 if the server enforces it. **This wants a live probe to settle.**

## Review

kaibo, cast `crusoe` (GLM-5.2). It confirmed routing, `model`-field handling, both guards and the error strings, and found a doc bug I had published to models: the `op` doc listed `search_prompt`/`select_prompt` as `inputs` keys when they are text `fields`. A model following it would have put a prompt where a digest goes and been refused for a reason that was our fault.

**So a test now pins the doc to the table** — this was the second bug in two commits from prose describing a table by hand. schemars renders a doc comment, not a table, so the doc can't be generated; it can be checked. Verified against a sabotaged build: changing one published price to 41 fails it.

## Testing

18 tests in `tests/media_inputs.rs` plus stability-level routing tests. Verified against sabotaged builds three times: unknown-`op` fallthrough, removing the arm's op guard, and a stale published price each failed exactly their own test and no others.

Full suite green (1222 passing, exit 0), clippy clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,13 +22,14 @@ record. Each later release appends a new section at the top.
   MCP client.
 - **`generate` reaches Stability's edit, control and upscale routes** through a new `op`
   parameter — twelve operations where three were wired.
-- **Each operation's credit cost is published on the `op` parameter**, so a model sees
-  that `upscale/conservative` costs twenty times `generate/core` before it picks.
+- **Each operation's cost is published on the `op` parameter** in the provider's own
+  unit, unconverted — so a model sees the twenty-fold spread before it picks a route.
 - **An unrecognized `op`, or one on a backend with no operations, is refused** rather
   than run as a text-to-image generation.
-- **`generate` takes input images by digest** — `inputs {"image": "<digest>"}` is
-  image-to-image on Stability's `ultra` and `sd3` routes, reusing an image already in
-  the store rather than re-sending it.
+- **`generate` takes input images by digest** — `inputs [{"field": "image", "digest":
+  "..."}]` is image-to-image on Stability's `ultra` and `sd3` routes, reusing an image
+  already in the store rather than re-sending it. A field name may repeat, for the
+  operations that take several source images under one name.
 - **`generate` carries several named input parts**, so an operation taking
   `init_image`+`style_image`, or an image plus an optional mask, is reachable.
 - **A media backend with no input-image route refuses a call that carries one**, rather
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ record. Each later release appends a new section at the top.
 - **`kaibo cas write FILE`** stores an image from the command line and prints its
   digest — the CLI half of `write_cas`, so an operator can put an image in without an
   MCP client.
+- **`generate` reaches an OpenAI-compatible images backend's `edits` and `variations`
+  routes** through the same `op` parameter, over multipart where `generations` is JSON.
 - **`generate` reaches Stability's edit, control and upscale routes** through a new `op`
   parameter — twelve operations where three were wired.
 - **Each operation's cost is published on the `op` parameter** in the provider's own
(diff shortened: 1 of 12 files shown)


tobert:
## Two changes from the provider survey, before this ships

Amy asked for a look at how much of this design is tied to Stability versus general, before building more on it. I pulled the live specs (Gemini's discovery doc, OpenAI's `openapi.yaml`). Most of the design generalizes — `MediaRequest{prompt,fields,inputs}`, `MediaInput.mime` (Gemini's `Blob` is literally `{mimeType,data}`), `MediaOutcome::{Complete,Deferred}` (already models `predictLongRunning`), and the `accepts_inputs`/`accepts_ops` seams. But it surfaced two things worth fixing before `generate`'s schema is public.

**`inputs` is now a list, not a map.** `BTreeMap<String, String>` cannot express two entries under one key — the second silently replaces the first, a dropped input wearing a success. Stability never needs it (every form field has a distinct name), so the map worked by luck of Stability's shape. OpenAI's `/images/edits` takes "one or more source images" as repeated `image` parts. A list also preserves the caller's order, where a map sorted by key.

**Costs are the provider's own unit, verbatim.** Amy's ruling: kaibo can't keep up with the pricing landscape, so it reports what the provider says and lets the client agent do the arithmetic. `credits: u32` was wrong in shape, not just units — Stability bills flat credits, OpenAI per image by size and quality, Gemini per image. `cost: &'static str` converts nothing.

Two new tests for the shape the map couldn't hold: two parts sharing one field name both survive, and the caller's order reaches the provider unsorted.

Next up, separately: OpenAI `edits` + `variations` (a three-row table on the same `op` mechanism — the test of whether `op` generalizes off Stability), then Gemini.

PR #168, opened 2026-08-22
feat(gemini): image generation through `generate`, like every other backend

tobert:
## What

A `gemini-images` backend kind. Gemini image generation reaches the existing `generate` tool, and from the caller's side nothing distinguishes it from Stability or the Images API.

Amy's call on the shape: *"couldn't we drive that in the background and make it look like the other image generators?"* — so **no new tool**. I had proposed a separate conversational lane; this is better. It costs no resident tool-description budget, reuses the whole media path, and leaves the genuinely conversational case where it belongs: *"if the image model really is conversational, it's a consultation having it create images."*

## A media kind whose wire is a completion wire

Gemini has no images endpoint. Image generation is `models/{model}:generateContent` — the same call as text — with `generationConfig.responseModalities` naming what comes back. So `gemini-images` is a separate `ProviderKind` from `gemini`, exactly as `openai-images` is from `openai`: one vendor, two uses of one endpoint, and a cast slot has to know which. A reasoning slot aimed here would resolve a completion model that answers in pictures.

## One seam change, and it isn't Gemini-specific

`MediaOutcome::Complete` becomes `{ artifacts, note }`. A `generateContent` response interleaves text and image parts, and a model that says *"I moved the sign left, the original crop cut it off"* has told the caller something the bytes don't.

**That channel is the TTS door.** `responseModalities` is an enum of TEXT/IMAGE/**AUDIO**, and `generationConfig` carries `speechConfig` beside `imageConfig`. Gemini TTS is this module's shape with a different modality — so the part-walk is written for parts in general, not images in particular. When that wave lands it should be a config block and a CAS extension, not a new module.

## The most valuable error carries the model's own words

A safety refusal, a clarifying question and "I can't do that" all arrive as a normal 200 with text and no picture. Reporting "no image" without the text would throw away the only explanation the caller gets.

## Review

kaibo, cast `crusoe`. Its best finding was one I'd have hated to ship: **`generate`'s description had become false** — "Stability accepts input images; the other media backends do not" — in text a calling agent reads to decide what to send. A stale description is bad; an actively wrong one is worse, because it's trusted.

Also fixed: eight stale enumeration sites in `docs/config.md` (embedded in the binary, served as `kaibo://config/guide`), including the per-kind `base_url` section, so an operator couldn't discover how to configure this at all. And a real API gap — **a blocked prompt was reported as "no candidates"**. Gemini refuses at two levels (`promptFeedback.blockReason`, and a candidate with `finishReason` and no content); both now carry the reason. Plus: every candidate is walked, not just the first, since `candidateCount` rides `fields` and each candidate is billed.

Two review items I didn't take as given: `imageConfig.imageSize` **is** real (`512`/`1K`/`2K`/`4K` in the discovery doc), and on the "third media kind" comment, **mine** was the wrong one — DashScope genuinely was third, Gemini is fourth.

## Testing

19 tests, including two over a real socket for truths the pure functions can't see: the model rides in the *path* with a `:generateContent` suffix, and the key rides a header rather than the `?key=` query parameter Google also accepts, so it stays out of anything that logs a URL.

Writing the unknown-parts test immediately caught a bug in the line I'd just written — the recording sat after the blob branch, so blob parts were recorded as unknown and genuinely unknown parts were skipped by an early `continue`.

Full suite green (1251 passing, exit 0), clippy clean.

## Wants a live probe

`responseModalities` defaults to `["TEXT", "IMAGE"]` because the spec makes it an exact match and errors on an unsupported combination, and image-only isn't universally supported. **That's my reading of the spec, not a verified fact** — it, and Stability's `edit/erase` prompt contradiction, are the two open questions a small amount of real spend would settle.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **A `gemini-images` backend kind** — Gemini image generation through `generate`, the
+  same face as every other image backend. Takes input images; has no named operations.
+- **`generate` shows what the model said** above the digests, for a backend whose image
+  generation is a conversation and answers with words as well as bytes.
+
 - **`kaibo cas write FILE`** stores an image from the command line and prints its
   digest — the CLI half of `write_cas`, so an operator can put an image in without an
   MCP client.
--- a/README.md
+++ b/README.md
@@ -518,7 +518,7 @@ config has three concepts for configuring models:
 
 - **backend** — a *connection*: which wire protocol (the completion kinds `anthropic`
   | `deepseek` | `gemini` | `openrouter` | `openai`, or the media kinds `stability` |
-  `openai-images` | `dashscope` behind the `image` role), base URL, and where its key
+  `openai-images` | `gemini-images` | `dashscope` behind the `image` role), base URL, and where its key
   comes from.
   Secrets never live in the TOML — only the *name* of an env var or the path to a key
   file. `openrouter` is a keyed gateway with a fixed endpoint — one key reaching every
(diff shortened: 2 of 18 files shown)


tobert:
some examples https://claude.ai/code/artifact/bba0b683-833f-498b-8e88-d00259db934a

PR #170, opened 2026-08-31
feat(bfl): FLUX image generation through generate

tobert:
Adds a `bfl` backend kind — Black Forest Labs' FLUX image family — as the fifth media kind behind `generate`. Same face, no new tool: `[backends.bfl]` plus a cast with an `image` slot, five starter operations (`flux-dev`, `flux-2-pro`, `flux-2-flex`, `flux-pro-1.1-ultra`, `flux-kontext-pro`).

## The decisions

**Why FLUX.** The ask was Midjourney-style concept images. Midjourney has no lawful API — its ToS (2026-05-27) bans automated access, and the third-party "Midjourney API" wrappers puppet Discord accounts. FLUX is the buildable version of that request: keyed, documented, and shaped for kaibo's media lane. Checked live before building, not from memory.

**Every BFL operation is asynchronous**, unlike any existing kind: a create returns `{id, polling_url, cost}` and the artifact comes from polling. The shape chosen is a hybrid — `generate` polls in-call for a 30s budget (FLUX is seconds-fast; a concept-image caller wants bytes, not a handle) and falls back to the existing `Deferred`/`job-N` lane past that. The spec (`docs/bfl.md`) offered exactly these two shapes; this is the first with the second as its bound, not a third invention.

**The poll leg is TLS-strict, and that came from review.** The first cut polled the provider-chosen `polling_url` through the general client with the `x-key` credential attached — but that client follows redirects (reqwest strips `Authorization` on cross-host redirects, not custom headers) and permits plaintext. `polling_url` is the same trust shape as `result.sample`, plus the credential, so it now gets the `artifact_fetch_client` posture: https-only, zero redirects, refused before a byte leaves. Fixed failing-first — the pre-fix code demonstrably dialled a plaintext listener; the fixed code was refused with the listener seeing zero connections.

**Cost is the provider's own unit, per request.** `AsyncResponse.cost` (credits) rides the result verbatim — no static cost table to go stale, per the 2026-08-22 native-units ruling. The deferred path has no note channel today, so cost is dropped on the rare slow path; recorded in the module doc as a gap a shared `Provenance` cost field would close for every kind.

**Field spellings are per-backend, and BFL ignores unknowns silently** — measured on the first live call (an `aspect_ratio` changed nothing). The `fields` doc now names BFL's own spellings (`width`/`height`, `seed`, `safety_tolerance`) and says so.

**Out of scope, on record** (`docs/bfl.md`): `flux-3-video` (CAS has no video extensions yet — refused loudly today), the `flux-tools/*` image family (a table extension later), finetunes, and webhooks (kaibo is stdio-only and receives no callbacks; `webhook_url`/`webhook_secret` are refused at the tool boundary before money is spent).

## Validation

Offline: 34 unit tests (every terminal status, request/response parsing, the inline-step decision), 3 real-socket transport tests (verbatim `polling_url` dial proven via TCP-accept under the strict client; plaintext refused with a zero-connections assertion and the create leg proven to have run), 5 config tests, one media-arm construction test. Live, with a real key: `generate --cast flux` returned a real artifact into the CAS on the inline path, cost reported as BFL's own 3 credits, and the strict poll client worked against the real regional polling host. Full suite green, clippy clean, `cargo tree -i aws-lc-rs` / `-i mimalloc` both empty, `tests/no_write_path.rs` marker count unchanged.

A cross-family kaibo review of the branch follows as a PR comment.

Co-authored-by: Claude Sonnet 5 (implementation), reviewed in-session by Claude Fable 5 (the poll-leg finding).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ record. Each later release appends a new section at the top.
 
 ### Added
 
+- **A `bfl` backend kind** — Black Forest Labs' FLUX image family through `generate`,
+  five named operations, polled in-call for a fast generation and as a `job-N` handle
+  only past that budget.
 - **A `gemini-images` backend kind** — Gemini image generation through `generate`, the
   same face as every other image backend. Takes input images; has no named operations.
 - **`generate` shows what the model said** above the digests, for a backend whose image
--- a/README.md
+++ b/README.md
@@ -520,7 +520,7 @@ config has three concepts for configuring models:
 
 - **backend** — a *connection*: which wire protocol (the completion kinds `anthropic`
   | `deepseek` | `gemini` | `openrouter` | `openai`, or the media kinds `stability` |
-  `openai-images` | `gemini-images` | `dashscope` behind the `image` role), base URL, and where its key
+  `openai-images` | `gemini-images` | `dashscope` | `bfl` behind the `image` role), base URL, and where its key
   comes from.
   Secrets never live in the TOML — only the *name* of an env var or the path to a key
   file. `openrouter` is a keyed gateway with a fixed endpoint — one key reaching every
--- a/docs/bfl.md
+++ b/docs/bfl.md
@@ -25,6 +25,19 @@ media lane. The motivating use is concept images.
   returned `polling_url` verbatim (regional host — never reconstruct it from
   the base URL) until `ResultResponse.status` leaves
   `Pending`/`Reasoning`/`Generating`.
+- **The poll leg is https-only with no redirects.** `polling_url` is an
+  address BFL chose, not the operator's own `base_url`, and the poll request
+  carries the `x-key` credential — the same trust shape
+  `crate::cas::fetch_artifact_bytes` already takes for `result.sample`, and
+  worse, since the key rides along. A plain `http://` `polling_url` is
+  refused before a single byte of the request leaves (no connection is
+  attempted at all — `crate::tls::artifact_fetch_client`'s `https_only(true)`),
+  and a redirecting `polling_url` is refused rather than followed
+  (`Policy::none()`) — reqwest strips the standard `Authorization` header on a
+  cross-host redirect but not a custom header like `x-key`, so following one
+  would carry the key to whatever host it names. The create call itself keeps
+  the general client: `base_url` is the operator's own configured endpoint,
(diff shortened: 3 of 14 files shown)


tobert:
## Cross-family review

kaibo cast `crusoe` (explorer `deepseek-ai/Deepseek-V4-Flash`, synth `zai/GLM-5.2`), six review areas, citations spot-verified before posting.

**Verdict: merge-ready.** Two minor findings; the actionable one is already fixed on the branch (`d5558e9`).

1. **Credential paths — clean.** The `x-key` key never reaches an error string, log, or Debug render: `BflClient`'s Debug is manual ("never render the key", `finish_non_exhaustive`), `BflImageModel` derives no Debug at all, and `src/bfl.rs` contains zero tracing/log statements. The poll leg is TLS-strict (scheme pre-check, `artifact_fetch_client`, post-response 3xx check), proven by the zero-connections plaintext-refusal transport test. The create leg's lax client is the documented operator-trusts-operator split — same client every provider uses, no new exposure. The `result.sample` fetch carries no key at all.
2. **Poll loop termination — correct.** Monotonic budget guarantees exit; every terminal status propagates via `?` on the spot; no double-poll; a `Ready` landing on the same iteration the budget expires still wins (`inline_step` checks Ready before budget).
3. **Op table — correct.** Five ops match `docs/bfl.md`'s table path-for-path; an unknown op is refused with a string naming the asked op, the valid set, and the omit-`op` default — never run as a guessed route.
4. **Error strings** — four of five terminal statuses followed the three-part refusal rule; `GenerationError` stopped at the provider's detail with no action clause. **Fixed in `d5558e9`** ("The provider ran the request and failed on its own side — retry the generation").
5. **Inputs — correct encoding; arity provider-enforced.** Base64 per named field, CAS read-not-rewritten, order preserved. The "up to 8" input limit is BFL's API claim, not a kaibo-side check — deliberately consistent with Stability and OpenAI Images, whose limits are also provider-enforced. Noted, no change.
6. **Filesystem writes — zero new sites.** No `std::fs` in `src/bfl.rs`; `tests/no_write_path.rs`'s blessed-line count unchanged at 5.

Also validated live before this PR went up: a real-key `generate --cast flux` returned an artifact through the inline path, BFL's native cost (3 credits) in the result, strict poll client working against the real regional polling host.

PR #173, opened 2026-09-01
chore(kaish): upgrade the read-only shell to kaish-kernel 0.17.0

tobert:
Jumps `kaish-kernel` 0.14.1 → 0.17.0, inheriting three releases' breaking changes at once. Amy: *"once kaish 0.17 drops we'll upgrade and release kaibo 0.4."*

**Stacked on #172** (the one-brace fix that makes `tests/config.rs` compile again). Base retargets to `main` automatically when that merges.

## Compile-time exposure was one line pair

Exactly what the 2026-08-21 pre-bump audit predicted. `IgnoreScope` became `#[non_exhaustive]` in 0.16 and broke the exhaustive match in `server/config_resource.rs`. The wildcard arm names the unrecognized variant rather than rendering a default label, mirroring the parse direction in `config::merge_kaish`, which refuses an unrecognized `scope` outright rather than falling back.

Everything else was free: kaibo builds no `ExecContext` literal (0.17's other BREAKING item), uses no plan-side redaction API, walks no AST, and 0.16's `execute → Result<_, KernelError>` flows through `anyhow`'s `?` unchanged because `KernelError: Error`.

## The compiler found less than the shell did

One compile error — but **four behavioral changes reached the model-facing surface**, and only running the shell found them. Two made kaibo's own prose false:

- **0.16 made `grep -r` prefix hits with the operand as written**, matching GNU. The `grep -rn PATTERN .` idiom our sandbox addendum teaches therefore started emitting `./src/foo.rs:12` for every citation the explorer earns. Measured:

  ```
  grep -rn needle .          →  ./d/a.txt:1:alpha needle     ← what we taught
  grep -rn needle            →  d/a.txt:1:alpha needle       ← bare operand
  grep -rn needle top.txt    →  1:gamma needle               ← no filename at all
  ```

  We now teach the bare form, the only one that always yields a repo-relative `file:line`. The old sentence's promise that the idiom works *"whether the target is a file or a directory"* went with it — the file case drops the filename, which is the half a citation needs.

  **Deliberately not a kaish ask.** 0.16's behavior is GNU-exact on purpose; asking kaish to diverge for our convenience would trade a correct rule for a cosmetic one.

- **0.16 replaced the bare `command not found`** for a refused external command with a message that names the refusal. The addendum quoted the old text.

Two more arrived free through `kaish-help`, which kaibo composes rather than restates: compound statements now feed pipes, and `yes`/`no` stopped being lexer errors.

## The sandbox boundary moved, and we accept it on a measured condition

0.17's lstat-by-default means a symlink inside the project pointing outside now renders its **target path string** through `ls -l`/`stat`/`readlink`/`find -type l`, where 0.14 refused outright.

We accept this rather than narrow it, because the property that makes it safe was measured rather than assumed — **there is no existence oracle**:

| link target | `cat` / `stat -L` / `file` / `wc` | `[[ -e ]]` |
|---|---|---|
| `/etc/hostname` (exists) | `permission denied: path escapes root` | false |
| `/etc/DEFINITELY-NOT-HERE` (missing) | *byte-identical refusal* | false |
| `/root/.ssh/id_rsa` (unreadable) | *byte-identical refusal* | false |

The refusal is decided by path arithmetic **before any syscall reaches the target**, so a hostile repo gets back exactly the string it wrote into its own link — nothing about the host, not even existence. And a link's target is bytes stored inside the allowed tree, so reading it *is* reading project content; refusing would make `ls -l` misdescribe a directory kaibo is allowed to list.

`mount_layer_symlink_discloses_its_target_string_but_no_host_fact` states both halves and fails loudly if the second ever weakens. Its leak assertion has a recorded positive control: point the link at an in-tree file carrying the marker and the assertion fires.

## Full sandbox battery A–G re-run live

A kernel/VFS bump trips the probe trigger. **All clear.** Three pass criteria in `docs/sandbox-probes.md` were false against 0.17 and are corrected in place (Battery A's `ln -s` refusal reason, Battery B's 127 message, Battery C's now non-empty `env`). Battery G is new for the symlink boundary.

Worth recording: **0.16 fixed an `env` that bypassed the external-commands gate, and kaibo was never exposed.** Lever (0) compiles `subprocess` out, so the host-spawn path it escaped through does not exist here — verified against both versions. That is the four-levers design paying for itself.

## Gates

- `cargo build`, `cargo clippy --all-targets` clean
- **1320 tests pass, 0 fail** (the one known `tests/credentials.rs` ETXTBSY parallel flake passes serially and reproduces on unmodified code)
- `cargo tree -i` empty for `aws-lc-rs`, `mimalloc`, `openssl-sys`
- All six kaish crates at 0.17.0 in lockstep in `Cargo.lock`

## Sent upstream, not fixed here

Two kernel bugs found while probing, reported to kaish-lead for 0.17.1:

1. **`!` binds only in a condition position**, contradicting the 0.16.0 changelog's claim that `! true && true` is `(! true) && true`. `if ! false` and `while ! true` work; `! true && echo x` is a parse error at 1:1.
2. **`readlink -f` fails on every operand** on a rooted LocalFs mount — an in-root regular file, an in-root link, a dangling link, and a missing file all return `No such file or directory: /tmp`, naming neither the operand nor its target. Plain `readlink` is fine.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -465,19 +465,43 @@ even for a one-line doc fix.
   reads that release's `.sha256` sidecars and pushes with your own `gh`/git auth, so
   there's no CI secret to rotate. Deliberately manual: releases are human-cut, so this
   is the ritual's last step, not a workflow job.
-- **kaish pin.** Currently `kaish-kernel = "0.14.1"`. The `0.14.0 → 0.14.1` bump is a
-  patch release — **zero** call-site changes, `cargo build`/`clippy --all-targets`/full
-  `cargo test` (749 passed) all clean, `cargo tree -i aws-lc-rs` and `-i mimalloc` both
-  empty, no **BREAKING** entries. Every changelog entry is a fix inside the shell
-  language itself, which model-facing kaish inherits for free with no Rust-side change:
-  `${s[0:5]}` string slicing (characters, not bytes; `${s:0:5}` — bash's different
-  convention — is now a loud error instead of silently vanishing); a pipeline's first
-  stage, `read`, and `grep`/`cat`'s streaming path no longer drop or skip buffered
-  stdin; `read` on empty stdin now fails instead of binding `""`; and `exit`/`return`/
-  `break`/`continue` now carry a block's already-printed output out instead of
-  discarding it. Net effect for kaibo: scripts the explorer already writes (a `read`
-  into a pipeline, an early `exit` inside a loop) now behave the way bash would, where
-  before they silently lost output or stdin.
+- **kaish pin.** Currently `kaish-kernel = "0.17.0"`, a 0.14.1 → 0.17.0 jump that
+  inherits three releases' breaks at once. **Compile-time exposure was one line pair**,
+  exactly as the pre-bump audit predicted: `IgnoreScope` went `#[non_exhaustive]` in
+  0.16, breaking the exhaustive match at `server/config_resource.rs`. The wildcard arm
+  we added names the unrecognized variant rather than rendering a default label,
+  mirroring the parse direction in `config::merge_kaish`, which refuses an unrecognized
+  `scope` outright. Everything else was free: kaibo builds no `ExecContext` literal
+  (0.17's other BREAKING item), uses no plan-side redaction API, walks no AST, and
+  0.16's `execute → Result<_, KernelError>` flows through `anyhow`'s `?` unchanged
+  because `KernelError: Error`.
+- **The lesson from the 0.17 bump: the compiler found less than the shell did.** One
+  compile error, but four *behavioral* changes reached the model-facing surface, and
+  only running the shell found them. Two made kaibo's own prose false and were fixed
+  in `kaish_syntax.rs`: 0.16 made `grep -r` prefix hits with the operand as written, so
+  the `grep -rn PATTERN .` idiom we teach started emitting `./src/foo.rs:12` for every
+  citation (the bare form is now taught, and the old sentence's claim that the idiom
+  works "whether the target is a file or a directory" went too — a named file drops the
+  filename, which is the half a citation needs); and 0.16 replaced the bare
+  `command not found` for a refused external command, which the addendum quoted. Two
+  more arrived free through `kaish-help`, which kaibo composes rather than restates —
+  compound statements now feed pipes, and `yes`/`no` stopped being lexer errors. **When
+  you bump kaish, diff the rendered contract and run the shell; do not stop at a green
+  build.** A throwaway crate that calls `compose(&Recipe::tool_description(), …)` under
+  both versions diffs the composed contract in one command.
+- **The 0.17 bump also moved the sandbox boundary, and the runbook with it.**
+  lstat-by-default means a symlink inside the project pointing outside now renders its
+  target path string through `ls -l`/`stat`/`readlink`/`find -type l`, where 0.14
+  refused. Accepted rather than narrowed, on a measured condition: existing, missing,
+  and unreadable targets refuse **byte-identically**, so there is no existence oracle
+  and a hostile repo learns only the string it wrote into its own link. A link's target
+  is bytes inside the tree, so reading it is reading project content. Pinned by
+  `containment.rs::mount_layer_symlink_discloses_its_target_string_but_no_host_fact`
+  and by Battery G in `docs/sandbox-probes.md`. Also worth knowing: 0.16 fixed an `env`
+  that bypassed the external-commands gate, and **kaibo was never exposed** — lever (0)
+  compiles `subprocess` out, so the host-spawn path it escaped through does not exist
+  here. That is the four-levers design paying for itself, and it is the kind of
+  evidence worth recording when it happens.
 - **The previous pin, kept because the reasoning pattern is the point.** The
   `0.13.0 → 0.14.0` bump moved **one** call site — a deletion: `kaish_syntax.rs`'s
   `strip_write_side_paragraphs`, whose target text kaish-help #297 made opt-in via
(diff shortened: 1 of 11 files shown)


tobert:
## Cross-family review — kaibo cast `crusoe` (DS4-Flash explorer, GLM-5.2 synth)

Dogfooded per the project's pre-merge rule: a different model lineage than wrote the change, aimed at the whole change rather than the diff. Every citation was verified against the tree before acting on it — presence is not support.

**Two real findings, both fixed in `0f8a058`. Two claims I checked and did not take.**

### Finding 1 (must fix, taken) — the grep fix was applied to one surface and missed five

I changed `KAISH_SANDBOX_ADDENDUM` and stopped. The review found the same `grep -rn PATTERN .` idiom in five other model- and operator-facing surfaces:

| surface | who reads it |
|---|---|
| `src/consult/prompts.rs` — explorer preamble | **every sweep** |
| `src/server/mod.rs` — MCP `run_kaish` tool description | the calling agent, resident |
| `src/server/mod.rs` — `kaibo://tools` doc | the calling agent, on demand |
| `src/kaish_syntax.rs` — `kaibo://kaish/sandbox` recipes | the model, on demand |
| `src/cli.rs` — `kaibo kaish -c` help | the operator |

**Fixing one was worse than fixing none.** A model reads the addendum every turn *and* the explorer preamble every sweep, so a half-fix taught two contradictory idioms with no way to tell which was right — and the dotted one produces exactly the `./src/foo.rs:12` citations this change existed to prevent. The review's framing of this as the headline finding is correct and I've recorded it as the durable lesson in `AGENTS.md`: when a kaish idiom changes, grep for it, because it is never in one place.

Also taken: `kaibo://kaish/sandbox`'s exit-code list still quoted the pre-0.16 bare `command not found`.

Deliberately **not** changed, and the review agreed on the first: test fixtures that merely contain the idiom (`progress.rs`, `consult/engine.rs`) are inputs, not teaching. And a named-directory operand (`grep -rn PATTERN DIR/`) is correct and stays — GNU joins the operand as written, so a named directory still cites a usable path. It is `.` specifically that adds the `./`.

### Finding 2 (should fix, taken) — the containment test promised three existence states and built two

The doc comment claimed an existing target, a missing one, and an unreadable one are indistinguishable; the test created only `to_existing` and `to_missing`. That gap was load-bearing, because the no-existence-oracle property is the *justification* for accepting the target-string disclosure at all — and as the review put it, a permission-shaped difference probes the host just as well as an existence-shaped one.

Added the third arm (a mode-000 target), with a recorded positive control: point it at an in-tree file and the assertion fires.

```
left:  "exit: 1 ... permission denied: path escapes root: <TARGET> is not under /tmp/..."
right: "exit: 0 --- stdout --- inside"
```

### Checked and not taken

- **`IgnoreScope` wildcard arm** — review concurred it is correct as written. The renderer returns a plain `String` with no error channel, `#[non_exhaustive]` requires *some* arm, and `unreachable!()` would panic on a variant a future kaish release may legitimately add. Naming the variant is the option that neither hides the gap nor invents a label.
- **Timing oracle** — raised as a theoretical uncovered channel. Not taken: the refusal is decided by path arithmetic before any syscall reaches the target, which is *why* there is nothing to time. A timing assertion in a unit test would be flaky and would pin the implementation rather than the property.
- **`--json` line anchors, `$(cmd)` typed values, `ls -R`** — review found no kaibo impact; I confirmed independently. `orientation.rs` is the only place kaibo parses kaish JSON (`glob`, `stat`, `wc -l`), none of which carry line anchors, and its tests run through the real kernel.

### Gates after the fixes

1319 pass, 0 fail, plus the known `tests/credentials.rs` ETXTBSY parallel flake that passes serially and reproduces on unmodified code.

tobert:
## The release waits for kaish 0.17.1 — this PR does not

Amy's call, verbatim: *"we're not gonna release kaibo until we get on 0.17.1, so those symlink issues won't ship."*

**What merges vs. what ships.** This PR merges and `main` goes to kaish 0.17.0, so development continues on the new pin. **v0.4.0 does not get cut until the pin is 0.17.1.** The split is deliberate: a merge is reversible and a release is not.

**What it keeps out of a release.** The bump's own probe run found `readlink -f` broken on *every* operand on a rooted LocalFs mount — an in-root regular file, an in-root link, a dangling link, and a missing file all return `No such file or directory: /tmp`, naming neither the operand nor its target. Plain `readlink` is fine. Reported upstream; kaish-lead has it for 0.17.1 and has been told it is now release-blocking for us rather than nice-to-have.

To be clear about which symlink item this is: the *target-string disclosure* documented above is *not* a bug and is not what we are waiting on — it is measured, accepted, tested, and documented in Battery G. It is `readlink -f` being broken that holds the release.

**The rule this produced** (`61d345b`, in AGENTS.md's **Cutting a release**):

> "Current" is not the whole gate: do not release on a kaish version carrying a known unfixed bug in a surface kaibo hands a model.

The old gate said to confirm the pins are "current", and that was not strong enough here — this bump's probe run is what *found* the bug, so "we are on the current pin" and "we are on a known-broken pin" were true at the same moment. Being on the newest tag is not the same as being on a good one. Shipping it anyway would have been a decision rather than an oversight, which is the distinction the rule now names.

## Rebased onto the merged main

#171 (transient-transport retry) and #172 (the brace) are both in `main` now, so this rebased onto `7b126c2`. Re-ran the gates against the rebased tree rather than trusting GitHub's MERGEABLE:

- `cargo clippy --all-targets` clean
- **1327 tests pass, 0 fail**
- the known `tests/credentials.rs` ETXTBSY parallel-exec flake still appears on some parallel runs and passes serially; it reproduces on unmodified code and is unrelated to this change

PR #174, opened 2026-09-02
chore(kaish): bump the read-only shell to kaish-kernel 0.17.1

tobert:
0.17.1 carries the fix that has been blocking kaibo v0.4.0: `readlink -f` and
`realpath` failed on **every** operand on a rooted mount, with a message naming
neither the operand nor its target (`readlink: o-exists: No such file or directory:
/tmp`). Both resolve now, and an escape refuses by name. We reported that shape
upstream during the 0.17.0 bump; it became two structural kaish PRs rather than a
patch — [#435](https://github.com/tobert/kaish/pull/435) (the router's mount-ancestor
synthesis was unreachable whenever `/` is mounted) and
[#434](https://github.com/tobert/kaish/pull/434) (`Filesystem::canonicalize` as a
defaulted trait method, moving containment out of the builtin into the VFS).

## The compiler found nothing, so the shell was the check

No API break reached kaibo — `Filesystem::canonicalize` is defaulted and kaibo
implements no `Filesystem` or `KernelBackend`. The composed tool contract is
byte-identical under both pins (throwaway crate calling
`compose(&Recipe::tool_description(), …)`), and so are all eleven `help` surfaces
kaibo renders, 43941 bytes each side.

So the check was the shell, per the rule the 0.17.0 bump wrote into AGENTS.md. Every
battery in `docs/sandbox-probes.md` was run against **both** a 0.17.0 and a 0.17.1
binary and diffed. A, B, D, E, F and G came back identical. Two changes are the whole
delta a model can see.

**1. The release blocker is fixed.** `readlink -f` / `realpath` resolve an in-tree
path (exit 0) and refuse an escape by name. Battery G3 re-run on the new canonicalize
path: existing, missing, and unreadable targets still refuse byte-identically, so the
new code introduces no existence oracle.

**2. The directories above the mount list again — new, and accepted.** 0.17.0 answered
"not found" for every directory above the project; 0.17.1 restores them as synthesized
directories. It is synthesis from kaibo's own mount table, not a host read: each level
names only the next component down to the project, so a model walking up recovers the
root path string the caller already handed it and nothing else. Counted rather than
argued — `ls /tmp` returns one entry where the host `/tmp` holds 3575. Adjacent
secrets, real siblings, and the state db and media CAS all stay invisible (E2/F2
re-run against a non-empty store: 4 KiB db, 210 CAS objects, both `not found`).

That observable is pinned by a new containment test with a **recorded positive
control** — move the mount up and the leak assertion fires — because a battery where
everything comes back empty proves nothing otherwise. Battery C's claim that `/home`
is "an inert stub that cannot be walked" is now false and is corrected in place, the
same way the 0.17.0 bump corrected three other criteria.

## The probe caught itself once

E1 run without `--root` created a state db, because the fixture was then outside every
allowed tree and the guard correctly did not fire. Re-run with the root pointing at the
fixture, both E1 and F1 refuse loudly and create nothing. The §0 question — *would this
read differently if the probe were broken, versus if the thing it audits were broken?*
— is what found it.

## Gates

- `cargo clippy --all-targets`: clean.
- `cargo test`: 1147 passed. The lone failure is the known `tests/credentials.rs`
  ETXTBSY exec race under parallelism (green serially, reproduces on unmodified code).
- containment: 25 passed, one new.
- `cargo tree -i` empty for `aws-lc-rs`, `mimalloc`, `openssl-sys`.

Cross-family review posted as a comment below.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -467,7 +467,7 @@ even for a one-line doc fix.
   reads that release's `.sha256` sidecars and pushes with your own `gh`/git auth, so
   there's no CI secret to rotate. Deliberately manual: releases are human-cut, so this
   is the ritual's last step, not a workflow job.
-- **kaish pin.** Currently `kaish-kernel = "0.17.0"` (from 0.14.1, inheriting three
+- **kaish pin.** Currently `kaish-kernel = "0.17.1"` (from 0.14.1, inheriting four
   releases' breaks; compile-time exposure was one `#[non_exhaustive]` match arm).
   **When you bump kaish, run the shell — a green build is not the check.** That bump's
   four behavioral changes all reached the model-facing surface and none broke
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,9 +81,11 @@ record. Each later release appends a new section at the top.
 
 ### Changed
 
-- **kaish upgraded to 0.17.0** (from 0.14.1) — the read-only shell gains symlink
+- **kaish upgraded to 0.17.1** (from 0.14.1) — the read-only shell gains symlink
   support, pipeline-stage compound statements, `set -o pipefail`, and base-aware
   arithmetic.
+- **`readlink -f` and `realpath` resolve a path again** — both failed on every operand
+  before, naming neither the operand nor its target.
 - **`grep -rn PATTERN` is what kaibo teaches now**, without the trailing `.` — kaish
   0.16 prefixes hits with the operand as written, so the bare form is the one that
   yields repo-relative `file:line` citations.
@@ -93,6 +95,8 @@ record. Each later release appends a new section at the top.
 - **`ls -l`, `stat`, `readlink`, and `find -type l` describe a symlink itself** instead
   of following it, so a link is visible as a link. A link pointing outside the project
   shows its target path; every read that would follow it out is still refused.
+- **The directories above the project list again** — each names only the next component
+  down to the project, so the shell shows the root path the caller already gave it.
 - **A compound statement can feed a pipe** — `for f in …; do …; done | grep x` is no
   longer a parse error.
 - **`yes` and `no` are ordinary strings**, not lexer errors, so `echo yes` runs.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1799,9 +1799,9 @@ dependencies = [
 
 [[package]]
 name = "kaish-glob"
-version = "0.17.0"
+version = "0.17.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8c827be08f68998d484f1f8ab75bc9286d5af73914e68922e3307df172b70c5f"
+checksum = "93656ec4bb59fa7c6cba023bfe1c3acdea85df527fee974de9aba753887c20ff"
 dependencies = [
  "async-trait",
  "ignore",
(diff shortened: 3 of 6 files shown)


tobert:
## Cross-family review — kaibo cast `crusoe` (DeepSeek-V4-Flash explorer, GLM-5.2 synth)

Asked for a holistic read of the bump, with the safety argument for the new observable
called out as the finding I most wanted to be told about. It came back **ship it**, with
three real gaps. Every citation was checked against the code before acting; all three
were true and all three are fixed in `e16c2db`.

**1. The test claimed the chain and probed one level.** The doc comment said "each
ancestor lists exactly the one component," but the fixture only had `outer/project` — so
a synthesis bug one level higher would have fired nothing. The fixture is now two levels
deep (`outer/mid/project`), each level carrying a sibling that must never appear.

**2. The gap that mattered: no existence probe.** The test covered bytes (`cat`, `grep`)
and discovery (`ls`, `find`), but not *existence*. `stat` and `realpath` answer "does
this exist" without returning any bytes — and `realpath` resolves for the first time in
this very release, which makes that pair the most plausible place for an oracle to have
arrived. Measured: a real host file beside the chain and a path that was never created
refuse byte-identically once the operand is normalized away, and only the synthesized
components resolve. Now pinned in the shape Battery G3 uses for links, and recorded in
the runbook note.

**3. A residual false claim in Battery C.** Its opening pass criterion still read
"everything outside the single mount comes back `not found`," which the carve-out two
lines below now contradicts. Scoped to *reads*, which is what it always meant.

It also asked me to confirm the changelog's "kaish upgraded to 0.17.1 (from 0.14.1)"
baseline — correct: 0.17.0 was never cut, the last release is `v0.3.0`, so the cumulative
framing from the last released pin is right.

The positive control was re-run against the rewritten test: move the mount up and the
leak assertion fires with `HOST LEAK: ls …/mid listed a real sibling`.

Not taken: nothing. All three findings were real.

tobert:
## CI note: the `tests/credentials.rs` flake, measured

The first CI run on this branch passed; the second failed on
`key_command_nonzero_exit_is_loud_and_never_leaks_stdout_or_stderr`
(`spawning key command .../stub-5`). That is the known `tests/credentials.rs` ETXTBSY
race, not this diff — but it is worse than our notes said, so I measured it rather than
asserting it:

| | parallel | serial |
|---|---|---|
| this branch | **6 / 20 failed** | 0 / 10 failed |
| `main` @ `f741a84`, unmodified | **3 / 15 failed** | — |

So: pre-existing, independent of the kaish bump, and a **~20–30% failure rate per run**.

**Shape.** `stub_script` (`tests/credentials.rs:208-215`) writes an `0o755` `#!/bin/sh`
stub which `resolve_key_from_cmd` then execs. Rust opens files `O_CLOEXEC`, but in the
window between another test thread's `fork` and its `exec` the child holds a duplicate of
*this* thread's write fd — and execing a file that is open for writing is `ETXTBSY`. It is
the harness, not the credential code.

**Fix shape, for its own PR:** a process-wide mutex the stub-using tests hold across both
`stub_script` and the `resolve_key_from_cmd` call, so a fork can never happen while a stub
write fd is open. That serializes ~10 tests rather than the suite. Retrying on `ETXTBSY`
would also go green but would hide a real race behind a loop.

I re-ran the job rather than papering over it. Flagging it because a merge gate that fails
a fifth of the time is one we will start ignoring — worth closing before v0.4.0.

PR #185, opened 2026-09-09
A gated tool refuses in kaibo's words, naming the knob

tobert:
Amy picked this off the tracker for 0.5.0 and added the shape: *"if we don't already I think the pattern where we return the list of valid tools on invalid tool calls is a nice touch"*.

We already do that one layer in — `UnknownToolFeedbackHook` answers a hallucinated tool call from the inner model team by naming the real toolset. But the gated-tool case is the **calling** agent hitting a route rmcp dropped, and it got rmcp's words.

**Stacked on #184** (`fix/worktree-follow-boundary`), which is where the changelog's `Unreleased` section already moved. Independent of it otherwise.

## The problem

A tool dropped by `--no-<tool>`, by staffing, or by the media store being off is *absent from the router*, so rmcp answers:

```
-32602: tool not found
```

Correct on the wire, and the same sentence a typo gets. From the caller's side those are indistinguishable — so a model reads "not found" and retries the spelling of a name that was never misspelled, when what it needed was either a word with the operator or a different tool.

## The decision: name the knob, and name what works

Four reasons, each pointing at a different knob:

| reason | what the refusal says |
|---|---|
| operator flag | `--no-deliberate` **and** `[tools] deliberate = false` — either one fixes it |
| media store off | `[cas] enabled = true`, with `kaibo example-config` for the shape |
| no cast can staff it | the cast shape it wants, and `kaibo://config` to compare casts |
| job verb, no producer | that nothing here mints a handle, and who the producers are |

**Every reason that applies is named, not the first one found.** `generate` can want both a cast and the media store; fixing one of two leaves the caller still blocked and no wiser.

The job-verb branch is the one that most needed writing: those verbs have no flag of their own, so a reader told only "not served" would go hunting for a `--no-job-get` that does not exist.

`ALL_TOOL_NAMES` gates all of it — a name kaibo never had falls through to the generic error, because there it is true.

## What it reads like

```
`deliberate` is a kaibo tool, but this server does not serve it. The operator turned it
off with `--no-deliberate` (`[tools] deliberate = false` says the same thing). Live here:
`consult`, `consult_submit`, `explore`, `job_cancel`, `job_get`, `job_list`, `job_wait`,
`list_models`, `oneshot`, `read_cas`, `run_kaish`, `write_cas`.
```

```
`job_get` is a kaibo tool, but this server does not serve it. Nothing here mints a job
handle, so there are no jobs to collect. The producers are `consult_submit`,
`batch_submit`, `deliberate` and `generate`; none of them is live. Live here:
`list_models`, `oneshot`, `read_cas`, `run_kaish`, `write_cas`.
```

```
`explore` is a kaibo tool, but this server does not serve it. No configured cast can
staff it: it needs a cast with an `explorer` slot. `kaibo://config` lists every cast and
what each one carries. Live here: `consult`, `consult_submit`, `job_cancel`, …
```

## Implementation

Writing `call_tool` on the impl suppresses the macro's version the same way the existing `list_tools` already does. The dispatch it wraps is unchanged — a real route reaches `self.tool_router.call` byte-identically; the only addition is a `has_route` miss consulting `gated_tool_refusal` first.

`gating_key_for` mirrors `ToolGating::enabled`'s match: that one answers *whether*, this one answers *which knob*. The live roster comes from `self.tool_router.list_all()` — the router itself, so it cannot describe a surface this server doesn't serve.

## Tests

The `mcp_stdio` test was **evolved rather than added to**, per the tracker note — it pinned rmcp's wording and now pins kaibo's. It was run against the pre-change behavior and fails there:

```
the refusal must name the tool and the flag that dropped it; got: Mcp error: -32602: tool not found
```

It is the only place that proves these words survive the trip over the wire, and it now also pins the *typo* half: `run_kiash` still gets the generic answer and no "Live here:". Without that, an override explaining every miss would look identical to one explaining the right ones.

Two unit tests in `src/server/mod.rs` cover the flag-off, unstaffable and job-verb branches — handler tests live there, and a `tests/` grep misses them.

**Reading the rendered refusals whole, not the diff, is what caught the one real bug**: the live roster sorted `consult_submit` ahead of `consult`, because `_` (0x5F) sorts before a backtick (0x60) and I was sorting the already-wrapped strings. Sort the names, then wrap.

Suite: **1341 passed / 0 failed / 23 ignored across 40 suites**, exit code captured directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,12 @@ record. Each later release appends a new section at the top.
   worktree must name its registration back, so one file in a repo you cloned to review
   cannot make any directory on the host readable.
 
+### Changed
+
+- **A tool this server does not serve now refuses in kaibo's words** — naming the flag, the
+  cast shape it wants, or the absent job producer, plus the tools that are live. A
+  misspelled name still reads as `tool not found`.
+
 ### Removed
 
 - **`explorer_max_turns` / `synth_max_turns` are no longer tool arguments or CLI flags** —
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -1001,6 +1001,24 @@ pub(crate) fn cast_requirement_for(name: &str) -> Option<&'static str> {
         .map(|(_, _, requirement)| *requirement)
 }
 
+/// The `[tools]` key that gates `name`, whose CLI spelling is `--no-<key with dashes>`.
+/// `None` for a tool with no flag of its own: the job verbs follow their producers, and
+/// the two CAS halves follow `[cas] enabled`. Mirrors [`ToolGating::enabled`]'s match —
+/// that one answers *whether*, this one answers *which knob*.
+fn gating_key_for(name: &str) -> Option<&'static str> {
+    match name {
+        "consult" | "consult_submit" => Some("consult"),
+        "explore" => Some("explore"),
+        "deliberate" => Some("deliberate"),
+        "oneshot" => Some("oneshot"),
+        "run_kaish" => Some("run_kaish"),
+        "batch_submit" => Some("batch"),
+        "list_models" => Some("list_models"),
+        "generate" => Some("generate"),
+        _ => None,
+    }
+}
+
 /// Every `#[tool]` route name kaibo can advertise — the fixed universe [`live_tools`]
 /// filters. `KaiboHandler::new` asserts each really exists on the router, so a renamed
 /// tool method fails the build rather than leaving a gate quietly inert.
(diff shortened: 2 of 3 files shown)


tobert:
Superseded by #186 — same branch, same commit (`46e3d6c`), retargeted at `main`.

This closed itself when #184 merged with `--delete-branch`: GitHub closes any PR whose base branch is deleted, and a closed PR cannot be reopened or retargeted while its base ref is missing. My mistake in the merge order — the child should be retargeted to `main` *before* the base merges, or the base branch kept until the stack is done.

PR #193, opened 2026-09-10
The history doc comment says what the history framing says

tobert:
#192 changed the session-history framing from re-verification to trust and acquisition, and left the doc comment above `consult_user_prompt` describing the behavior it replaced: "History prepends the prior `(question, answer)` pairs and steers the model to re-confirm any span a prior answer cited."

A standalone `explore` over the prompt-composition path found it, reported it as the one thing that looked stale in the area, and named how to confirm — the framing text and `session_history_asks_for_more_evidence_not_a_re_read` both pin the opposite.

Internal comment only: no published text, no behavior change, no test change. The two claims the old comment made that are still true — the exploration runs fresh every turn rather than replaying a stored report, and the files win over a prior answer where they disagree — stay.

### DIFF
--- a/src/consult/prompts.rs
+++ b/src/consult/prompts.rs
@@ -514,14 +514,16 @@ pub fn batch_system_prompt(override_: Option<&str>) -> String {
 }
 
 /// Build the consult driver's user prompt from the question, any caller-supplied
-/// `context`, and any prior session turns. Pure and offline-testable: this framing
-/// is the whole of the context-seed and multi-turn hand-off, so it's worth pinning.
+/// `context`, any prior session turns, and any attached files. Pure and
+/// offline-testable: this framing is the user-turn half of the context seed — the
+/// preamble installs the same posture for the phase — and the whole of the multi-turn
+/// hand-off, so it's worth pinning.
 ///
-/// With **no** context and **no** history this is exactly the bare question — a
-/// stateless, unseeded consult is byte-for-byte unchanged. Supplied `context`
-/// (a diff summary, a prior report, pasted source) is framed as *trusted starting
-/// evidence*: a grounded `file:line` rarely needs re-deriving, and the steer is to
-/// investigate for *more* when the context isn't enough — the CLAUDE.md acquisition,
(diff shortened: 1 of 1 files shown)


tobert:
Reviewed by kaibo, cast `deepseek` (explorer and synth both `deepseek-flash`, via the DeepSeek Platform key), pointed at this branch with `src/consult/prompts.rs` attached and the change described in prose rather than as a diff.

It agreed the rewritten sentence is now accurate, then turned the question back on the whole block — which is the rule here for prose — and found three more claims in it that were wrong, all predating the stale sentence this PR came from:

1. **The bare-question condition named two of its three terms.** The guard is `history.is_empty() && context.is_none() && attached.is_empty()`; an attach-only call is legal and is not the bare question. The doc above `empty_history_yields_the_bare_question` carried the same omission, and that test passes all three empty.
2. **The input inventory omitted `attached`** — the parameter that renders the largest block of the prompt.
3. **"the whole of the context-seed" overclaimed.** The context bytes ride only through this function, but the trust posture is installed twice: here and in the consult preamble's fourth paragraph. The history test's own doc depends on that duplication ("four paragraphs earlier in the same request"), and the reviewer counted the paragraphs to confirm it.

All three verified against source before acting; the second commit fixes them, plus a hedge ("rarely needs re-deriving") that was weaker than what the prompt actually says.

Its wider pass for other old-posture text found none — thirteen posture-carrying sites across `prompts.rs`, `engine.rs`, `session.rs`, `server/mod.rs`, `cli.rs`, `AGENTS.md`, `CHANGELOG.md` and `docs/config.md` all read the current way. Two observations it raised need no change: a scripted test fixture re-reads a span its *own* delegated sweep reported, which the preamble explicitly endorses, and the negative assertion at `prompts.rs:1851` now guards a string that appears nowhere else in the tree, so the positive assertion beside it is the one carrying the property.

It was straight about its own limit: `git` is refused in the sandbox, so it could not read the diff and took the description of the old wording from the supplied context rather than verifying it.

PR #204, opened 2026-09-22
Upgrade rmcp to 3.4; test the newest protocol through server/discover

tobert:
Follows the dependency stack (#201, #202), which held rmcp at 3.1.4 because the 3.4 bump failed `tests/mcp_stdio.rs`. Amy said to proceed with the port.

## What changed upstream

rmcp 3.2 made MCP `2026-07-28` a protocol with no `initialize` handshake. A client names its protocol version and capabilities in each request's `_meta` and enters with `server/discover`. An `initialize` asking for `2026-07-28` is now answered with `2025-11-25`, the newest version that still has a handshake. `ProtocolVersion::LATEST` is `2025-11-25` too.

## Decisions

- **`sep_2549_cache_fields` needed no code change.** `RequestContext::protocol_version()` reads the request's `_meta` before the handshake version, so kaibo's `ttlMs: 0` / `cacheScope: private` fill reaches stateless sessions. rmcp's default `discover` answers the same floor from `get_info`. rmcp's handler macros still answer `cacheScope: public`, so kaibo keeps its own fill.
- **The SEP-2549 guard now enters the way a `2026-07-28` client does.** `raw_exchange` takes a message script. The newest-version test sends `server/discover`, `tools/list` and `resources/list`, each with `_meta` and no `initialize`, and asserts the cache fields on all three plus kaibo's serverInfo and instructions in discovery. A second test pins rmcp's `initialize` downgrade, so an rmcp release that changes it shows up beside the guard that depends on it.
- **Negative control:** with the fill disabled, the stateless test fails on `tools/list` ("must carry the REQUIRED `ttlMs`").
- `ServerInfo` → `ServerConfig` (the alias is deprecated in 3.4). The `get_info` comment no longer claims the instructions are read once at `initialize`, because `discover` calls `get_info` on every request.

## Not in this PR

A stateless `2026-07-28` session would still get kaibo's session-wide log notifications (floor `Info`) without a `subscriptions/listen`, and the per-request `logLevel` in `_meta` is ignored. This is not a regression: under rmcp 3.1 the same drain served `2026-07-28` sessions via `initialize`. It is recorded as follow-up work.

## Verification

- `cargo test --no-fail-fast`: 42 binaries, 1385 passed, 0 failed. `cargo clippy --all-targets -- -D warnings` clean.
- `cargo tree -i` reports no match for aws-lc-rs, openssl-sys, and mimalloc.
- Live check through Claude Code (binary swap + `/mcp`): pending. See the comment below.


tobert:
## kaibo review: `consult`, cast `deepseek` (deepseek-flash explorer + synth)

The first attempt was cancelled after about 10 minutes. It hung on one step: an `xxd` of rmcp's compiled `.rmeta` piped to `grep`, because the crate source is outside the allowed set. The retry was given rmcp 3.4's behavior as trusted `context`. Each finding below was checked against the rmcp 3.4 source.

**Fixed in d2c7b38**
- The stateless test drove only 3 of the 6 list/read results. It now also covers `resources/templates/list`, `resources/read` and `prompts/list`. Negative control: with `read_resource`'s fill disabled, it fails on `resources/read`.
- The absence assertions in both legacy-shape tests passed vacuously on a `null` result (an errored request). Each now asserts it got a result object first.
- Two comments still described every session as starting with `initialize` (`main.rs` log drain, `kaish_syntax.rs`).

**Rejected**
- "`get_prompt` misses the required cache fields." rmcp 3.4's `GetPromptResult` has no `ttl_ms` or `cache_scope` (`model.rs:4268`), so there is nothing to fill. The reviewer also flagged this one as unverified.

**Follow-up, not in this PR**
- In a stateless `2026-07-28` session, the log drain sends session-wide `notifications/message` with no `subscriptions/listen`, and ignores send errors. Progress notifications are request-tied, but their send errors are also ignored.
- Found while reviewing, unrelated to rmcp: the hung `xxd | grep` reproduced through `kaibo kaish -c`. It ran until an outer 150 s kill, and `KAISH_EXEC_TIMEOUT` (30 s) never fired. A builtin that never yields can wedge the kaish worker.

tobert:
Live check passed (2026-09-22). I swapped the #204 debug binary into the MCP path and Amy ran `/mcp`. Claude Code reconnected and loaded the tool list. `run_kaish` read the worktree, the `kaibo://config` resource resolved, and `oneshot` (cast `deepseek`) answered. `/proc/<pid>/exe` for the server that started at the reconnect matches the branch binary byte for byte. The `initialize` downgrade to `2025-11-25` is acceptable to Claude Code.

